PR #78, opened 2018-03-14
Add more ssh options to host configuration

kke:
There could be even more, tried to pick up the obvious ones like `port`.

> This is the set of options that Net::SSH.start recognizes. See Net::SSH.start for a description of each option.

```ruby
[
  :auth_methods, :bind_address, :compression, :compression_level, :config,
  :encryption, :forward_agent, :hmac, :host_key,
  :keepalive, :keepalive_interval, :keepalive_maxcount, :kex, :keys, :key_data,
  :languages, :logger, :paranoid, :password, :port, :proxy,
  :rekey_blocks_limit,:rekey_limit, :rekey_packet_limit, :timeout, :verbose,
  :global_known_hosts_file, :user_known_hosts_file, :host_key_alias,
  :host_name, :user, :properties, :passphrase, :keys_only, :max_pkt_size,
  :max_win_size, :send_env, :use_agent, :number_of_password_prompts
]
```

Problemos: the log level, aka `verbose:` takes a symbol, but I have no clue how to coerce a string into a symbol in the dry-struct/types thing and still be able to validate them. (also that is probably better to move out of the options struct and used directly `if ENV['DEBUG']` in the actual `ssh/client.rb`.)

Further improvementos: the ssh_key_path could be moved under the `ssh_opts`.

Example config:

```yaml
  - address: 192.168.100.100
    private_address: 192.168.100.100 # just to advertise correct ip with vagrant
    user: vagrant
    #user: test
    role: master
    ssh_key_path: ~/.vagrant.d/insecure_private_key
    ssh_opts:
      verbose: info
      port: 33
```

Haven't tried all of the options. Also no sort of specs before greenlighted.


SpComb:
Needs merge and docs (`cluster.example.yml`)

kke:
It's possible to define most of these in ~/.ssh/config which will be read by Net::SSH by default on `.start`.

PR #105, opened 2018-03-19
Local persistent volumes

### LINKED ISSUE #104: Persistent local storage
We should provide some sane way to consume host local storage.

See:
- https://kubernetes.io/docs/concepts/storage/storage-classes/#local
- https://github.com/kubernetes-incubator/external-storage/tree/master/local-volume/provisioner

jakolehm:
Fixes #104

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ local-volume-provisioner:
 
 #### Options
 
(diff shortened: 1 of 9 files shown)


kke:
From the days of Kupo, close? :) @jakolehm

PR #111, opened 2018-03-19
Refactor phases to run in parallel across nodes

### LINKED ISSUE #110: Nodes should be configured in parallel
Optimize `kupo up` times by handling nodes in parallel.

All usages of  `Kupo::SSH::Client.for_host` and `Kupo::Kube.client` need to be made threadsafe. It's enough to use a single SSH client for each host and run all the phases per node in serial, but things like `Kupo::Phases::JoinNode` and `Kupo::Phases::LabelNode` can't share the master node's SSH/kube clients if they run in parallel.

SpComb:
Fixes #110

Change the way that phases are run: instead of running all phases in serial for each node in turn, run all nodes in parallel for each phase in turn. The `up` command output is now structured as a phase header with a human-readable title, and then all of the (parallel) phase output is prefixed with the host.

Running the phases in separate threads requires that the `Pharos::SSH::Client` and `Pharos::Kube::Client` must be used in a threadsafe way. Assuming that `net-ssh` and `kubeclient` are not explicitly designed to be threadsafe, that means that each such connection is only allowed to be used by one thread at a time.

For `Pharos::SSH::Client`, this means that each per-node thread for a phase is only allowed to use the SSH connection for that node. This requires change to `Pharos::Phases::JoinNode`: instead of having each node use the `@master_ssh` connection to run `kubeadm token create`, create the token as part of the master node phase, and then use that token in the worker node phase.

For `Pharos::Kube::Client`, any phases that use the `@master` kube client cannot be run in parallel. This is not a problem for the master node phases, because there is only one master host, so it cannot be run in parallel. However, the `Pharos::Phases::LabelNode` runs for each host and also uses the kube client, so it must be run in serial instead.

The global `Kupo::SSH::Client.for_host` and `Pharos::Kube.client` interfaces using class variables are not threadsafe and thus dangerous: not only is the class instance variable hash access racy, but returning the same connection object to two two threads may lead to corrupted requests/connections?

For the SSH connections, this PR also refactors the phases to taken an explicit `ssh` parameter instead of calling `Kupo::SSH::Client.for_host`. The `Pharos::PhaseManager` and `Pharos::SSH::Manager` ensure that each per-node phase thread is only able to use the SSH connection for that node, and it is easier to reason about the thread-safetyness of the code.

The global `Kupo::Kube.client` interface remains non-threadsafe, and it also has an implicit ordering dependency on the `Kupo::Phases::ConfigureClient`... As a workaround the `UpCommand` makes sure to not run any phases that use the kube client in parallel, but this is a bad design because it requires the `UpCommand` to know what the phases are doing... Any phases that are changed to use the kube client must also be changed to run in serial, otherwise they risk subtle threading bugs.

The `Pharos::Kube` interface should somehow be refactored to get rid of the non-threadsafe `Pharos::Kube.client`, but I'm not entirely sure what that would look like yet.

## TODO
- [x] Output formatting to track what's happening across each node
- [ ] Better parallel error handling for the `manager.each_node` map
- [x] Fix specs


lib/kupo/manager.rb
```diff
@@ -0,0 +1,63 @@
+module Kupo
+  class Manager
```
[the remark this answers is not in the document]
SpComb: Dropping the `config` and making this `Pharos:SSH::Manager` would also make sense.

Then this change would just essentially be about removing the pseudo-global `Pharos::SSH` class methods to prevent any thread-unsafe usage - which is indeed the point, using explicit "dependency injection" for the `@ssh` in the phases...

SpComb:
Rewritten for the `Pharos` renaming, as well as the kubeadm/kubelet upgrade changes.

Now contains more phase refactoring in order to use the same common parameters across all phases.

lib/pharos/addon.rb
```diff
@@ -2,11 +2,11 @@
 
 require 'dry-validation'
 require_relative 'addons/struct'
-require_relative 'phases/logging'
+require_relative 'logging'
 
 module Pharos
   class Addon
-    include Pharos::Phases::Logging
+    include Pharos::Logging
```
kke: This is in the #117 too

lib/pharos/configuration/host.rb
```diff
@@ -16,7 +16,11 @@ class Host < Dry::Struct
       attribute :ssh_key_path, Pharos::Types::Strict::String.default('~/.ssh/id_rsa')
       attribute :container_runtime, Pharos::Types::Strict::String.default('docker')
 
-      attr_accessor :os_release, :cpu_arch, :hostname
+      attr_accessor :os_release, :cpu_arch, :hostname, :join_command
+
+      def to_s
+        "#{address}"
```
kke: or just `address.to_s`

lib/pharos/addon.rb
SpComb: This is necessary here because `Pharos::Addon` uses `Pharos::Phases::Logging`, and the logging in `Pharos::Phases::Base` is customized to include the host as a prefix.

lib/pharos/ssh/manager.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      # @param hosts [Array<Pharos::Configuration::Host>]
+      # @return [Array<...>]
+      def with_hosts(hosts)
+        threads = hosts.map do |host|
+          # from the main thread
+          client = client_for(host)
+
+          Thread.new do
```
jakolehm: Probably would make sense to use thread pool here.

lib/pharos/configuration/host.rb
kke: well of course if you want to de-reference then that is fine i suppose :)

lib/pharos/ssh/manager.rb
SpComb: Using a thread pool for minimizing `Thread.new` overhead is probably not that relevant here, because phases are pretty heavyweight (remote SSH/API calls).

When running with very large numbers of nodes, you might also want to start limiting the numbers of nodes handled in parallel at some point... but that can also happen later as an optimization, without needing to chance the interface used by `Pharos::UpCommand`.

OTOH the error handling could be a bit more fine-tuned, right now it joins the threads in order, so if some nodes fail with an exception, then it's a bit random as to which error this raises, and how long it waits for the earlier nodes to complete before raising.

lib/pharos/up_command.rb
```diff
@@ -77,27 +77,39 @@ def configure(config)
...
+      load_phases
+      puts pastel.green("==> Starting to craft cluster ...")
+
+      # set workdir to the same dir where config was loaded from
+      # so that the certs etc. can be referenced more easily
+      Dir.chdir(config_yaml.dirname) do
+        handle_phase(Phases::ValidateHost, config.hosts, config: config)
+        handle_phase(Phases::ConfigureHost, config.hosts, config: config)
+
+        handle_phase(Phases::ConfigureMaster, master_hosts, config: config, master: master_host)
```
kke: these could be something like:
`queue << Phases::ConfigureHost.new(config.hosts, config: config)`

and the thing that runs the phases / queue could then just expect that it can `queue.pop.call` them without any args.

this way it would be possible also to do something like:

```
queue.later do 
  do stuff
end
```

and it would magically work.

Also then you wouldn't have to normalize the phases, you can just `queue << FooPhase.new(whatever_args_it_wants)`

SpComb: Not sure how putting them into a `Queue` would be any better than just calling them normally from a function? All the phases run in serial, it's much simpler that way... there's no need to parallelize the phases.

kke: queue was not capitalized.

but yes, no need for a queue there, but the thing that does the `.call` wouldn't then have to do the `.new` and play with the arguments.

kke:
```
handle_phase(Phases:Foofoo.new(xyz))
```

SpComb: Yeah, that part is mostly about abstracting the `ssh` away, without needing to defer it until after `new`.

However, then there's also the question of what the abstraction should be for e.g. phases that  don't need `SSH`, or even those that are not node-specific, but apply to the entire kube cluster (e.g. `ConfigureNetwork`). However, that is also going to need bigger changes for multi-master support, and I don't know what those are going to look like yet... the `Pharos::Kube.apply_stack(@master.address, ...)` has to change as well.

SpComb:
Examples of the current WIP output format:

screenshot from 2018-04-09 17-28-38
screenshot from 2018-04-09 17-29-03

lib/pharos/phases/base.rb
```diff
@@ -1,17 +1,42 @@
...
+        @ssh = ssh
+        @master = master
+      end
+
+      def logger
+        @logger ||= Logger.new($stdout).tap do |logger|
+          logger.progname = "#{@host}"
+          logger.level = ENV["DEBUG"] ? Logger::DEBUG : Logger::INFO
+          logger.formatter = proc do |_severity, _datetime, progname, msg|
+            "    [%<progname>s] %<msg>s\n" % { progname: progname, msg: msg }
```
SpComb: This needs something to align the `progname` to keep the output between multiple nodes neat.

jakolehm:
> Examples of the current WIP output format, without DEBUG:

Looks quite ok already, I would change phase class name to something more human readable.

lib/pharos/phases/base.rb
```diff
@@ -1,17 +1,42 @@
...
 require_relative '../phases'
 
 module Pharos
   module Phases
     class Base
-      include Pharos::Phases::Logging
+      # @return [String]
+      def self.title
+        # XXX: prettier
+        self.name
```
SpComb: This currently returns the machine-readable `Pharos::Phases::ConfigureHost` name, needs to be humanized.

lib/pharos/up_command.rb
```diff
@@ -77,27 +77,39 @@ def configure(config)
...
+        handle_phase(Phases::ConfigureClient, master_hosts, config: config, master: master_host)
+        handle_phase(Phases::ConfigureDNS, master_hosts, config: config, master: master_host)
+        handle_phase(Phases::ConfigureNetwork, master_hosts, config: config, master: master_host)
+        handle_phase(Phases::LabelNode, master_hosts, config: config, master: master_host)
+        handle_phase(Phases::StoreClusterYAML, master_hosts, master: master_host, config_content: config_yaml.read(ENV.to_h))
+
+        handle_phase(Phases::ConfigureKubelet, worker_hosts(config), config: config, master: master_host)
+        handle_phase(Phases::JoinNode, worker_hosts(config), config: config, master: master_host)
+        handle_phase(Phases::LabelNode, worker_hosts(config), config: config, master: master_host)
```
SpComb: These all use the kube API, and should not run in parallel with multiple threads. This may even break if the kubeclient is not threadsafe.

lib/pharos/up_command.rb
```diff
@@ -77,27 +77,39 @@ def configure(config)
...
+        handle_phase(Phases::ConfigureHost, config.hosts, config: config)
+
+        handle_phase(Phases::ConfigureMaster, master_hosts, config: config, master: master_host)
+        handle_phase(Phases::ConfigureClient, master_hosts, config: config, master: master_host)
+        handle_phase(Phases::ConfigureDNS, master_hosts, config: config, master: master_host)
+        handle_phase(Phases::ConfigureNetwork, master_hosts, config: config, master: master_host)
+        handle_phase(Phases::LabelNode, master_hosts, config: config, master: master_host)
+        handle_phase(Phases::StoreClusterYAML, master_hosts, master: master_host, config_content: config_yaml.read(ENV.to_h))
+
+        handle_phase(Phases::ConfigureKubelet, worker_hosts(config), config: config, master: master_host)
```
SpComb: The `ConfigureKubelet` phase could run in parallel across both the master + worker hosts, *if* this is an initial install... the upgrade case is a little more complicated.

That logic is currently handled within the `ConfigureMaster` phase, it might make sense to deal with here instead?

lib/pharos/phases/base.rb
kke: (or stick it in the back)

How about some sort of color pool like in kontena logs? Like use `colors(progname[-1].to_i)` for example.

kke:
The up command is getting bloated.

lib/pharos/phase_manager.rb
```diff
@@ -0,0 +1,75 @@
...
+
+      run(phases, parallel: parallel) do |phase|
+        phase.call
+      end
+    end
+
+    # @param dirs [Array<String>]
+    def load_phases(dirs)
+      dirs.each do |dir|
+        Dir.glob(File.join(dir, '*.rb')).each { |f| require(f) }
```
kke: Maybe put a sort in there

SpComb: This is from the `AddonManager`, which also needs a `sort`.

lib/pharos/phases/base.rb
```diff
@@ -8,8 +8,7 @@ module Phases
     class Base
       # @return [String]
       def self.title
-        # XXX: prettier
-        self.name
+        self.const_get(:PHASE_TITLE) || self.name
```
kke: Could be:

```ruby
def self.title(title = nil)
  @title ||= title || name
end

class FooPhase < Base
  title "Configure foofoo"

  def call
    ...
  end
end
```

lib/pharos/phases/base.rb
SpComb: I think if we're already using colors for the `UpCommand` headers (and also the SSH debug output), then it would become a mess... if we wanted to use per-node colors, then IMO we would need to stop using colors for the top-level `UpCommand` output.

The basic INFO-level phase logging is terse enough for this to be readable even without colors, IMO... better to just keep each phase relatively short so the nodes don't get too out of sync.

The SSH `DEBUG` output probably also needs some per-host progname headers... I don't think it's worth the effort to get too fancy with those.

kke: I don't see any use for the DEBUG_SSH unless it's flipped like in #117 (the full blown festive  mega-debug thing only with DEBUG_SSH).

lib/pharos/phases/base.rb
```diff
@@ -1,17 +1,45 @@
...
+require 'logger'
 require_relative '../phases'
 
 module Pharos
   module Phases
     class Base
-      include Pharos::Phases::Logging
+      # @return [String]
+      def self.title
+        self.const_get(:PHASE_TITLE) || self.name
```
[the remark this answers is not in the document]
kke: The one in the comment.  imo.

lib/pharos/phases/base.rb
SpComb: I originally tried `self.title=(title)`, but it seems like `title = "foo"` doesn't invoke that, I guess it defines a const instead?

Is it a good pattern to use the same method as both an accessor and setter as well?

SpComb:
I can do this PR without any phase refactoring, just leaving the `Pharos::SSH.client_for` in... all it *really* needs is the change to `Pharos::Phases::JoinNode` to not use `@master_ssh`.

The point of the phases refactoring is to prevent anyone from writing thread-unsafe code by switching to a dependency injection type pattern, where the manager explicitly passes a different `Pharos::SSH::Client` to each thread. However, the same problem also applies to the `Pharos::Kube::Client`, and that isn't refactored in this PR... so it would really need MOAR REFACTORING, and I don't want to do that in this PR.

lib/pharos/phases/configure_client.rb
```diff
@@ -1,28 +1,33 @@
...
         File.chmod(0o600, config_file) if File.exist?(config_file)
         File.write(config_file, config_data.gsub(%r{(server: https://)(.+)(:6443)}, "\\1#{@master.address}\\3"), perm: 0o600)
         logger.info { "Configuration saved to #{config_file}" }
       end
 
-      def config_dir
-        File.join(Dir.home, '.pharos')
+      def create_join_command
+        logger.info { "Creating node bootstrap token ..." }
+        @master.join_command = @ssh.exec!("sudo kubeadm token create --print-join-command").split(' ')
```
SpComb: This will now create one bootstrap token on each `pharos-cluster up` run, instead of having `JoinNode` create one for each node during the initial run, but not creating any if the nodes are already joined.

So fewer tokens on the initial `pharos-cluster up` run, but more on later runs.

lib/pharos/up_command.rb
```diff
@@ -152,42 +141,39 @@ def worker_hosts(config)
...
+      # master is now configured and can be used
+      handle_phase(Phases::ConfigureDNS, [master_host], master: master_host)
+      handle_phase(Phases::ConfigureNetwork, [master_host], master: master_host)
+      handle_phase(Phases::ConfigureMetrics, [master_host], master: master_host)
+      handle_phase(Phases::LabelNode, master_hosts, master: master_host)
+      handle_phase(Phases::StoreClusterYAML, [master_host], master: master_host, config_content: config_yaml.read(ENV.to_h))
+
+      handle_phase(Phases::ConfigureKubelet, worker_hosts, ssh: true, parallel: true) # TODO: also run this phase in parallel with Phases::ConfigureMaster?
+      handle_phase(Phases::JoinNode, worker_hosts, ssh: true, parallel: true, master: master_host) # XXX: this only uses the @master.join_command from the config, and is thus threadsafe
+      handle_phase(Phases::LabelNode, worker_hosts, master: master_host) # NOTE: uses the @master kube API for each node, not threadsafe
```
SpComb: This is the problem: we need to know what each phase is doing with the `master` in order to know if it's safe to run in parallel. If the phase uses `Pharos::Kube.client(@master.address)`, then we can't run it in parallel.

lib/pharos/up_command.rb
```diff
@@ -152,42 +141,39 @@ def worker_hosts(config)
...
+      handle_phase(Phases::ValidateHost, all_hosts, ssh: true, parallel: true)
+      handle_phase(Phases::ConfigureHost, all_hosts, ssh: true, parallel: true)
+
+      handle_phase(Phases::ConfigureMaster, master_hosts, ssh: true)
+      handle_phase(Phases::ConfigureClient, master_hosts, ssh: true)
+
+      # master is now configured and can be used
+      handle_phase(Phases::ConfigureDNS, [master_host], master: master_host)
+      handle_phase(Phases::ConfigureNetwork, [master_host], master: master_host)
+      handle_phase(Phases::ConfigureMetrics, [master_host], master: master_host)
```
SpComb: These three phases don't actually use the `@host` at all... they only use the `@config` and `Pharos::Kube.client(@master.address)`. The `[master_host]` is only used for the log output.

lib/pharos/up_command.rb
```diff
@@ -152,42 +141,41 @@ def worker_hosts(config)
...
+      handle_phase(Phases::StoreClusterYAML, [master_host], master: master_host, config_content: config_yaml.read(ENV.to_h))
+      handle_phase(Phases::ConfigureBootstrap, [master_host], master: master_host)
+
+      handle_phase(Phases::JoinNode, worker_hosts, ssh: true, parallel: true)
+
+      handle_phase(Phases::LabelNode, all_hosts, master: master_host, ssh: false, parallel: false) # NOTE: uses the @master kube API for each node, not threadsafe
```
SpComb: Assuming that it's fine to label all the nodes at the end, rather than labeling the master nodes before joining the worker nodes.

jakolehm:
> The exception would be Phases::ConfigureMaster for the case of multiple masters, and needing to copy certs from the leader to the other master nodes... could that get split up into a separate phase for the leader + parallel phases for the followers?

Yes, this makes sense.

lib/pharos/phase_manager.rb
```diff
@@ -0,0 +1,75 @@
...
+      @options = options
+
+      load_phases(dirs)
+    end
+
+    # @param phases [Array<Pharos::Phases::Base>]
+    # @return [Array<...>]
+    def run_parallel(phases)
+      threads = phases.map { |phase|
+        Thread.new do
```
SpComb: This is now an entirely generic parallel map, with nothing at all specific to phases.

lib/pharos/phase_manager.rb
SpComb: None of the files under `phases/*` require anything anymore, so the ordering *shouldn't* matter :thinking:

lib/pharos/up_command.rb
```diff
@@ -152,42 +141,41 @@ def worker_hosts(config)
...
-      Phases::ConfigureDNS.new(master, config).call
-      Phases::ConfigureNetwork.new(master, config.network).call
-      Phases::ConfigureMetrics.new(master).call
-      Phases::LabelNode.new(master, master).call
-      Phases::StoreClusterYAML.new(master, config_yaml.read(ENV.to_h)).call
+    def handle_phases(config)
+      all_hosts = config.hosts
+      master_hosts = master_hosts(config)
+      master_host = master_hosts[0]
+      worker_hosts = worker_hosts(config)
```
SpComb: This part is a bit of a mess, I think the `config` should be stored as an instance variable instead of getting passed around as a function argument in order to clean this up...

lib/pharos/phase.rb
```diff
@@ -0,0 +1,67 @@
...
+
+    # @param host [Pharos::Configuration::Host]
+    # @param config [Pharos::Config]
+    # @param ssh [Pharos::SSH::Client]
+    # @param master [Pharos::Configuration::Host]
+    def initialize(host, config: nil, ssh: nil, master: nil)
+      @host = host
+      @config = config
+      @ssh = ssh
+      @master = master
```
SpComb: @kke the only thing that this `@master` is now used for is now `Pharos::Kube.client(@master.address)`:

```
lib/pharos/phases/configure_dns.rb:          @master.address,
lib/pharos/phases/configure_metrics.rb:        cert_manager = Pharos::Kube::CertManager.new(@master, 'metrics-server-client-cert', namespace: 'kube-system')
lib/pharos/phases/configure_metrics.rb:          @master.address, 'metrics-server',
lib/pharos/phases/configure_metrics.rb:        cert_manager = Pharos::Kube::CertManager.new(@master, 'heapster-client-cert', namespace: 'kube-system')
lib/pharos/phases/configure_metrics.rb:          @master.address, 'heapster',
lib/pharos/phases/configure_network.rb:        kube_client = Pharos::Kube.client(@master.address)
lib/pharos/phases/configure_network.rb:          @master.address, 'weave',
lib/pharos/phases/label_node.rb:        @kube ||= Pharos::Kube.client(@master.address)
lib/pharos/phases/store_cluster_yaml.rb:          Pharos::Kube.update_resource(@master.address, configmap)
lib/pharos/phases/store_cluster_yaml.rb:          Pharos::Kube.create_resource(@master.address, configmap)
```

That's what I mean by the `Pharos::Kube` refactoring: these class method calls with the `@master.address` should be replaced by something called `@kube` with instance methods.

SpComb: The only exception is the `ConfigureMaster#upgrade?` use of `Pharos::Kube.config_exists?(@host.address)`... I suspect that check is broken for the multi-master case, and needs some different logic anyways.

lib/pharos/cluster_manager.rb
```diff
@@ -0,0 +1,89 @@
...
+      apply_phase(Phases::StoreClusterYAML, [config.master_host], master: config.master_host, config_content: @config_content)
+      apply_phase(Phases::ConfigureBootstrap, [config.master_host], ssh: true) # using `kubeadm token`, not the kube API
+
+      apply_phase(Phases::JoinNode, config.worker_hosts, ssh: true, parallel: true)
+
+      apply_phase(Phases::LabelNode, config.hosts, master: config.master_host, ssh: false, parallel: false) # NOTE: uses the @master kube API for each node, not threadsafe
+    end
+
+    def apply_phase(phase_class, hosts, **options)
+      puts pastel.cyan("==> #{phase_class.title} @ #{hosts.join(' ')}")
```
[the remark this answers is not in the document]
kke: Overly fancy:

```ruby
> hosts.group_by { |h| h.split('.').first(2).join('.') }.map { |n, h| "#{n}.* (#{h.size} hosts)" }.join(', ')
=> "10.2.* (2 hosts), 192.168.* (2 hosts)"
```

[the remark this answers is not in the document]
SpComb: I figure we can tweak the output later, right now IMO this is the simplest thing that works, unless we leave out the `@ nodes...` part completely?

SpComb:
This PR is now about 90% refactoring for phase/node ordering and output formatting changes, 5% changes to the join command handling, and maybe 5% about actually running each phase in parallel.

lib/pharos/cluster_manager.rb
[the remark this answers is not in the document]
jakolehm: Let's keep this simple, we can improve outputs in follow-up pr's if necessary.

lib/pharos/phase.rb
```diff
@@ -0,0 +1,67 @@
+# frozen_string_literal: true
+
+require 'logger'
+
+module Pharos
+  class Phase
+    PHASE_TITLE = nil
```
jakolehm: What is this?

SpComb: Old code from before `title "..."`

lib/pharos/phase_manager.rb
```diff
@@ -64,9 +64,7 @@ def prepare_phase(phase_class, host, ssh: false, **options)
     def apply(phase_class, hosts, parallel: false, **options)
       phases = hosts.map { |host| prepare_phase(phase_class, host, **options) }
 
-      run(phases, parallel: parallel) do |phase|
-        phase.call
-      end
+      run(phases, parallel: parallel, &:call)
```
kke: Looks funny that way. Maybe the runner should do the `.call`ing.

lib/pharos/ssh/manager.rb
```diff
@@ -8,13 +9,11 @@ def initialize
...
       def client_for(host)
-        @clients[host] ||= Pharos::SSH::Client.new(host.address, host.user, keys: [host.ssh_key_path]).tap do |client|
-          client.connect
-        end
+        @clients[host] ||= Pharos::SSH::Client.new(host.address, host.user, keys: [host.ssh_key_path]).tap(&:connect)
       end
 
       def disconnect_all
-        @clients.each do |host, client|
+        @clients.each do |_host, client|
```
kke: could be `@clients.each_value(&:disconnect)` to avoid the `_host`

lib/pharos/phase_manager.rb
SpComb: Blame `rubocop`

```
lib/pharos/phase_manager.rb:67:39: C: Style/SymbolProc: Pass &:call as an argument to run instead of a block.
      run(phases, parallel: parallel) do |phase| ...
```

lib/pharos/phase_manager.rb
```diff
@@ -64,7 +64,13 @@ def prepare_phase(phase_class, host, ssh: false, **options)
     def apply(phase_class, hosts, parallel: false, **options)
       phases = hosts.map { |host| prepare_phase(phase_class, host, **options) }
 
-      run(phases, parallel: parallel, &:call)
+      run(phases, parallel: parallel) do |phase|
```
kke: (I still don't see why the runner doesn't do the calling, it's already doing all sorts of magic with `prepare_phase` method for the `phases`, so it obviously expects the `phases` to be phases, which can be `.call`ed)

SpComb: `prepare_phase` is used from `apply`, which is also the same method that knows about `call`?

lib/pharos/phase_manager.rb
```diff
@@ -0,0 +1,76 @@
...
+
+    # @param phases [Array<Pharos::Phases::Base>]
+    # @return [Array<...>]
+    def run_parallel(phases)
+      threads = phases.map { |phase|
+        Thread.new do
+          begin
+            yield phase
+          rescue StandardError => exc
+            puts " [#{phase}] #{exc.class}: #{exc.message}"
```
SpComb: If we get some better `Pharos::Logging`, then this should probably be something like `logger.error`... ATM it makes very little difference.

This is needed to report errors as soon as they happen, because it may take some time before the `thread.value` re-raises it.

PR #117, opened 2018-03-20
Refactor debug logging

kke:
More modular, less `iffing`, less garbage string generation.

Including `Pharos::Logging` will give you `logger` and `debug?`.

Flips the logic of `DEBUG=` vs `DEBUG_SSH=`, you only get the fancy ssh debug with `DEBUG_SSH`.


jakolehm:
Conflicts ...

kke:
Already partially done in #111

PR #133, opened 2018-03-21
Pharos::SSH::Tempfile

kke:


lib/pharos/ssh/client.rb
```diff
@@ -243,26 +261,10 @@ def read_file(path)
...
-      def with_tmpfile(contents, prefix: "pharos")
-        path = File.join('/tmp', prefix + '.' + SecureRandom.hex(16))
-
-        upload(StringIO.new(contents), path)
-
-        yield path
-      ensure
-        exec("rm #{path}") if path
+      def write_file(path, content)
+        exec!("cat | sudo tee #{path} > /dev/null", stdin: content)
```
kke: I guess it doesn't actually need the 🐱

kke:
#142 and #143 are now `Pharos::SSH::RemoteFile` and `::RemoteCommand` maybe this should be `::RemoteTempfile` to match those :)

kke:
Closed in favor of #223

PR #143, opened 2018-03-22
Pharos::SSH::RemoteCommand

kke:
The saga continues

Extracts the Exec class from `ssh/client.rb` to a separate `Pharos::SSH::RemoteCommand` class.


kke:
Closed in favor of #223

PR #157, opened 2018-03-28
Throw usage error if cluster.yml has unexpected top level keys

### LINKED ISSUE #93: Throw error on unknown keys in cluster.yml

kke:
Partially fixes #93


SpComb:
Making this specific to the top-level config keys in `UpCommand` seems a little too limited to be useful, most config keys are in the nested schemas... the `input` macro at https://github.com/dry-rb/dry-validation/issues/179#issuecomment-304094883 seems like a better approach, although that seems to be limited to a generic "has unknown keys" rather than listing the invalid keys :/

kke:
Yep

PR #168, opened 2018-04-03
Allow configuring kube-proxy mode

SpComb:
Supports running kube-proxy in IPVS mode (beta).


SpComb:
TBH it doesn't really make a whole lot of sense to configure this, given that eventually `ipvs` will be made the default, which assumes that it gains feature parity with the current `iptables` default. The only usecases for this configuration knob are:

* experimenting with the new `ipvs` mode while still in beta (possibly less CPU overhead than iptables => generally better perf, unless suffering from softirq imbalances)
* as a workaround for any issues with the default mode, such as falling back to `userspace`... or to `iptables` once `ipvs` becomes the default

SpComb:
> docs missing

Could be worth mentioning the different options + a link to https://kubernetes.io/docs/concepts/services-networking/service/#virtual-ips-and-service-proxies and/or https://github.com/kubernetes/kubernetes/tree/master/pkg/proxy/ipvs for additional context.

I suppose that would happen in the `README.md`.

jakolehm:
> I suppose that would happen in the README.md.

Yes.

SpComb:
Looks like kubeadm 1.10 has some changes to the `FeatureGates`:

```
    $ sudo kubeadm init --ignore-preflight-errors all --skip-token-print --config /tmp/kubeadm.cfg.4178303c34da8fb9c128a17f449c8cc4
    unable to decode config from "/tmp/kubeadm.cfg.4178303c34da8fb9c128a17f449c8cc4" [v1alpha1.MasterConfiguration: KubeProxy: v1alpha1.KubeProxy: Config: v1alpha1.KubeProxyConfiguration: FeatureGates: ReadMapCB: expect { or n, but found ", error found in #10 byte of ...|reGates":"SupportIPV|..., bigger context ...|iguration","kubeProxy":{"config":{"featureGates":"SupportIPVSProxyMode=true","mode":"ipvs"}},"kubern|...]
    ! 1
    $ rm /tmp/kubeadm.cfg.4178303c34da8fb9c128a17f449c8cc4
    ! 0
SSH exec failed with code 1: sudo kubeadm init --ignore-preflight-errors all --skip-token-print --config /tmp/kubeadm.cfg.4178303c34da8fb9c128a17f449c8cc4
unable to decode config from "/tmp/kubeadm.cfg.4178303c34da8fb9c128a17f449c8cc4" [v1alpha1.MasterConfiguration: KubeProxy: v1alpha1.KubeProxy: Config: v1alpha1.KubeProxyConfiguration: FeatureGates: ReadMapCB: expect { or n, but found ", error found in #10 byte of ...|reGates":"SupportIPV|..., bigger context ...|iguration","kubeProxy":{"config":{"featureGates":"SupportIPVSProxyMode=true","mode":"ipvs"}},"kubern|...]
```

lib/pharos/phases/configure_master.rb
```diff
@@ -296,6 +298,25 @@ def generate_authentication_token_webhook_config(webhook_config)
...
+
+        if @config.kube_proxy.mode
+          config['kubeProxy']['config']['mode'] = @config.kube_proxy.mode
+        end
+
+        if @config.kube_proxy.mode == 'ipvs'
+          config['kubeProxy']['config']['featureGates']['SupportIPVSProxyMode'] = true
+        end
+
+        config
```
SpComb: This return is only here to workaround bad rubocop suggestions:

```
lib/pharos/phases/configure_master.rb:313:9: C: Style/GuardClause: Use a guard clause instead of wrapping the code inside a conditional expression.
        if @config.kube_proxy.mode == 'ipvs'
        ^^
```

SpComb:
...except that the kube-proxy configmap that kubeadm 1.10 generates does not work for the 1.9 kube-proxy:

```
error: v1alpha1.KubeProxyConfiguration: FeatureGates: ReadString: expects " or n, parsing 458 ...reGates":{... at {"apiVersion":"kubeproxy.config.k8s.io/v1alpha1","bindAddress":"0.0.0.0","clientConnection":{"acceptContentTypes":"","burst":10,"contentType":"application/vnd.kubernetes.protobuf","kubeconfig":"/var/lib/kube-proxy/kubeconfig.conf","qps":5},"clusterCIDR":"10.32.0.0/12","configSyncPeriod":"15m0s","conntrack":{"max":null,"maxPerCore":32768,"min":131072,"tcpCloseWaitTimeout":"1h0m0s","tcpEstablishedTimeout":"24h0m0s"},"enableProfiling":false,"featureGates":{"SupportIPVSProxyMode":true},"healthzBindAddress":"0.0.0.0:10256","hostnameOverride":"","iptables":{"masqueradeAll":false,"masqueradeBit":14,"minSyncPeriod":"0s","syncPeriod":"30s"},"ipvs":{"minSyncPeriod":"0s","scheduler":"","syncPeriod":"30s"},"kind":"KubeProxyConfiguration","metricsBindAddress":"127.0.0.1:10249","mode":"ipvs","nodePortAddresses":null,"oomScoreAdj":-999,"portRange":"","resourceContainer":"/kube-proxy","udpIdleTimeout":"250ms"}
```

So I don't think it's currently possible to configure IPVS for kube-proxy 1.9 using kubeadm 1.10?

SpComb:
Tested to work on kube 1.10

jakolehm:
@SpComb docs?

SpComb:
https://github.com/kontena/pharos-docs/pull/9

PR #187, opened 2018-04-04
README: Document the hosts section of cluster.yml

kke:


kke:
Required fields on top, optional/fields with defaults on bottom.

PR #197, opened 2018-04-05
Option to read hosts from terraform output json

jakolehm:
Improves integration with Terraform.

### main.tf:
```tf
variable "auth_token" {
  description = "Packet authentication token"
}

variable "project_id" {
  description = "Packet project id"
}

provider "packet" {
  auth_token = "${var.auth_token}"
}

resource "packet_device" "pharos_master" {
  hostname         = "pharos-master"
  plan             = "baremetal_0"
  facility         = "ams1"
  operating_system = "ubuntu_16_04"
  billing_cycle    = "hourly"
  project_id       = "${var.project_id}"
}

resource "packet_device" "pharos_worker" {
  count            = 3
  hostname         = "pharos-worker-${count.index}"
  plan             = "baremetal_0"
  facility         = "ams1"
  operating_system = "ubuntu_16_04"
  billing_cycle    = "hourly"
  project_id       = "${var.project_id}"
}

output "pharos" {
  value = {
    masters = {
      address         = "${packet_device.pharos_master.*.network.0.address}"
      private_address = "${packet_device.pharos_master.*.network.2.address}"
      role            = "master"
      user            = "root"
    }

    workers = {
      address         = "${packet_device.pharos_worker.*.network.0.address}"
      private_address = "${packet_device.pharos_worker.*.network.2.address}"

      label = {
        foo = "bar"
      }

      user = "root"
      role = "worker"
    }
  }
}
```

### cluster-without-hosts.yml

```yaml
network:
  pod_network_cidr: 172.32.0.0/16
  trusted_subnets:
    - 10.80.0.0/16
addons: {}
```

### Usage 
```sh
$ terraform apply
$ terraform output -json > tf.json
$ pharos-cluster up -c cluster-without-hosts.yml --hosts-from-tf ./tf.json
```


jakolehm:
@kke imo it would make sense to pipe output from `terraform output` directly to `pharos-cluster` but stdin read assumes currently full cluster.yml. Any ideas how to improve?

matti:
This looks great!  The only thing that concerns me is that this assumes that all workers have the same labels and connection settings. How about setting the workers as a list of those maps:

```
    workers = [
      {
          address         = "${packet_device.pharos_hard_worker.*.network.0.address}"
         ...
      },
     {
          address         = "${packet_device.pharos_slow_worker.*.network.0.address}"
          ...
      }
    ]
  }
```

?

jakolehm:
Maybe my example was a bit too simplified.. but you can already have different "groups" of workers.

```tf
output "pharos" {
  value = {
    masters = {
      address         = "${packet_device.pharos_master.*.network.0.address}"
      private_address = "${packet_device.pharos_master.*.network.2.address}"
      role            = "master"
      user            = "root"
    }

    type_0 = {
      address         = "${packet_device.pharos_worker.*.network.0.address}"
      private_address = "${packet_device.pharos_worker.*.network.2.address}"

      label = {
        metal = "type0"
      }

      user = "root"
      role = "worker"
    }

    type_2 = {
      address         = "${packet_device.pharos_worker2.*.network.0.address}"
      private_address = "${packet_device.pharos_worker2.*.network.2.address}"

      label = {
        metal = "type2"
      }

      user = "root"
      role = "worker"
    }
  }
}
```

matti:
ok! didnt realize that is was this fancy

kke:
Defaulting to `./cluster.yml` is a bit problematic here. It's difficult to figure out if it should it read it first and merge the stuff from stdin or should input from stdin make it skip trying to read the default file at all? 

Also maybe the config path could be a param instead of an option.

If the config param would be multivalued and it would `config_list.map {|x| parse(x)}.reduce(:merge)`, then you could do something like `terraform --foo | pharos-cluster up -c cluster.yml -c network_settings.yml`. Of course this would require that it can detect which inputs to pass through the terraform parser (which i guess could be done by just checking if the first char is `{`).

jakolehm:
> Defaulting to ./cluster.yml is a bit problematic here.

I agree, should we drop the default?

kke:
Depends a bit how it's supposed to be used.

If the idea is to have something like:
```
addons/foofoo/foo.rb
cluster.yml
```

in git repos, then maybe it makes sense.

kke:
Or maybe default to `./cluster.yml` only as the final last solution if there's no param and nothing in the stdin.

But this too will be confusing if you're used to do `pharos-cluster up` and then want to pipe the terraform stuff on top, you need to specify the `cluster.yml` as option explicitly.

jakolehm:
Ok, maybe stdin change should not be part of this PR. IMO reading from json file is good enough for most use cases.

lib/pharos/up_command.rb
```diff
@@ -22,13 +31,47 @@ def default_config_yaml
 
     def execute
       puts pastel.green("==> Reading instructions ...")
-      configure(load_config)
+      config_hash = load_config
+      if hosts_from_tf
+        puts pastel.green("==> Importing hosts from Terraform ...")
+        config_hash['hosts'] = load_tf_json
```
SpComb: Perhaps a merge rather than a replace?

jakolehm: Now it appends to hosts array.

PR #201, opened 2018-04-05
Use Pharos::YamlFile in phases/base parse_resource_file

kke:


lib/pharos/phases/base.rb
```diff
@@ -21,8 +21,7 @@ def exec_script(script, vars = {})
 
       def parse_resource_file(path, vars = {})
```
kke: This looks a whole lot like the one in `Pharos::Kube`, which I think is in a wrong place too

kke: Hmm, it's a class method. This could just call it with `Pharos::Kube.parse_resource_file` I think?

SpComb:
It's used for non-YAML files too.

lib/pharos/phases/base.rb
SpComb: That returns a `Kubeclient::Resource`, which is the wrong thing here.

lib/pharos/phases/base.rb
```diff
@@ -21,8 +21,7 @@ def exec_script(script, vars = {})
...
         path = File.realpath(File.join(__dir__, '..', 'resources', path))
-        yaml = File.read(path)
-        Pharos::Erb.new(yaml).render(vars)
+        Pharos::YamlFile.new(path).load(vars)
```
SpComb: This needs to return a string with the encoded YAML?

kke:
So does it need to work with non-yamls?

SpComb:
Fixes part of #203 but not completely:

```
    Pushing audit configs to master
    $ sudo mkdir -p /etc/pharos/audit
    ! 0
No such file or directory @ realpath_rec - /app/lib/pharos/resources/audit/webhook-config.yml
/app/lib/pharos/phases/base.rb:23:in `realpath'
/app/lib/pharos/phases/base.rb:23:in `parse_resource_file'
/app/lib/pharos/phases/configure_master.rb:74:in `install'
/app/lib/pharos/phases/configure_master.rb:29:in `call'
/app/lib/pharos/up_command.rb:132:in `handle_masters'
/app/lib/pharos/up_command.rb:46:in `block in configure'
/app/lib/pharos/up_command.rb:45:in `chdir'
/app/lib/pharos/up_command.rb:45:in `configure'
/app/lib/pharos/up_command.rb:25:in `execute'
/app/lib/pharos/root_command.rb:14:in `run'
```

SpComb:
> So does it need to work with non-yamls?

It's only used for .yml right now, but there's a need for templating non-yml files like systemd units as well.

SpComb:
Tested to fix #203 now

kke:
I guess we need to bring `Pharos::Erb` back in the future and make `YamlFIle` use that for the parse part (and move the `YamlFile::Namespace` under it)

SpComb:
> I guess we need to bring Pharos::Erb back in the future and make YamlFIle use that for the parse part (and move the YamlFile::Namespace under it)

Yeah, most of the `Pharos::YamlFile` logic is not actually YAML-specific at all.

kke:
`Pharos::ResourceFile` sounds good too. It could also have `.to_kube`.

lib/pharos/phases/base.rb
```diff
@@ -21,8 +21,7 @@ def exec_script(script, vars = {})
 
       def parse_resource_file(path, vars = {})
         path = File.realpath(File.join(__dir__, '..', 'resources', path))
-        yaml = File.read(path)
-        Pharos::Erb.new(yaml).render(vars)
+        Pharos::YamlFile.new(path).read(vars)
```
SpComb: The `Pharos::YamlFile#read` is entirely generic, and not actually YAML-specific, so this works nicely for any "resource" files...

Should probably be renamed to something like `upload_config_file`, using `lib/pharos/configs` or such... the `resources` are for Kube resources.

jakolehm: Maybe open an issue about this?

SpComb: #204

PR #222, opened 2018-04-09
Multi-master support

### LINKED ISSUE #12: Support multiple masters

### LINKED ISSUE #140: Master does not get upgraded if ~/.pharos/* is missing
If you e.g. run kupo on a different machine that doesn't have any of the `~/.kupo` master configs in place yet, then you need to run `kupo up twice` in order to upgrade. The first run will log `Kubernetes control plane is up-to-date` and leave the master on the old version while upgrading the workers.

https://github.com/kontena/kupo/blob/2b544f0e1a4ad324605034ff3818d01a5d2f283d/lib/kupo/phases/configure_master.rb#L33-L34

```
    Checking if Kubernetes control plane is already initialized ...
    $ [ -e /etc/kubernetes/admin.conf ]
    ! 0
    Kubernetes control plane is up-to-date.
    Fetching kubectl config ...
    $ sudo cat /etc/kubernetes/admin.conf
    ...
    Configuration saved to /root/.kupo/167.99.36.141
    ...
==> worker: 167.99.36.158
    Configuring Kubernetes packages ...
    $ sudo chmod +x /tmp/665320911f92bf67753af5f9b38dcc59 && sudo /tmp/665320911f92bf67753af5f9b38dcc59
    ++ kubelet --version
    + '[' 'Kubernetes v1.9.5' = 'Kubernetes v1.10.0-rc.1' ']'
    ...
```

jakolehm:
This PR adds etcd clustering and multi-master functionality. Each master host embeds etcd peer. Etcd peer and client communication is secured with TLS.

Fixes #12 
Fixes #140 

Todo:
- [x] etcd certificates
- [x] etcd cluster
- [x] multiple master hosts
- [x] kubelet proxy
- [x] control-plane (api) external address


lib/pharos/scripts/configure-etcd-certs.sh
```diff
@@ -0,0 +1,125 @@
+#!/bin/bash
+
+set -e
+
+mkdir -p /etc/pharos/pki/etcd
+
+
+if [ ! -e  /usr/local/bin/cfssl ]; then
+    curl -s -L -o /usr/local/bin/cfssl \
+        https://pkg.cfssl.org/R1.2/cfssl_linux-amd64
```
jakolehm: Arch needs to come from variable.

lib/pharos/scripts/configure-etcd.sh
```diff
@@ -0,0 +1,81 @@
...
+    - --advertise-client-urls=https://${PEER_IP}:2379
+    - --listen-peer-urls=https://${PEER_IP}:2380
+    - --initial-advertise-peer-urls=https://${PEER_IP}:2380
+    - --client-cert-auth=true
+    - --peer-client-cert-auth=true
+    - --initial-cluster=${INITIAL_CLUSTER}
+    - --initial-cluster-token=pharos-etcd-token
+    - --initial-cluster-state=new
+
+    image: k8s.gcr.io/etcd-amd64:${ETCD_VERSION}
```
jakolehm: Arch needs to come from variable.

lib/pharos/scripts/configure-etcd.sh
```diff
@@ -0,0 +1,81 @@
...
+
+if [ -e /etc/kubernetes/manifests/pharos-etcd.yaml ]; then
+  exit
+fi
+
+mkdir -p /etc/systemd/system/kubelet.service.d
+cat <<EOF >/etc/systemd/system/kubelet.service.d/5-pharos-etcd.conf
+[Service]
+ExecStart=
+ExecStart=/usr/bin/kubelet --pod-manifest-path=/etc/kubernetes/manifests/ --read-only-port=0 --cadvisor-port=0
```
jakolehm: @SpComb should we lock-down this more?

lib/pharos/scripts/configure-etcd.sh
```diff
@@ -0,0 +1,81 @@
...
+    - --peer-key-file=/etc/kubernetes/pki/etcd/peer-key.pem
+    - --listen-client-urls=https://${PEER_IP}:2379
+    - --advertise-client-urls=https://${PEER_IP}:2379
+    - --listen-peer-urls=https://${PEER_IP}:2380
+    - --initial-advertise-peer-urls=https://${PEER_IP}:2380
+    - --client-cert-auth=true
+    - --peer-client-cert-auth=true
+    - --initial-cluster=${INITIAL_CLUSTER}
+    - --initial-cluster-token=pharos-etcd-token
+    - --initial-cluster-state=new
```
jakolehm: Health check missing.

lib/pharos_cluster.rb
```diff
@@ -10,6 +10,7 @@
 module Pharos
   CRIO_VERSION = '1.9'
   KUBE_VERSION = ENV.fetch('KUBE_VERSION') { '1.9.6' }
-  KUBEADM_VERSION = ENV.fetch('KUBEADM_VERSION') { KUBE_VERSION }
+  KUBEADM_VERSION = ENV.fetch('KUBEADM_VERSION') { '1.10.0' }
```
jakolehm: Not sure yet if 1.10.0 is actually needed here.

lib/pharos/scripts/configure-etcd.sh
SpComb: I don't entirely understand what's going on here... are we not using the `10-kubeadm.conf` dropin onto these nodes?

This configuration is really unsafe, AFAIK it leaves the kubelet API wide open and unauthenticated.

lib/pharos/phases/configure_master.rb
```diff
@@ -252,8 +337,8 @@ def upgrade
         )
 
         cfg = generate_config
-        @ssh.with_tmpfile(cfg.to_yaml) do |tmp_file|
```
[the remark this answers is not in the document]
SpComb: #229

lib/pharos/phases/configure_metrics.rb
```diff
@@ -23,24 +17,9 @@ def initialize(master)
       end
 
       def call
-        configure_metrics_server
```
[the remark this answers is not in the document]
jakolehm: There is no reason to run both metrics-server and heapster now that api-server is configured properly.

lib/pharos/phases/configure_master.rb
```diff
@@ -95,19 +108,26 @@ def install
       end
 
       def generate_config
+        extra_sans = Set.new
+        @config.master_hosts.each do |h|
+          extra_sans += [h.address, h.private_address].compact.uniq
```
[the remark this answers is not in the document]
jakolehm: Yes, I think it would also work. Anyway this logic needs improvement because admin probably should "mask" multiple masters with LB or DNS round-robin.

lib/pharos/scripts/configure-etcd.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/bash
+
+set -e
+
+if [ -e /etc/kubernetes/manifests/pharos-etcd.yaml ]; then
+  exit
```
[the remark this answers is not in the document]
jakolehm: Yeah, not sure if it should be handled here or in `ConfigureEtcd` phase.

jakolehm:
@SpComb @jnummelin PTAL (overall design before I merge parallel stuff)

lib/pharos/phases/configure_cluster_info.rb
```diff
@@ -0,0 +1,27 @@
...
+      end
+
+      def client
+        @client ||= Pharos::Kube.client(@master.address)
+      end
+
+      def call
+        logger.info { "Updating public cluster-info ..." }
+        configmap = client.get_config_map('cluster-info', 'kube-public')
+        kubeconfig = YAML.safe_load(configmap.data['kubeconfig'])
```
kke: Using the `Pharos::YamlFile`:
```ruby
  kubeconfig = Pharos::YamlFile.new(
    StringIO.new(configmap.data['kubeconfig']),
    override_filename: "#{@master.address}:kubeconfig"
   ).load
```

(it should maybe have something like `Pharos::YamlFile.from_string(xyz, "#{master.address}:kubeconfig")` to avoid the StringIO)

lib/pharos/phases/configure_etcd.rb
```diff
@@ -0,0 +1,72 @@
...
+      def initialize(host, config)
+        @host = host
+        @config = config
+        @ssh = Pharos::SSH::Client.for_host(@host)
+      end
+
+      def call
+        sync_ca unless @host == @config.etcd_leader
+
+        logger.info(@host.address) { 'Configuring etcd certs ...' }
```
kke: The `Pharos::Phase` base already sets the address as logger progname (using it this way with the central logger would be better, having a separate logger for each phase makes it thread unsafe)

lib/pharos/phases/configure_etcd.rb
```diff
@@ -0,0 +1,72 @@
...
+          KUBE_VERSION: Pharos::KUBE_VERSION,
+          ARCH: @host.cpu_arch.name,
+          PEER_NAME: "etcd#{peer_index + 1}"
+        )
+      end
+
+      # @return [Array<String>]
+      def initial_cluster
+        i = 0
+        @config.etcd_hosts.map { |h|
```
kke: `with_index` takes base number as arg: 

```ruby
@config.etcd_hosts.map.with_index(1) do |h, i|
```

lib/pharos/phases/configure_master.rb
```diff
@@ -41,9 +42,16 @@ def install?
       end
 
       def upgrade?
-        return false unless Pharos::Kube.config_exists?(@master.address)
+        manifest = File.join(KUBE_DIR, 'manifests', 'kube-apiserver.yaml')
+        file = @ssh.file(manifest)
+        return false unless file.exist?
+        return false if file.read =~ /kube-apiserver-.+:v#{Pharos::KUBE_VERSION}/
```
kke: ruby 2.4 has `.match?` which will skip assigning `$1 ..` globals

jnummelin:
Overall design looks pretty good to me.

Didn't read every single line too carefully as lot of things will still change when fixing all the conflicts. :)

lib/pharos/up_command.rb
```diff
@@ -140,37 +140,38 @@ def validate_hosts(config)
...
 
-    # @param config [Pharos::Config]
-    # @return [Array<Pharos::Configuration::Node>]
-    def worker_hosts(config)
-      config.hosts.select { |h| h.role == 'worker' }
+    # @param features [Pharos::Config]
+    def handle_masters(config)
+      config.master_hosts.each do |master|
+        log_host_header(master)
+        Phases::ConfigureMaster.new(master, config).call
```
SpComb: This is no longer running the `ConfigureHost` phase on non-etcd master nodes?

lib/pharos/phases/configure_master.rb
```diff
@@ -118,36 +134,108 @@ def generate_config
...
+      # Copies certificates from leading master
+      def copy_kube_certs
+        leader = @config.master_leader
+        leader_ssh = Pharos::SSH::Client.for_host(leader)
+        @ssh.exec!("sudo mkdir -p #{KUBE_DIR}/pki")
+        %w(ca.crt ca.key sa.key sa.pub).each do |file|
+          path = File.join(KUBE_DIR, 'pki', file)
+          contents = leader_ssh.file(path).read
+          @ssh.file(path).write(contents)
+          @ssh.exec!("sudo chmod 0400 #{path}")
```
SpComb: TODO: this needs something like `@ssh.file(path).write(contents, perm: 0o400)` to handle it correctly.

lib/pharos/phases/configure_master.rb
```diff
@@ -118,36 +134,108 @@ def generate_config
...
+          extra_sans += [h.address, h.private_address].compact.uniq
+        end
+
+        extra_sans.to_a
+      end
+
+      # Copies certificates from leading master
+      def copy_kube_certs
+        leader = @config.master_leader
+        leader_ssh = Pharos::SSH::Client.for_host(leader)
```
SpComb: Either needs a separate phase for running `kubeadm alpha phase certs` on the leader, or then run the `ConfigureMaster` phase first on the leader, then grab the generated certs somewhere locally before running the phase on the follower nodes to deploy the existing CA/SA certs.

lib/pharos/phases/configure_master.rb
```diff
@@ -118,36 +134,108 @@ def generate_config
...
+        @ssh.file('/etc/pharos/etcd/certificate.pem').write(File.open(@config.etcd.certificate))
+        @ssh.file('/etc/pharos/etcd/certificate-key.pem').write(File.read(@config.etcd.key))
+      end
+
+      def copy_internal_etcd_certs
+        @ssh.exec!('sudo mkdir -p /etc/pharos/pki/etcd')
+        return if @ssh.file('/etc/pharos/pki/etcd/client.pem').exist?
+
+        etcd_peer = @config.etcd_hosts[0]
+        etcd_ssh = Pharos::SSH::Client.for_host(etcd_peer)
```
SpComb: Hmm... isn't this now most likely going to be `localhost`, for the case of a `cluster.yaml` with a single master node?

SpComb: Unless you have separate `etcd` nodes, then the etcd certs are already going to be in place after the `ConfigureEtcd` phase?

Feels like this operation should only be necessary in some specific circumstances, so maybe it shouldn't happen unconditionally.

lib/pharos/phases/configure_kubelet.rb
```diff
@@ -24,6 +24,7 @@ def initialize(host, config)
 
       def call
         configure_cni
+        configure_kubelet_proxy if @host.role == 'worker'
         configure_kube
```
SpComb: I don't understand how this would work... the `configure-kubelet-proxy.sh` starts the kubelet with `/etc/systemd/system/kubelet.service.d/5-pharos-kubelet-proxy.conf` in place, but then this immediately runs `configure-kube.sh` which nukes that file and restarts the kubelet?

I guess the kubelet-proxy pod must remain running even when the kubelet gets restarted with the kubeadm configuration... feels a bit sketchy, but I guess it works?

SpComb: What happens if the node setup fails after the `ConfigureKubelet`, but before the `JoinNode` completes... if the host gets rebooted, then the kubelet-proxy pod is no longer going to be running, but the `configure-kubelet-proxy.sh` is not going to re-configure the kubelet to start it either?

Not the most robust way to run the proxy, I think... should it maybe be something like a systemd service with `rkt` instead, so it's independent of the kubelet?

SpComb: I guess that would just mean that the `Waiting kubelet-proxy to launch on port 6443...` bit is perhaps unnecessary, and it could actually  cause`pharos-cluster up` to break when retrying after errors, whereas just blindly running `kubeadm join` would fix it and allow the kubelet-proxy to start up again?

jakolehm: Port wait is there to avoid timing issues. But it might actually work by blindly running `kubeadm join`. Need to test this after I finish merging master to this branch.

lib/pharos/phase.rb
```diff
@@ -61,5 +61,15 @@ def exec_script(script, vars = {})
     def parse_resource_file(path, vars = {})
       Pharos::YamlFile.new(resource_path(path)).read(vars)
     end
+
+    # @return [Hash]
+    def mem_storage
```
jakolehm: Not sure what is the proper name for this?

lib/pharos/cluster_manager.rb
```diff
@@ -47,12 +47,17 @@ def validate
...
 
-      apply_phase(Phases::ConfigureMaster, config.master_hosts, ssh: true, parallel: false)
+      apply_phase(Phases::ConfigureMaster, config.master_hosts, ssh: true, parallel: false, master: config.master_host)
+      apply_phase(Phases::ConfigureMaster, config.master_hosts, ssh: true, parallel: false, master: config.master_host)
       apply_phase(Phases::ConfigureKubelet, config.worker_hosts, ssh: true, parallel: true) # TODO: also run this phase in parallel for the master nodes, if not doing an upgrade?
-      apply_phase(Phases::ConfigureClient, config.master_hosts, ssh: true, parallel: true)
+      apply_phase(Phases::ConfigureClient, [config.master_host], ssh: true, parallel: false)
 
       # master is now configured and can be used
+      apply_phase(Phases::ConfigureClusterInfo, [config.master_host], ssh: true, parallel: false)
```
SpComb: This is a kube API phase, not a node SSH phase? Needs `master: ..` not `ssh: true`

lib/pharos/phase.rb
SpComb: I'll offer `ClusterContext`

lib/pharos/cluster_manager.rb
jakolehm: Yeah, `ssh` is not required.. but does it need `master`?

lib/pharos/phases/configure_cluster_info.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module Pharos
+  module Phases
+    class ConfigureClusterInfo < Pharos::Phase
+      title "Configure cluster-info"
+
+      def client
+        @client ||= Pharos::Kube.client(@host.address)
```
SpComb: These should always be `@master.address`, unless it were really doing something specific to the kube API on that specific host.

lib/pharos/phase.rb
jakolehm: `def cluster_context` ? Sounds good to me.

lib/pharos/cluster_manager.rb
jakolehm: This somehow feels wrong:
```ruby
apply_phase(Phases::ConfigureClusterInfo, [config.master_host], ssh: false, parallel: false, master: config.master_host)
```

Probably because I need to pass `master_host` twice...

lib/pharos/phase.rb
SpComb: Yeah, `cluster` because it's used to store things that are not node-specific... might also consider some kind of `host_context` later, for things like the `cpu_arch` which are currently set on the `Pharos::Configuration::HostNode` by `ValidateHost`.

lib/pharos/phases/configure_cluster_info.rb
jakolehm: Fixed

lib/pharos/cluster_manager.rb
SpComb: Yes, it kind of is, because those phases are not actually host-specific, but run against the cluster. Some phases run multiple instances, some phase classes only ever run one instance... but the meaning of the `master` is consistent across all of those, it's always used to refer to the kube cluster as a whole.

lib/pharos/cluster_manager.rb
```diff
@@ -47,12 +47,17 @@ def validate
...
 
-      apply_phase(Phases::ConfigureMaster, config.master_hosts, ssh: true, parallel: false)
+      apply_phase(Phases::ConfigureMaster, config.master_hosts, ssh: true, parallel: false, master: config.master_host)
+      apply_phase(Phases::ConfigureMaster, config.master_hosts, ssh: true, parallel: false, master: config.master_host)
       apply_phase(Phases::ConfigureKubelet, config.worker_hosts, ssh: true, parallel: true) # TODO: also run this phase in parallel for the master nodes, if not doing an upgrade?
-      apply_phase(Phases::ConfigureClient, config.master_hosts, ssh: true, parallel: true)
+      apply_phase(Phases::ConfigureClient, [config.master_host], ssh: true, parallel: false)
 
       # master is now configured and can be used
+      apply_phase(Phases::ConfigureClusterInfo, [config.master_host], ssh: false, parallel: false, master: config.master_host)
```
jakolehm: This causes a race condition... `ConfigureMaster` sets master specific ip and it's fixed to localhost here. Not sure how we can actually fix it without updating `kubeadm`?

jakolehm: I mean: if node joins to the cluster between these two phases it will be configured with single master ip.

SpComb: I don't understand, what race condition?

lib/pharos/cluster_manager.rb
```diff
@@ -47,12 +47,17 @@ def validate
     def apply_phases
       apply_phase(Phases::ValidateHost, config.hosts, ssh: true, parallel: true)
       apply_phase(Phases::ConfigureHost, config.hosts, ssh: true, parallel: true)
+      apply_phase(Phases::ConfigureCfssl, config.etcd_hosts, ssh: true, parallel: true)
+      apply_phase(Phases::ConfigureEtcdCa, [config.etcd_host], ssh: true, parallel: false)
+      apply_phase(Phases::ConfigureEtcd, config.etcd_hosts, ssh: true, parallel: true)
 
-      apply_phase(Phases::ConfigureMaster, config.master_hosts, ssh: true, parallel: false)
+      apply_phase(Phases::ConfigureMaster, config.master_hosts, ssh: true, parallel: false, master: config.master_host)
+      apply_phase(Phases::ConfigureMaster, config.master_hosts, ssh: true, parallel: false, master: config.master_host)
```
SpComb: This runs `Phases::ConfigureMaster` twice on the same nodes?  :thinking:

jakolehm: Yes, noticed it too.. something went south with merge. Will fix asap.

lib/pharos/cluster_manager.rb
SpComb: But the `JoinNode` only happens later, well after the `Phases::ConfigureMaster` and `Phases::ConfigureClusterInfo`?

lib/pharos/phase.rb
jakolehm: About arch: yes, it's a bit problematic.. currently we don't really support clusters with different archs.

lib/pharos/config.rb
```diff
@@ -51,5 +48,25 @@ def master_host
...
+      if etcd_hosts.empty?
+        master_hosts
+      else
+        etcd_hosts
+      end
+    end
+
+    # @return [Pharos::Configuration::Host]
+    def master_leader
+      master_hosts[0]
```
SpComb: This is the same as the `master_host`? Looks like this isn't used anywhere anymore.

jakolehm: Yes, removed.

lib/pharos/cluster_manager.rb
jakolehm: True, I'm thinking use cases where nodes are joining via autoscaling group etc.

lib/pharos/cluster_manager.rb
```diff
@@ -47,12 +47,16 @@ def validate
     def apply_phases
       apply_phase(Phases::ValidateHost, config.hosts, ssh: true, parallel: true)
       apply_phase(Phases::ConfigureHost, config.hosts, ssh: true, parallel: true)
+      apply_phase(Phases::ConfigureCfssl, config.etcd_hosts, ssh: true, parallel: true)
+      apply_phase(Phases::ConfigureEtcdCa, [config.etcd_host], ssh: true, parallel: false)
+      apply_phase(Phases::ConfigureEtcd, config.etcd_hosts, ssh: true, parallel: true)
 
-      apply_phase(Phases::ConfigureMaster, config.master_hosts, ssh: true, parallel: false)
+      apply_phase(Phases::ConfigureMaster, config.master_hosts, ssh: true, parallel: false, master: config.master_host)
```
SpComb: The `master: ...` here is problematic, because you can't actually use it for `Pharos::Kube.client` until the `Phases::ConfigureClient` has been run. That's why I left it out of all these earlier phases... we shouldn't need it for the upgrade case anymore?

jakolehm: Yeah, it seems that `ConfigureMaster` is not using master anymore.. will cleanup.

lib/pharos/cluster_manager.rb
SpComb: Hmm.... probably only relevant for the `ConfigureMaster` upgrade case, where kubeadm might replace the `cluster-info` configmap?

SpComb: It looks like kubeadm generates the `cluster-info` configmap based on the `/etc/kubernetes/admin.conf`... so should we be using `api.advertiseAddress = http://localhost:6443`?

jakolehm: I tried to set `api.advertiseAddress = 127.0.0.1` but it did not have any effect.

SpComb: Quick read of the kubeadm code says that it should end up using the `cfg.API.AdvertiseAddress` in both the `/etc/kubernetes/admin.conf` and thus the `cluster-info` configmap... need to test, I guess.

jakolehm: I'll test asap, it would be much better way to solve this.

lib/pharos/phases/configure_master.rb
```diff
@@ -4,10 +4,10 @@ module Pharos
...
       title "Configure master"
-
+      KUBE_DIR = '/etc/kubernetes'
       PHAROS_DIR = '/etc/pharos'
+      SHARED_CERT_FILES = %w(ca.crt ca.key sa.key sa.pub).freeze
       AUTHENTICATION_TOKEN_WEBHOOK_CONFIG_DIR = '/etc/kubernetes/authentication'
-
       AUDIT_CFG_DIR = (PHAROS_DIR + '/audit').freeze
 
       def client
```
SpComb: The `client` and `kubeadm_configmap` are no longer used, maybe remove them?

Although I think we may still want to look at the kubeadm configmap in the future, in order to figure out if things like `etcd` have changed... however, the implementation needs fixing because of #140: this assumes that the `ConfigureClient` phase was run on the same `pharos-cluster` machine earlier, so upgrades would break if you ran `pharos-cluster` from a fresh machine that had never been run before.

lib/pharos/phase.rb
```diff
@@ -61,5 +61,15 @@ def exec_script(script, vars = {})
     def parse_resource_file(path, vars = {})
       Pharos::YamlFile.new(resource_path(path)).read(vars)
     end
+
+    # @return [Hash]
+    def cluster_context
+      self.class.cluster_context
```
kke:
```ruby
def cluster_context
  Pharos::Phase.cluster_context
end

def self.cluster_context
  @cluster_context ||= {}
end
```

On instance level this would have the same effect but without the class variable. `Phases::FooFoo.cluster_context` would get messed up though.

I suppose this ultimately belongs to `ClusterManager` which should pass a self-reference to all phases, sub-managers and whatever it creates.

lib/pharos/cluster_manager.rb
SpComb: Unfortunately it looks like kubeadm overrides `API.AdvertiseAddress: "127.0.0.1"` with the external address associated with the default route:

* https://github.com/kubernetes/kubernetes/blob/v1.9.6/cmd/kubeadm/app/util/config/masterconfig.go#L40-L42
* https://github.com/kubernetes/apimachinery/blob/895b82eadff27f5a6b19d80628a04645823c1e46/pkg/util/net/interface.go#L384

I don't understand why, though... if you're binding to `127.0.0.1`, then the only address that you can connect to is also going to be `127.0.0.1`... it doesn't make any sense to me to default to the external IP address in that case.

The only other caller of `utilnet.ChooseBindAddress` seems to be for `kubeapiserver --advertise-address=0.0.0.0 --insecure-bind-address=127.0.0.1`... in that case it would chose the external IP as the advertise address, but if you pass `--advertise-address=127.0.0.1` then it would respect that: https://github.com/kubernetes/kubernetes/blob/v1.9.6/pkg/kubeapiserver/options/serving.go#L57

Feels like a bug in kubeadm's use of `netutil.ChooseBindAddress` to me...

lib/pharos/phase.rb
SpComb: Yeah, the context could be a `Pharos::Phase#initialize` parameter like `config` if you wanted to make this fancier.

lib/pharos/phases/configure_master.rb
SpComb: This should also fix #140 now

lib/pharos/scripts/configure-kubelet-proxy.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+set -eu
+
+if [ ! -e /usr/bin/kubelet ]; then
+    mkdir -p /etc/systemd/system/kubelet.service.d
+    cat <<EOF >/etc/systemd/system/kubelet.service.d/5-pharos-kubelet-proxy.conf
```
SpComb: The condition here isn't ideal... `pharos-cluster up` gets stuck in the `Waiting etcd to launch ...` / `Waiting kubelet-proxy to launch ...` if you already have the kubelet installed.

lib/pharos/phases/configure_kubelet.rb
jakolehm: Blindly running `kubeadm join` does not work.

lib/pharos_cluster.rb
jakolehm: It's required, we need `ControlPlaneEndpoint`.

lib/pharos/up_command.rb
```diff
@@ -10,7 +10,7 @@ class UpCommand < Pharos::Command
       end
     end
 
-    option '--hosts-from-tf', 'PATH', 'Path to terraform output json' do |config_path|
+    option '--tf-json', 'PATH', 'Path to terraform output json' do |config_path|
```
jakolehm: Note: breaking change

lib/pharos/up_command.rb
```diff
@@ -79,8 +84,10 @@ def build_config(config_hash)
 
       config = Pharos::Config.new(schema)
 
+      # inject api_endpoint to each host object
+      config.hosts.each { |h| h.api_endpoint = config.api&.endpoint }
```
jakolehm: Not really happy with this but we can refactor internals later.

jakolehm:
@SpComb @jnummelin ok for merge?

lib/pharos/up_command.rb
SpComb: Yes, this would be related to #225

SpComb:
> @SpComb @jnummelin ok for merge?

Main issue is that we break existing pharos clusters, because we replace etcd on the master host... is that okay for our 0.x releases?

jakolehm:
> Main issue is that we break existing pharos clusters, because we replace etcd on the master host... is that okay for our 0.x releases?

I'll add separate issue about this. I think it's ok but pharos-cluster should error out if user is trying to upgrade from < 0.6.

lib/pharos/phases/configure_etcd.rb
```diff
@@ -0,0 +1,61 @@
...
+          i += 1
+          "etcd#{i}=https://#{h.peer_address}:2380"
+        }
+      end
+
+      def sync_ca
+        return if @ssh.file(File.join(CA_PATH, 'ca.pem')).exist?
```
SpComb: Good enough but not ideal: if you upload `ca.pem` first but uploading `ca-key.pem` fails for whatever reason, then re-running `pharos-cluster up` is not going to upload the missing `ca-key.pem`. It would be better to attempt to make these things idempotent and robust against random errors.

lib/pharos/phases/configure_master.rb
```diff
@@ -90,16 +87,22 @@ def generate_config
         config = {
           'apiVersion' => 'kubeadm.k8s.io/v1alpha1',
           'kind' => 'MasterConfiguration',
+          'nodeName' => @host.hostname,
           'kubernetesVersion' => Pharos::KUBE_VERSION,
-          'apiServerCertSANs' => [@host.address, @host.private_address].compact.uniq,
+          'api' => {
+            'advertiseAddress' => @host.peer_address,
+            'controlPlaneEndpoint' => '127.0.0.1'
```
SpComb: I thought the `ControlPlaneEndpoint` was validated to be a DNS address, I'm a little surprised that an IP address also works?

SpComb: https://github.com/kubernetes/kubernetes/blob/v1.10.1/cmd/kubeadm/app/util/endpoint.go#L44

lib/pharos/phases/configure_etcd.rb
jakolehm: Should we then just write these files always? I'm pretty sure that there is no speed difference if we check every file vs write every file?

lib/pharos/phases/configure_master.rb
jakolehm: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L132-L143

lib/pharos/phases/configure_etcd.rb
SpComb: Not entirely sure, but eventually we'd need to also be able to update these files, for e.g. cert rotation (5 year validity currently)... I figure this is probably good enough to merge for now, we can later solve this for other files like the `/etc/systemd/system/kubelet.service.d/5-pharos.conf` as well.

Maybe change this to check the `file.exist?` for each cert file if you want to be safe.

lib/pharos/phases/configure_master.rb
SpComb: Yeah, it seems almost accidential that IPv4 addresses also pass that validation... I don't think that's intentional, and e.g. IPv6 addresses won't.

lib/pharos/up_command.rb
kke: 👍

lib/pharos/phases/configure_master.rb
SpComb: I'd consider changing this to `localhost` (although check for `::1` vs `127.0.0.1`)

lib/pharos/up_command.rb
SpComb: Yes, this makes sense because it's no longer just `pharos_hosts`, but also `pharos_api`.

lib/pharos/phases/configure_etcd.rb
kke: Well there is the controversial `@ssh.file.with_existing do |file|` :)

lib/pharos/scripts/configure-etcd-ca.sh
```diff
@@ -0,0 +1,46 @@
...
+    MIRROR="http://mirrors.kernel.org/ubuntu"
+else
+    MIRROR="http://ports.ubuntu.com"
+fi
+
+CFSSL_URL="${MIRROR}/pool/universe/g/golang-github-cloudflare-cfssl/golang-cfssl_1.2.0+git20160825.89.7fb22c8-3_${ARCH}.deb"
+
+if [ ! -e  /usr/bin/cfssl ]; then
+    curl -sL -o /tmp/cfssl.deb ${CFSSL_URL}
+    dpkg -i /tmp/cfssl.deb && rm /tmp/cfssl.deb
```
SpComb: Isn't this already part of the `ConfigureCfssl` phase?

lib/pharos/phases/configure_etcd.rb
SpComb: > Well there is the controversial @ssh.file('/tmp/foofoo').with_existing do |file| :)

Not sure what you mean, this would need `with_not_existing`.

But it needs something much more complex than that if you also want to deal with the file changing... if you replace the CA certs, then you'd also need to restart the daemons using them.

lib/pharos/scripts/configure-etcd.sh
```diff
@@ -0,0 +1,90 @@
...
+      path: /var/lib/etcd
+      type: DirectoryOrCreate
+    name: etcd-data
+  - hostPath:
+      path: /etc/pharos/pki
+      type: DirectoryOrCreate
+    name: etcd-certs
+EOF
+systemctl daemon-reload
+systemctl restart kubelet
```
SpComb: This is unconditionally restarting the kubelet every time you run `pharos-cluster up`? I don't think that should be happening if etcd is already running.

lib/pharos/scripts/configure-kube.sh
```diff
@@ -2,6 +2,22 @@
...
+fi
+
 apt-mark unhold kubelet kubectl kubeadm
-apt-get install -y kubelet=${KUBE_VERSION}-00 kubectl=${KUBE_VERSION}-00 kubeadm=${KUBE_VERSION}-00
-apt-mark hold kubelet kubectl kubeadm
\ No newline at end of file
+apt-get install -y kubelet=${KUBE_VERSION}-00 kubectl=${KUBE_VERSION}-00 kubeadm=${KUBEADM_VERSION}-00
+apt-mark hold kubelet kubectl kubeadm
+
+systemctl restart kubelet
```
SpComb: Same here, this is unconditionally restarting the kubelet?

lib/pharos/scripts/configure-etcd-ca.sh
jakolehm: Yes, will cleanup.

lib/pharos/scripts/configure-etcd.sh
```diff
@@ -0,0 +1,90 @@
...
+    - etcd
+    - --name=${PEER_NAME}
+    - --cert-file=/etc/kubernetes/pki/etcd/server.pem
+    - --key-file=/etc/kubernetes/pki/etcd/server-key.pem
+    - --trusted-ca-file=/etc/kubernetes/pki/ca.pem
+    - --peer-trusted-ca-file=/etc/kubernetes/pki/ca.pem
+    - --data-dir=/var/lib/etcd
+    - --peer-cert-file=/etc/kubernetes/pki/etcd/peer.pem
+    - --peer-key-file=/etc/kubernetes/pki/etcd/peer-key.pem
+    - --listen-client-urls=https://${PEER_IP}:2379
```
SpComb: Consider `--listen-client-urls=https://0.0.0.0:2379` as this prevents using the straightforward `https://localhost:2379/`?

Although if the peer IP is the private IP, then that would also expose etcd on the public IP... so this is safer in that case.

SpComb: Or this can take multiple URLs, so `--listen-client-urls=https://localhost:2379,https://${PEER_IP}:2379` would work nicely

lib/pharos/phase.rb
```diff
@@ -61,5 +61,15 @@ def exec_script(script, vars = {})
...
+    end
+
+    # @return [Hash]
+    def self.cluster_context
+      @@cluster_context ||= {} # rubocop:disable Style/ClassVars
```
jnummelin: Should this be some sort of thread safe hash now that phases can execute in parallel?

jakolehm: Probably, but it might be a bad idea anyway to write same key from multiple parallel phases?

SpComb: I've understood that ruby's standard `Hash` is safe for use in multiple MRI threads, so I'm not entirely sure what something like [`Concurrent::Hash`](http://ruby-concurrency.github.io/concurrent-ruby/Concurrent/Hash.html) actually does.

In terms of race conditions within our code, only one phase/thread should write to the `cluster_context`, and then following phases/threads can read from it. If we were to inappropriately change something like the `ConfigureMaster` to run in parallel, then the races would be in the actual  `copy_kube_certs`/`cache_kube_certs` methods, not in the hash accesses.

kke: https://github.com/ruby-concurrency/concurrent-ruby/blob/master/lib/concurrent/hash.rb#L15-L16

To me it looks like that is all there is in `Concurrent::Hash` for MRI.

kke: I'd still change that to `@cluster_context` and change the instance-level accessor to point to `Pharos::Phase.cluster_context`, because why not. Class variables are weird.

jnummelin: well, I wasn't really 100% sure if standard Hash was safe to use in many threads, but it seems to be. at least in our cases right now.

kke: We can do that stuff to the `Pharos::Kube::Session` (or some `Pharos::Session` thing) once this and #225 is in.

lib/pharos/scripts/configure-etcd.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
+
+set -e
+
+if [ ! -e /etc/kubernetes/kubelet.conf ]; then
+  mkdir -p /etc/systemd/system/kubelet.service.d
+  cat <<EOF >/etc/systemd/system/kubelet.service.d/5-pharos-etcd.conf
```
SpComb: BTW: calling these `5-pharos-*.conf` is super confusing, because that implies that they would sort before the `10-kubeadm.conf`. Instead, they actually sort after, so they should really be called something like `50-pharos-*.conf`:

```
$ ls -l /etc/systemd/system/kubelet.service.d/
total 8
-rw-r--r-- 1 root root 937 Mar 27 15:50 10-kubeadm.conf
-rw-r--r-- 1 root root 243 Apr 13 14:09 5-pharos.conf
```

SpComb: And renaming these to `05-pharos.conf` would probably break it, they need to come after the `10-kubeadm.conf`.

lib/pharos/scripts/configure-kube.sh
SpComb: The `apt-get install kubeadm` will already trigger `systemctl restart kubelet` as part of the postinst script during an initial install.

I suppose this restart is only needed if `kubeadm` is already installed, but the `/etc/kubernetes/kubelet.conf` was missing during the previous `configure-kubelet-proxy.sh` / `configure-etcd.sh` phase, so we need to restart the `kubelet` without the temporary dropins?

Maybe this should be part of the `if [ -e /etc/systemd/system/kubelet.service.d/5-pharos-*.conf ]` blocks then?

SpComb: Maybe this should be part of the `if [ -e /etc/systemd/system/kubelet.service.d/5-pharos-*.conf ]` blocks?

Although for the initial install case, that would then end up restarting the `kubelet` in the default configuration before the `apt-get install kubeadm`, which again exposes the unauthenticated kubelet API by default... :unamused:

lib/pharos/scripts/configure-etcd.sh
jakolehm: Created an issue about this: https://github.com/kontena/pharos-cluster/issues/242

lib/pharos/phases/configure_master.rb
jakolehm: Changed to `localhost`.

lib/pharos/scripts/configure-etcd.sh
jakolehm: Fixed.

lib/pharos/scripts/configure-etcd.sh
```diff
@@ -78,8 +65,19 @@ spec:
       type: DirectoryOrCreate
     name: etcd-certs
 EOF
-systemctl daemon-reload
-systemctl restart kubelet
+
+if [ ! -e /etc/kubernetes/kubelet.conf ]; then
```
SpComb: Good, no more unnecessary kubelet restarts, but now we're back to the case where `pharos-cluster up` won't work after a `kubeadm reset`... however, that's probably okay. Using `kubeadm reset && apt-get purge kubelet kubeadm kubectl` is enough for testing (although that leaves behind some `/etc/systemd/system/kubelet.service.d/*.conf` bits, but that's okay as long as the `5-pharos-etcd.conf` here overrides them).

This condition can probably be changed back to the original `if [ ! -e /usr/bin/kubelet ]`, which makes more sense here. I don't think we can reasonably expect this to behave correctly in all possible kubelet configuration states because of the extra complexity in the two-stage kubelet configuration.

SpComb: Makes more sense to write this as it used to be: `if [ ! -e /usr/bin/kubelet ]`

lib/pharos/scripts/configure-kubelet-proxy.sh
```diff
@@ -37,8 +24,18 @@ spec:
   hostNetwork: true
 EOF
 
-systemctl daemon-reload
-systemctl restart kubelet
+if [ ! -e /etc/kubernetes/kubelet.conf ]; then
```
SpComb: Makes more sense to write this as it used to be: `if [ ! -e /usr/bin/kubelet ]`

jakolehm:
> I don't like the two-stage kubelet configuration complexity for getting the static etcd/kube-proxy pods running

I agree that it's a bit complex but I didn't really come up with anything better (without installing etcd to host... which would take us to packaging etcd).

lib/pharos/scripts/configure-kubelet-proxy.sh
SpComb: More sense = the block condition should be something that's satisfied by executing the block

Right now this is some side effect of the `JoinNode` phase, which happens after the `configure-kube.sh`... spooky action at a distance.

PR #225, opened 2018-04-10
Refactor Pharos::Kube

kke:


kke:
Aaand it does not do what it was expected to do, which I guess was something like:
```ruby
Kube.new(master.address).stack('foofoo').apply
```

kke:
image

That approach, and this too, and also the original, share a shortcoming which is that they parse/build/whatever the same resource multiple times.

For the parallelism it might make sense to instead have something like:

```ruby
stack = Stack.new('foofoo')
hosts.each do |host|
  magic_parallel_box(host).post { stack.apply(host) }
end
```

kke:
Now with `Pharos::Kube.session('10.1.2.3').stack('foofoo').apply`

The session and client pools rely heavily on `Kubeclient` being threadsafe.

The session, stack and resource are immutable so I don't think they should pose any parallel problems theirself.

If `Kubeclient` is not confidently thread friendly, making the `Pharos::Kube.client` act as some sort of blocking connection pool thing could be a good next step?

kke:
Or should it perhaps be something like `Pharos::Kube.session(config).masters.each { |m| m.stack('foofoo').apply }`?

SpComb:
> If Kubeclient is not confidently thread friendly, making the Pharos::Kube.client act as some sort of blocking connection pool thing could be a good next step?

Either a threadsafe + blocking connection pool like `Pharos::Kube.client do |client| ... end`, or have the `PhaseManager` in #111 pass each phase a different `Pharos::Kube::Session` as a parameter... I would prefer the later.

kke:
The later does not ensure that the `Pharos::Kube.client`s are only used by a single thread yet

SpComb:
> The later does not ensure that the Pharos::Kube.clients are only used by a single thread yet

Yeah, having `Pharos::PhaseManager` pass in the session would still require a connection pool, where the session gets returned to the connection pool once the phase is done... because the kube session for something like `LabelNodes` is not node-specific as for the SSH sessions.

Something like a parallel phase for all the master nodes could use a session-per-node without needing any kind of connection pooling, but I don't think we actually have any phase like that.

kke:
Hmm, just noticed something:

https://github.com/kontena/pharos-cluster/blob/dfca39b5b04a24f22a7645e59f8ed3e26a8c4019/lib/pharos/kube.rb#L15-L29

The `Pharos::Kube.client` memoizes only using the version.

So:

```ruby
> client = Pharos::Kube.client('10.2.3.4', 'v1')
> client2 = Pharos::Kube.client('127.0.0.1', 'v1')
> client == client2
 => true
```

SpComb:
> The Pharos::Kube.client memoizes only using the version.

It's not *necessarily* a bug... all of the master nodes form a cluster and the kube API endpoints are equivalent, so this will behave just fine. It doesn't actually matter which master address you use to talk to the kube API.

Still a good sign of the `Pharos::Kube.foo(host, ...)` really being in need of some refactoring.

lib/pharos/kube/stack.rb
```diff
@@ -0,0 +1,72 @@
...
+      # @return [Array<Kubeclient::Resource>]
+      def prune(checksum)
+        pruned = []
+        @session.api_groups.each do |api_group|
+          client = @session.resource_client(api_group.preferredVersion.groupVersion)
+          client.entities.each do |type, meta|
+            next if type.end_with?('_review')
+            objects = client.get_entities(type, meta.resource_name, label_selector: "#{RESOURCE_LABEL}=#{@name}")
+            objects.map { |obj| Resource.new(@session, obj) }.each do |obj|
+              next unless obj.metadata.annotations.nil? || obj.metadata.annotations[RESOURCE_ANNOTATION] != checksum
```
SpComb: This form seems to have about four negatives... unable to figure out if this does the right thing, or if it deletes all the wrong resources instead. The old `select { ... }` form was much clearer TBH.

jakolehm: I agree that this is too complex.

lib/pharos/kube/resource.rb
```diff
@@ -0,0 +1,86 @@
...
+    class Resource
+      def initialize(session, resource, vars = {})
+        @resource =
+          case resource
+          when Kubeclient::Resource
+            resource
+          when Hash
+            Kubeclient::Resource.new(resource)
+          else
+            Kubeclient::Resource.new(Pharos::YamlFile.new(resource).load(vars))
```
SpComb: This goes back to the excessive type confusion thing again... IMO it's not a good design to be way too flexible in terms of the variety of different types accepted.

This is demonstrated by the `vars` getting totally ignored unelss you pass something that's a `String` :-1: 

These might make sense as separate class methods, like `def self.resource_from_file(file, vars = {})`

lib/pharos/kube/stack.rb
```diff
@@ -0,0 +1,79 @@
...
+            group_client.get_entities(type, meta.resource_name, label_selector: "#{RESOURCE_LABEL}=#{@name}")
+          end
+
+          prunables = objects.select do |obj|
+            annotations = obj.metadata.annotations
+            annotations.nil? || annotations[RESOURCE_ANNOTATION] != checksum
+          end
+
+          prunables.each do |obj|
+            pruned << obj if @session.resource(obj).delete
```
kke: The previous result included also the resources where delete failed due to resource not found.

Not sure if intentional.

lib/pharos/kube/resource.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+require 'kubeclient'
+require 'pathname'
+
+module Pharos
+  module Kube
+    class Resource
+      def initialize(session, resource)
```
SpComb: Needs YARDocs

lib/pharos/kube/resource.rb
```diff
@@ -0,0 +1,86 @@
...
+            kind: 'DeleteOptions',
+            apiVersion: 'v1',
+            propagationPolicy: 'Foreground'
+          )
+        end
+      rescue Kubeclient::ResourceNotFoundError
+        false
+      end
+
+      def metadata
```
SpComb: What about accessing the actual `spec` or other data on the `@resource`?

It seems that `Pharos::Kube::Resource` is currently only usable for writing resources to kube, but not reading them back from kube. This resource wrapper currently doesn't work for e.g. something like `ConfigureNetwork#ensure_passwd` or the `Pharos::Kube::CertManager` needs to get at the resource secrets and status... 

Not entirely sure what the fix should be, though, because this mixes up the verb actions like `get` or `delete` with the attributes like `data`, `spec` or `status`... maybe those could be accessed via `def []` or something?

SpComb: Also, I don't think we can just write out accessors for all the different resource type fields either... different resource types will have different fields in addition to the generic `kind`, `apiVersion` and `metadata`... and custom resources types can have any schema at all.

kke: The `.get` could wrap the responses inside a `Pharos::Kube::Resource`.

Then you could do `x = @session.resource(foofoo).get; x.delete if x.foofoo.nil?`

It could have `[]` and `method_missing` which would still allow access to `resource[:get]` in case you need `resource.get`.

Some orms have `.attributes`, that could work too.

lib/pharos/kube/resource.rb
```diff
@@ -0,0 +1,136 @@
...
+        end
+      rescue Kubeclient::ResourceNotFoundError
+        false
+      end
+
+      # Access an attribute of the resource
+      # @example
+      #   resource[:metadata].labels ||= {}
+      def [](key)
+        @resource.send(key)
```
SpComb: Doesn't `[:foo]` also work for `Kubeclient::Resource`?

lib/pharos/kube/resource.rb
```diff
@@ -0,0 +1,136 @@
...
+      end
+
+      # Set a value to an attribute on the resource
+      def []=(key, value)
+        @resource.send("#{key}=", value)
```
SpComb: Would `@resource[key] = value` also work?

lib/pharos/kube/resource.rb
```diff
@@ -0,0 +1,136 @@
...
+      end
+
+      # Like Hash#fetch
+      # @example
+      #   resource.fetch(:metadata, {}).fetch(:foo, nil)
+      #   metadata= resource.fetch(:metadata) { resource.metadata = {} }
+      #   metadata['labels'] = ['foo']
+      # @param key [String,Symbol]
+      # @param default [Object] value to use if nil returned
+      def fetch(key, default = nil)
```
SpComb: Hmm... `get` and `fetch` both sound the same, but do entirely different things :(

kke: `get` sounds like http get, `fetch` sounds and acts like `hash.fetch`

SpComb: I guess `Hash` doesn't have a `#get` method.

lib/pharos/kube/resource.rb
kke: Seems so.

```ruby
irb(main):002:0> k = Kubeclient::Resource.new
=> #<Kubeclient::Resource>
irb(main):003:0> k.metadata = {}
=> {}
irb(main):004:0> k.metadata
=> #<Kubeclient::Resource>
```

=>

```ruby
irb(main):005:0> k = Kubeclient::Resource.new
=> #<Kubeclient::Resource>
irb(main):008:0> k[:metadata] = {}
=> {}
irb(main):009:0> k.metadata
=> #<Kubeclient::Resource>
```

kke:
> Still not entirely happy with the Pharos::Kube::Resource interface

It does have a ridiculous amount of ways to achieve the same thing:


```ruby
r = @session.resource(foo: { bar: 'baz' })
r.foo.bar
r[:foo][:bar]
r.dig(:foo, :bar) # <-- just noticed Kubeclient::Resource has .dig
r.fetch(:foo)[:bar] # but not r.fetch(:foo).fetch(:bar) <-- fetch is bad.
r.attributes.foo.bar
r.attributes[:foo].bar
```

lib/pharos/kube/stack.rb
```diff
@@ -0,0 +1,88 @@
...
+
+      # @param checksum [String]
+      # @return [Array<Kubeclient::Resource>]
+      def prune(checksum)
+        pruned = []
+
+        @session.api_groups.each do |api_group|
+          group_client = @session.resource_client(api_group.preferredVersion.groupVersion)
+
+          entities = group_client.entities.reject { |type, _| type.end_with?('_review') }
```
kke: The entities are memoized in `Pharos::Kube::Client`, but I suppose they don't change much. It does have a short window of possible race condition.

https://github.com/kontena/pharos-cluster/blob/caae8bf112e89ce42d007285e3b7c03accdf145b/lib/pharos/kube/client.rb#L9-L14

jakolehm:
Conflicts...

kke:
There may be something wrong with the pruning here, I think it raises http 404.

kke:
```ruby
if metadata.selfLink
  api_group = metadata.selfLink.split("/")[1]
  path = metadata.selfLink.gsub("/#{api_group}/#{@api_version}", '')
  @client.rest_client[path].delete
else
```

I wonder if gsubbing the selflink is actually needed, it would make sense that the returned selflink is already valid?

kke:
Final AFAIK.

lib/pharos/kube/session.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+require_relative '../kube'
+
+module Pharos
+  module Kube
+    class Session
+      # @param host [String,Phaross::Configuration::Host]
+      def initialize(host)
+        @host = host.respond_to?(:address) ? host.address : host
```
jakolehm: Needs to be `host.api_address`

PR #228, opened 2018-04-11
Pretty formatting for the node list when running parallel

kke:
Applies the fancy formatting from #111 review comment to the node list in phase headers:

```ruby
hosts = ['10.2.3.4', '10.2.3.56', '192.168.100.4', '192.168.23.254', '172.15.23.51', '83.123.12.2', '100.100.100.100', '32.2.1.2']

Pharos::ClusterManager.new('foo', config_content: 'foo').format_hosts(hosts)
=> "8 hosts"

Pharos::ClusterManager.new('foo', config_content: 'foo').format_hosts(hosts.first(5))
=> "10.2.* (2 hosts) 192.168.* (2 hosts) 172.15.23.51"

Pharos::ClusterManager.new('foo', config_content: 'foo').format_hosts(hosts.first(3))
=> "10.2.3.4 10.2.3.56 192.168.100.4"
```

### DIFF
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -6,7 +6,7 @@ module Pharos
 
     # XXX: should be in some kind of output helper?
     def pastel
-      @pastel ||= Pastel.new(enabled: $stdout.tty?)
+      @pastel ||= Pastel.new
     end
 
     attr_reader :config
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -6,7 +6,7 @@ module Pharos
 
     # XXX: should be in some kind of output helper?
     def pastel
-      @pastel ||= Pastel.new
+      @pastel ||= Pastel.new(enabled: $stdout.tty?)
     end
 
     attr_reader :config
@@ -66,7 +66,7 @@ module Pharos
     end
 
     def apply_phase(phase_class, hosts, **options)
-      puts pastel.cyan("==> #{phase_class.title} @ #{hosts.join(' ')}")
+      puts pastel.cyan("==> #{phase_class.title} @ #{format_hosts(hosts)}")
 
       phase_manager.apply(phase_class, hosts, **options)
     end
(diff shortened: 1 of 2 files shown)


lib/pharos/cluster_manager.rb
```diff
@@ -82,5 +82,22 @@ def apply_addons
...
+
+    def format_hosts(hosts)
+      return hosts.join(' ') if logger.debug?
+      case hosts.size
+      when 0
+        '<unknown>'
+      when 1..3
+        hosts.join(' ')
+      else
+        grouped = hosts.group_by { |h| h.split('.').first(2).join('.') + '.*' }
```
SpComb: Not convinced that grouping by IPv4 `/16` is the right thing to do, but I don't know what the best way to deal with very large numbers of hosts would be, in terms of the CLI output... perhaps some kind of spinner-like thing tied into the parallel execution threadpool showing which nodes are currently processing, and how many failed etc.

kke: Probably it's all very irrelevant to the user. I don't think the IPs are interesting at all and if anything fails, it should probably just abort asap.

It could just say "on worker nodes".

SpComb: The more nodes you have, the more likely it is for something to just fail randomly. I think aborting ASAP on errors is actually the wrong thing to do.

SpComb: > It could just say "on worker nodes".

Yes, that's also a good option.

jakolehm: > The more nodes you have, the more likely it is for something to just fail randomly. I think aborting ASAP on errors is actually the wrong thing to do.

Yes, it's most likely too risky to abort everything if one node throws an error.

kke:
2cool4school

SpComb:
I think eventually the phases should be tagged with what what types of nodes they run on, and then the output could be something along the lines of `Apply ConfigureKubelet @ 5 worker nodes`.

That would require having `apply_nodes` handle the node selection based on some role matching, instead of just passing the list of nodes.

PR #231, opened 2018-04-11
Fix --no-color

kke:
Fixes `--no-color` / `pharos-cluster up > log.txt` color on/off switching that got broken in #111

### DIFF
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -4,16 +4,12 @@ module Pharos
   class ClusterManager
     include Pharos::Logging
 
-    # XXX: should be in some kind of output helper?
-    def pastel
-      @pastel ||= Pastel.new
-    end
-
     attr_reader :config
 
-    def initialize(config, config_content:)
+    def initialize(config, config_content:, **opts)
       @config = config
       @config_content = config_content
+      @pastel = opts.fetch(:pastel) { Pastel.new }
     end
 
     # @return [Pharos::SSH::Manager]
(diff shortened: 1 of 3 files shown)


lib/pharos/cluster_manager.rb
```diff
@@ -4,16 +4,12 @@ module Pharos
...
-      @pastel ||= Pastel.new
-    end
-
     attr_reader :config
 
-    def initialize(config, config_content:)
+    def initialize(config, config_content:, **opts)
       @config = config
       @config_content = config_content
+      @pastel = opts.fetch(:pastel) { Pastel.new }
```
SpComb: Could this go into some kind of `OutputHelper', so that it doesn't need to be an option here?

We're going to eventually need something similar in the `Pharos::Logging` as well.

SpComb: I guess the `Pharos::Command` `--no-color` is going to need to set some global flag like `Pharos::OutputHelper.enabled= false` in order to work.

kke: Everything could go through logger. It would make controlling the verbosity, colorfulness, etc quite nice, the formatter could be changed to include timestamps, progname, etc. Also logger has some sort of mutex thing, so it shouldn't start writing a line in the middle of another.

The `Pharos::Logging` could have its own Pastel and have some formatting stuff:

```ruby
logger.info logger.header('Sharpening tools')
```

kke: Btw, something like this would work without having to include stuff:

```ruby
module Pharos
  Header = Pastel.new.green.detach
end
```

```ruby
module Pharos
  module Terraform
    class JsonParser
      def initialize
        puts Header.('this is super green')
      end
    end
end
```

lib/pharos/ssh/remote_command.rb
```diff
@@ -115,7 +115,7 @@ def run
       def initialize_debug
         if self.class.debug?
           @debug = true
-          @pastel = Pastel.new
+          @pastel = Pastel.new($stdout.tty?)
```
jakolehm: Why this fix is directly with `$stdout`?

kke: Because it's not aware of the command.

More reasons to have a global `configure`able output helper thing.

PR #232, opened 2018-04-11
Upgrade weave 2.3.0

### LINKED ISSUE #224: Upgrade weave net to 2.3.0
https://github.com/weaveworks/weave/releases/v2.3.0

SpComb:
Fixes  #224 Upgrade Weave Net to 2.3.0
Revert #189 `WEAVE_STATUS_ADDR` changes and update docs to match

The weave router now only exposes `/metrics` on port 6782, leaving it in the same category of prometheus metrics endpoints that we already leave exposed.

### DIFF
--- a/lib/pharos/resources/weave/daemon-set.yml.erb
+++ b/lib/pharos/resources/weave/daemon-set.yml.erb
@@ -38,9 +38,6 @@ spec:
                 secretKeyRef:
                   name: weave-passwd
                   key: weave-passwd
-            # defaults to 0.0.0.0:6782, limit to block public access
-            - name: WEAVE_STATUS_ADDR
-              value: 127.0.0.1:6782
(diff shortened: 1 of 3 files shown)


lib/pharos/resources/weave/daemon-set.yml.erb
```diff
@@ -38,9 +38,6 @@ spec:
                 secretKeyRef:
                   name: weave-passwd
                   key: weave-passwd
-            # defaults to 0.0.0.0:6782, limit to block public access
-            - name: WEAVE_STATUS_ADDR
-              value: 127.0.0.1:6782
```
SpComb: Alternatively we could switch to `WEAVE_METRICS_ADDR=127.0.0.1:6782` to lock this down even further... would need to know how we intend to scrape Prometheus metrics from the nodes.

PR #234, opened 2018-04-13
Update nginx-ingress to 0.12.0

### LINKED ISSUE #216: Upgrade ingress-nginx to 0.12.0

jnummelin:
fixes #216

### DIFF
--- a/lib/pharos/addons/ingress_nginx.rb
+++ b/lib/pharos/addons/ingress_nginx.rb
@@ -4,7 +4,7 @@ module Pharos
   module Addons
     class IngressNginx < Pharos::Addon
       name 'ingress-nginx'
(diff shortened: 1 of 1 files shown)


jakolehm:
No changes to yaml files?

jnummelin:
Nope. Unless there's some new things we'd like to config. Looking at the changelog `--enable-dynamic-configuration` might make sense to configure as it's disabled by default.

@jakolehm Did you have some specific things in mind we should change in the configs?

jakolehm:
> @jakolehm Did you have some specific things in mind we should change in the configs?

No, just wanted to double-check that configs didn't change.

jnummelin:
I'll make separate issue for enabling the live-reload

PR #236, opened 2018-04-13
Secrets encryption config for api server

### LINKED ISSUE #177: Enable secrets encryption
https://kubernetes.io/docs/tasks/administer-cluster/encrypt-data/

jnummelin:
A bit rough on the edges still.

Need to figure out proper way to get the generated keys from the phase up to the command and output there. Outputting this kind of info to user from phases is really ugly

After this enabled, the secrets get encrypted in etcd like:
```
/ # ETCDCTL_API=3 etcdctl get /registry/secrets/default/secret1
/registry/secrets/default/secret1
k8s:enc:aescbc:v1:key1:??[?Ӈ&?H???ݻ?lPw?=?*?????Y??3?M?MDNKuO??<?S?~x?O??43??~?? ??Lq???؏C????Gda=#ܶ?^S??B	???C?I⏧?M???E??Ҕ??,?y???(??g?[????9??$S?}
```

fixes #177

### DIFF
--- a/lib/pharos/phases/configure_master.rb
+++ b/lib/pharos/phases/configure_master.rb
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
 require "base64"
 
 module Pharos
@@ -42,7 +43,6 @@ module Pharos
       end
 
       def install
-
         configure_kubelet
 
         cfg = generate_config
@@ -134,17 +134,15 @@ module Pharos
 
         # Configure audit related things if needed
         configure_audit_webhook(config) if @config.audit&.server
-        
+
         # Set secrets config location and mount it to api server
-        config['apiServerExtraArgs'].merge!(
-          "experimental-encryption-provider-config" => SECRETS_CFG_FILE
-        )
+        config['apiServerExtraArgs']['experimental-encryption-provider-config'] = SECRETS_CFG_FILE
         config['apiServerExtraVolumes'] << {
           'name' => 'k8s-secrets-config',
           'hostPath' => SECRETS_CFG_DIR,
           'mountPath' => SECRETS_CFG_DIR
         }
-        
+
         config
       end
 
@@ -330,18 +328,14 @@ module Pharos
 
       def configure_secrets_encryption
         # Generate keys
-        # TODO: Put the keys into some "context" and dump to user in the up command
-        # Only at that level we should use pastel etc. IMO
-        puts "==> Generating secret encryption keys"
-        puts "NOTE: These keys are highly sensitive, keep them safe!!!"
         key1 = Base64.strict_encode64(SecureRandom.random_bytes(32))
         key2 = Base64.strict_encode64(SecureRandom.random_bytes(32))
-        puts "Key1: #{key1}"
-        puts "Key2: #{key2}"
-        puts "NOTE: These values are base64 encoded already!"
+        cluster_context['secrets_encryption'] = {
+          'key1' => key1,
+          'key2' => key2
+        }
 
-        @ssh.file(SECRETS_CFG_FILE).write(
-          parse_resource_file('secrets/encryption-config.yml.erb', key1: key1, key2: key2))
+        @ssh.file(SECRETS_CFG_FILE).write(parse_resource_file('secrets/encryption-config.yml.erb', key1: key1, key2: key2))
       end
 
       def upgrade
--- a/lib/pharos/phases/configure_master.rb
+++ b/lib/pharos/phases/configure_master.rb
@@ -75,7 +75,6 @@ module Pharos
 
         copy_kube_certs
 
-        @ssh.exec!("sudo mkdir -p #{SECRETS_CFG_DIR}")
         configure_secrets_encryption
 
         logger.info { "Initializing control plane ..." }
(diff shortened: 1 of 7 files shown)


lib/pharos/phases/configure_master.rb
```diff
@@ -70,6 +75,9 @@ def install
 
         copy_kube_certs
 
+        @ssh.exec!("sudo mkdir -p #{SECRETS_CFG_DIR}")
+        configure_secrets_encryption
```
SpComb: The `generate_config` gets used for both `install` and also `upgrade`, so the upgrade path is currently broken, leading to apiserver errors:

```
I0417 07:56:33.343388       1 server.go:121] Version: v1.9.6
error opening encryption provider configuration file "/etc/pharos/secrets-encryption/config.yml": open /etc/pharos/secrets-encryption/config.yml: no such file or directory
```

Depending on how kube itself handles migration to the encryption provider, this could work by also conditionally generating the `.../secrets-encryption/config.yml` as part of the `upgrade`?

lib/pharos/up_command.rb
```diff
@@ -104,6 +104,13 @@ def configure(config, config_content:)
 
       puts pastel.green("==> Starting to craft cluster ...")
       manager.apply_phases
+      if Phase.cluster_context['secrets_encryption']
+        puts pastel.red("==> Configured secrets encryption with keys:")
+        puts pastel.red("    Key1:#{Phase.cluster_context['secrets_encryption']['key1']}")
+        puts pastel.red("    Key2:#{Phase.cluster_context['secrets_encryption']['key2']}")
+        puts pastel.red("    Keep these safe!")
```
SpComb: Doesn't sound reasonable to expect the user to copy-paste these somewhere from the output.

IMO in practice this would be more like: When backing up the kube master node, you must also backup the secrets encryption keys in `/etc/pharos/secrets-encryption/config.yml` in addition to the etcd database in `/var/lib/etcd`.

SpComb: And that advice should probably go into our "Backing up and restoring your kube cluster" docs, rather than into the `pharos-cluster up` output...

lib/pharos/phases/configure_master.rb
```diff
@@ -310,6 +326,18 @@ def upload_authentication_token_webhook_certs(webhook_config)
...
+      def configure_secrets_encryption
+        # Generate keys
+        key1 = Base64.strict_encode64(SecureRandom.random_bytes(32))
+        key2 = Base64.strict_encode64(SecureRandom.random_bytes(32))
+        cluster_context['secrets_encryption'] = {
+          'key1' => key1,
+          'key2' => key2
+        }
+
+        @ssh.file(SECRETS_CFG_FILE).write(parse_resource_file('secrets/encryption-config.yml.erb', key1: key1, key2: key2))
```
SpComb: For something like plaintext secrets encryption keys in particular, it would be a good idea to keep this file non-world-readable on the master host.

The `/var/lib/etcd/member/` directory seems to be 0700 already.

lib/pharos/phases/configure_master.rb
```diff
@@ -70,6 +75,9 @@ def install
 
         copy_kube_certs
 
+        @ssh.exec!("sudo mkdir -p #{SECRETS_CFG_DIR}")
```
SpComb: `sudo install -d -m 0700 #{SECRETS_CFG_DIR}`

lib/pharos/phases/configure_master.rb
jnummelin: Kube itself does not do any migration automatically. It is doable with `kubectl` by "re-writing" the secrets with new encryption config in place.

I'll see if there's some way to make it work in upgrade too.

SpComb: Is it able to read both encrypted and existing unencrypted secrets when configured with `--experimental-encryption-provider-config=`?

I don't think we necessarily need any explicit migration to rewrite the secrets, but it would be good to handle configuration changes like this correctly on upgrades. The same comment about breaking the configuration on upgrades seems to also apply to the auth/audit webhooks, except those are `cluster.yml` changes which we don't support (yet?), whereas this is a new configuration that gets enabled by default.

jnummelin: AFAIK it's able to read both encrypted and existing unencrypted secrets, needs maybe `identity` provider also configured as the last provider:

> When reading resources from storage each provider that matches the stored data attempts to decrypt the data in order.

lib/pharos/phases/configure_master.rb
jakolehm: Why `mkdir` here and not in `configure_secrets_encryption`?

lib/pharos/up_command.rb
jakolehm: I agree that output is not the best place for these. Maybe just hint where encryption is stored?

lib/pharos/phases/configure_master.rb
jnummelin: It'll move there in fixup commits soon :)

lib/pharos/up_command.rb
jnummelin: Right, I'll remove it completely from the output.

Not sure if it makes sense to "hint" on backing up this as we don't make those hints for anything else.

But in the end, our backup automation should take this in too.

lib/pharos/phases/configure_master.rb
```diff
@@ -310,14 +325,39 @@ def upload_authentication_token_webhook_certs(webhook_config)
         @ssh.file(PHAROS_DIR + '/token_webhook/key.pem').write(File.open(File.expand_path(webhook_config[:user][:client_key]))) if webhook_config[:user][:client_key]
       end
 
+      def secrets_encryption_exist?
+        logger.debug { "Checking if secrets encryption is already configured" }
```
jakolehm: Missing ` ...`? And should it say something if already configure?

lib/pharos/phases/configure_master.rb
```diff
@@ -310,14 +325,39 @@ def upload_authentication_token_webhook_certs(webhook_config)
...
+        cfg_file = @ssh.file(SECRETS_CFG_FILE)
+        return false unless cfg_file.exist?
+        return false unless cfg_file.read.match?(/aescbc/)
+
+        true
+      end
+
+      def configure_secrets_encryption
+        logger.info { "Creating secrets encryption configuration" }
```
jakolehm: Missing ` ...` .

lib/pharos/phases/configure_master.rb
jnummelin: fixed

lib/pharos/phases/configure_master.rb
jnummelin: fixed

PR #247, opened 2018-04-16
Handle 0.5 -> 0.6 migration

### LINKED ISSUE #239: Handle upgrade to 0.6.0
Probably we should just error out and say that upgrade is not possible because of etcd clustering changes.

jakolehm:
Handles upgrade from 0.5.0 to 0.6.0.

Fixes #239

### DIFF
--- a/lib/pharos/phases/configure_master.rb
+++ b/lib/pharos/phases/configure_master.rb
@@ -48,32 +48,17 @@ module Pharos
           copy_external_etcd_certs
         end
 
-        if @config.audit&.server
-          logger.info { "Pushing audit configs to master ..." }
-          @ssh.exec!("sudo mkdir -p #{AUDIT_CFG_DIR}")
-          @ssh.file("#{AUDIT_CFG_DIR}/webhook.yml").write(
-            parse_resource_file('audit/webhook-config.yml.erb', server: @config.audit.server)
-          )
-          @ssh.file("#{AUDIT_CFG_DIR}/policy.yml").write(parse_resource_file('audit/policy.yml'))
-        end
+        push_audit_config if @config.audit&.server
 
         # Generate and upload authentication token webhook config file if needed
-        if @config.authentication&.token_webhook
-          webhook_config = @config.authentication.token_webhook.config
-          logger.info { "Generating token authentication webhook config ..." }
-          auth_token_webhook_config = generate_authentication_token_webhook_config(webhook_config)
-          logger.info { "Pushing token authentication webhook config ..." }
-          upload_authentication_token_webhook_config(auth_token_webhook_config)
-          logger.info { "Pushing token authentication webhook certificates ..." }
-          upload_authentication_token_webhook_certs(webhook_config)
-        end
+        push_authentication_token_webhook_config if @config.authentication&.token_webhook
 
         copy_kube_certs
 
         logger.info { "Initializing control plane ..." }
 
         @ssh.tempfile(content: cfg.to_yaml, prefix: "kubeadm.cfg") do |tmp_file|
-          @ssh.exec!("sudo kubeadm init --config #{tmp_file}")
+          @ssh.exec!("sudo kubeadm init --ignore-preflight-errors all --skip-token-print --config #{tmp_file}")
         end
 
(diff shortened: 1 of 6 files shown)


SpComb:
Does this migrate the contents of the etcd database on the master node?

SpComb:
> Does this migrate the contents of the etcd database on the master node?
> 
> It works without migration.

I'm assuming that means that both the kubeadm `etcd` pod and our `pharos-etcd` pod both use the same `/var/lib/etcd/member` data dir, so it's only the certs that are changing?

The `etcd --name` seems to also change, but I'm guessing that will get ignored if there's already an existing cluster in the datadir.

jakolehm:
> I'm assuming that means that both the kubeadm etcd pod and our pharos-etcd pod both use the same /var/lib/etcd/member data dir, so it's only the certs that are changing?

Yes.

> The etcd --name seems to also change, but I'm guessing that will get ignored if there's already an existing cluster in the datadir.

Not sure about the --name but I tested migration and etcd does work without problems.

lib/pharos/phases/configure_master.rb
```diff
@@ -48,32 +48,17 @@ def install
...
-        end
+        push_authentication_token_webhook_config if @config.authentication&.token_webhook
 
         copy_kube_certs
 
         logger.info { "Initializing control plane ..." }
 
         @ssh.tempfile(content: cfg.to_yaml, prefix: "kubeadm.cfg") do |tmp_file|
-          @ssh.exec!("sudo kubeadm init --config #{tmp_file}")
+          @ssh.exec!("sudo kubeadm init --ignore-preflight-errors all --skip-token-print --config #{tmp_file}")
```
jnummelin: Why do we ignore preflight errors?

jakolehm: They don't really make sense here because `pharos-cluster` should smoothen the ride. Without ignore it's not possible to re-initialize kubeadm.

jnummelin: Fair enough. 

The downside is that some errors will show up only on later steps and might be harder to debug. But AFAIK we can't really work with the check on now.

lib/pharos/phases/migrate_worker.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module Pharos
+  module Phases
+    class MigrateWorker < Pharos::Phase
+      def call
+        migrate_0_5_to_0_6 if migrate_0_5_to_0_6?
+      end
+
+      def migrate_0_5_to_0_6?
+        !@ssh.file('/etc/kubernetes/kubelet.conf').read.include?('localhost:6443')
```
SpComb: The `file.read(...)` will fail if the file does not yet exist, causing new installs to fail:

```
==> Pharos::Phases::MigrateWorker @ 167.99.36.158 167.99.36.223
    $ sudo cat /etc/kubernetes/kubelet.conf
    $ sudo cat /etc/kubernetes/kubelet.conf
    cat:     /etc/kubernetes/kubelet.conf    : No such file or directory    
    cat:     ! 1
    /etc/kubernetes/kubelet.conf    : No such file or directory [Pharos::Phases::MigrateWorker @ 167.99.36.158] Pharos::SSH::RemoteCommand::ExecError: SSH exec failed with code 1: sudo cat /etc/kubernetes/kubelet.conf
cat: /etc/kubernetes/kubelet.conf: No such file or directory
    
SSH exec failed with code 1: sudo cat /etc/kubernetes/kubelet.conf
cat: /etc/kubernetes/kubelet.conf: No such file or directory
/app/lib/pharos/ssh/remote_command.rb:66:in `run!'
/app/lib/pharos/ssh/client.rb:61:in `exec!'
/app/lib/pharos/ssh/remote_file.rb:46:in `read'
/app/lib/pharos/phases/migrate_worker.rb:11:in `migrate_0_5_to_0_6?'
/app/lib/pharos/phases/migrate_worker.rb:7:in `call'
/app/lib/pharos/phase_manager.rb:70:in `block in apply'
/app/lib/pharos/phase_manager.rb:26:in `block (2 levels) in run_parallel'
```

PR #251, opened 2018-04-17
Set DEBIAN_FRONTEND=noninteractive in scripts

kke:
Removes these errors/warnings:

```
debconf: unable to initialize frontend: Dialog
debconf: (Dialog frontend will not work on a dumb terminal, an emacs shell buffer, or without a controlling terminal.)
debconf: falling back to frontend: Readline
debconf: unable to initialize frontend: Readline
debconf: (This frontend requires a controlling tty.)
debconf: falling back to frontend: Teletype
```


SpComb:
This might be something related to the scripts actually running from `stdin`... do the `apt-get install` processes also get the same stdin pipe used by `sh' to read the script? That must depend on how `sh` behaves... not sure.

kke:
Googling around showed that this is a common grievance especially in dockerfiles.

PR #265, opened 2018-04-18
Support calico as an alternative network provider

### LINKED ISSUE #17: Support high-performance network plugin
In addition to `weave` (which is the default). Possible candidates are:

- calico
- canal
- kube-router

SpComb:
Fixes #17

Using calico v3.1 with the kube datastore, without typha.


lib/pharos/config_schema.rb
```diff
@@ -33,10 +33,18 @@ def self.messages
           optional(:endpoint).filled(:str?)
         end
         optional(:network).schema do
+          optional(:provider).filled(included_in?: %(weave calico))
           optional(:dns_replicas).filled(:int?, gt?: 0)
           optional(:service_cidr).filled(:str?)
           optional(:pod_network_cidr).filled(:str?)
-          optional(:trusted_subnets).each(type?: String)
+          # TODO: deprecate trusted_subnets
```
SpComb: Need to figure out `Dry::Struct` validations to make it either reject old `config.network.trusted_subnets`, or warn about it and default the new `config.network.weave.trusted_subnets` to the deprecated field's value.

lib/pharos/phases/configure_calico.rb
```diff
@@ -0,0 +1,43 @@
...
+          name: 'calico-cni', version: CALICO_VERSION, license: 'Apache License 2.0'
+        )
+      )
+
+      def validate
+        case @host.cpu_arch.name
+        when 'amd64'
+
+        else
+          fail "Unsupported CPU architecture: #{@host.cpu_arch.name}"
```
SpComb: Calico for ARM is still WiP: https://github.com/projectcalico/calicoctl/issues/1475

> So far we have 4 components on multiple archs, 1 with an open PR, and more to come

Although it sounds like they're aiming for multi-arch images, so we shouldn't need to template it for the manifests: 

> Single image tag with multi-arch manifests is following and in process.

lib/pharos/resources/calico/typha/deployment.yml
```diff
@@ -0,0 +1,77 @@
+# This manifest creates a Deployment of Typha to back the above service.
+
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: calico-typha
```
jakolehm: Typha files are not used?

lib/pharos/config_schema.rb
SpComb:
```
==> Reading instructions ...
==> Invalid configuration file:
---
:network:
  :trusted_subnets:
  - cannot be defined
```

lib/pharos/resources/calico/typha/deployment.yml
jakolehm: Not sure if I like to keep "dead config" here. Could we add these when we actually use them?

jakolehm: @jnummelin wdyt?

jnummelin: I agree, we should add them only if they're used.

SpComb: They're part of the YAML installation package and referenced by the `daemonset`, just disabled by default... I imagine that we'll want to have optional support for typha eventually, it's recommended for use on clusters >50 nodes.

I don't see how it would hurt to hang onto them.

jakolehm:
@SpComb conflicts...

PR #267, opened 2018-04-18
Allow disabling master node taint 

SpComb:
Useful for a minimal cluster with three master nodes, allowing workload pods to run on the master nodes.

### DIFF
--- a/lib/pharos/config_schema.rb
+++ b/lib/pharos/config_schema.rb
@@ -26,6 +26,7 @@ module Pharos
               optional(:user).filled
               optional(:ssh_key_path).filled
               optional(:container_runtime).filled(included_in?: ['docker', 'cri-o'])
+              optional(:no_taint_master).filled(:bool?)
             end
           end
         end
--- a/lib/pharos/configuration/host.rb
+++ b/lib/pharos/configuration/host.rb
@@ -15,6 +15,7 @@ module Pharos
       attribute :user, Pharos::Types::Strict::String.default('ubuntu')
       attribute :ssh_key_path, Pharos::Types::Strict::String.default('~/.ssh/id_rsa')
       attribute :container_runtime, Pharos::Types::Strict::String.default('docker')
+      attribute :no_taint_master, Pharos::Types::Strict::Bool
 
       attr_accessor :os_release, :cpu_arch, :hostname, :api_endpoint
 
--- a/lib/pharos/phases/configure_master.rb
+++ b/lib/pharos/phases/configure_master.rb
@@ -93,6 +93,8 @@ module Pharos
           }
         }
 
+        config['noTaintMaster'] = true if @host.no_taint_master
+
         if @host.container_runtime == 'cri-o'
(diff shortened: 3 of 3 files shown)


lib/pharos/configuration/host.rb
```diff
@@ -15,6 +15,7 @@ class Host < Dry::Struct
       attribute :user, Pharos::Types::Strict::String.default('ubuntu')
       attribute :ssh_key_path, Pharos::Types::Strict::String.default('~/.ssh/id_rsa')
       attribute :container_runtime, Pharos::Types::Strict::String.default('docker')
+      attribute :no_taint_master, Pharos::Types::Strict::Bool
```
SpComb: Suggestions for a better config option?

kke: Could it be a role?

SpComb: `role: worker-master`?

jnummelin: IMO mixing taints into roles sounds like a bad idea. :)

Couldn't you, at least theoretically, put a taint on any node, not just on master? If so, `taint: true/false` would make sense in my head. The default would depend on the role, for masters it would default to `true`, for workers to `false`.

SpComb: This is specific to the `node-role.kubernetes.io/master:NoSchedule` taint used by kubeadm for kube master nodes.

Alternatively, we could handle `taints` similarly to `labels`, and that way we could also update them.

It seems like the kubeadm markmaster phase should know how to toggle the taint on/off, but `kubeadm upgrade` doesn't actually seem to run the markmaster phase, so currently this flag can only be used for the initial cluster up... unless we explicitly called `kubeadm alpha phase mark-master` instead.

SpComb:
Updated to use `kubeadm alpha phase mark-master` to reconfigure the master node taints.

lib/pharos/config_schema.rb
```diff
@@ -27,6 +27,7 @@ def self.messages
               optional(:user).filled
               optional(:ssh_key_path).filled
               optional(:container_runtime).filled(included_in?: ['docker', 'cri-o'])
+              optional(:master_taint).filled(:bool?)
```
jakolehm: Could this be more generic `taints` array.. if user sets it empty then master is not tainted?

SpComb: See also previous discussion... IMO the master taint is a special case.

jakolehm: Ok, but I think that `master_taint` is not generic enough as a host level option. It feels really weird tome if user can define this for hosts that have role=worker... and what happens when we introduce `taints` array?

SpComb: What would the taints array look like?

```yaml
hosts:
  - address: 192.0.2.1
    taints:
      - key: node-role.kubernetes.io/master
        effect: NoSchedule
```

This will then need to (conditionally) disable the kubeadm-managed taints, and then patch the remaining taints into the node using the `LabelNode` phase.

PR #269, opened 2018-04-19
Refactor phase component registration

kke:
`Pharos::Phases.components` shouldn't contain anything but `Pharos::Phases::Component`s, so:

```ruby
      register_component(	
        Pharos::Phases::Component.new(	
          name: 'cri-o', version: Pharos::CRIO_VERSION, license: 'Apache License 2.0'	
        )	
      )
```

can become:

```ruby
      register_component 'cri-o', version: Pharos::CRIO_VERSION, license: 'Apache License 2.0'
```

Also noticed `kubeadm` wasn't registered and cfssl version was hardcoded in the shell script and registered separately in the configure cfssl phase.

### DIFF
--- a/lib/pharos/phase.rb
+++ b/lib/pharos/phase.rb
@@ -19,8 +19,7 @@ module Pharos
     # @param version [String]
     # @param license [String]
     def self.register_component(name, version:, license:)
-      slug = name.gsub(/[^a-z0-9]+/, '_')
-      Pharos::Phases.components[slug] = Pharos::Phases::Component.new(
+      Pharos::Phases.components[name] = Pharos::Phases::Component.new(
         name: name,
         version: version,
         license: license
--- a/lib/pharos/phase.rb
+++ b/lib/pharos/phase.rb
@@ -14,8 +14,17 @@ module Pharos
       "#{self.class.title} @ #{@host}"
     end
 
-    def self.register_component(component)
-      Pharos::Phases.register_component component
+    # Register a component to the Pharos::Phases component registry
+    # @param name [String]
+    # @param version [String]
+    # @param license [String]
+    def self.register_component(name, version:, license:)
+      slug = name.gsub(/[^a-z0-9]+/, '_')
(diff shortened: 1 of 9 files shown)


lib/pharos/phase.rb
```diff
@@ -14,8 +14,17 @@ def to_s
...
     end
 
-    def self.register_component(component)
-      Pharos::Phases.register_component component
+    # Register a component to the Pharos::Phases component registry
+    # @param name [String]
+    # @param version [String]
+    # @param license [String]
+    def self.register_component(name, version:, license:)
+      slug = name.gsub(/[^a-z0-9]+/, '_')
```
SpComb: If the `name` is used as an openstruct method name, then should it be a symbol instead of a string? 

Alternatively just use `components['cri-o'].version` instead?

lib/pharos/scripts/configure-cfssl.sh
```diff
@@ -8,7 +8,7 @@ else
     MIRROR="http://ports.ubuntu.com"
 fi
 
-CFSSL_URL="${MIRROR}/pool/universe/g/golang-github-cloudflare-cfssl/golang-cfssl_1.2.0+git20160825.89.7fb22c8-3_${ARCH}.deb"
+CFSSL_URL="${MIRROR}/pool/universe/g/golang-github-cloudflare-cfssl/golang-cfssl_${VERSION:?"Need to set VERSION"}_${ARCH}.deb"
```
SpComb: Use `set -ue` instead, then the `${VERSION}` will fail if not defined.

lib/pharos/phase.rb
kke: Can't be a symbol beacuse the name is shown in the component version listing `pharos-cluster version`. Otherwise would have to do something like `register_component :foo, name: 'foo'`

SpComb:
Partially happened in #327

kke:
Yep, the part about not having to `register_component(Pharos::Phases::Component.new(..))`.

Too conflicty, rewriting if relevant.

PR #276, opened 2018-04-19
Fix external etcd

SpComb:
Fixes #274
Fixes #275

PR #277, opened 2018-04-19
Refactor master phases

SpComb:
* Faster execution by also running `ConfigureKubelet` in parallel for master nodes, not only worker nodes
* Simplify phase execution by getting rid of the `ConfigureMaster` => `ConfigureKubelet` nesting
* Also update the external etcd, audit webhook and token authentication certs/configs after the initial install, as required by `kubeadm upgrade` after `cluster.yml` changes
* Correctly handle kube PKI certs for interrupted/resumed multi-master installs (always use certs from the first master node) 

# Changes
* Refactor `ConfigureMaster` logic for generating the kubeadm config file into `Pharos::Kubeadm::Config`
* Split the `ConfigureMaster` phase into three phases: `ConfigureMaster`, `UpgradeMaster`, `InstallMaster`
* Run the `ConfigureKubelet` phase in parallel across all nodes simultaneously, after the `UpgradeMaster` but before `InstallMaster`
* Move the remote `~/.kube/config` setup into the `ConfigureClient` phase

# TODO
- [x] Fix specs
- [ ] Better spec coverage for the various uses of `Pharos::Kubeadm::Config`
- [ ] Testing of audit webhook, token auth webhook configurations


lib/pharos/phases/configure_master.rb
```diff
@@ -1,361 +1,69 @@
...
-
-        copy_kube_certs
-
-        logger.info { "Initializing control plane ..." }
-
-        @ssh.tempfile(content: cfg.to_yaml, prefix: "kubeadm.cfg") do |tmp_file|
-          @ssh.exec!("sudo kubeadm init --ignore-preflight-errors all --skip-token-print --config #{tmp_file}")
-        end
-
-        cache_kube_certs
```
SpComb: Bug in master: the `cache_kube_certs`/`copy_kube_certs` is not handled correctly if you interrupt the `pharos-cluster up` during the initial install after the first master node was installed... when you later retry, the first master host will already be `installed?`, so it will skip the `install` => `cache_kube_certs`, and the remaining master hosts will use different certs than the first master node.

jakolehm:
Great improvement!

lib/pharos/phases/configure_client.rb
```diff
@@ -6,17 +6,36 @@ class ConfigureClient < Pharos::Phase
...
+      def read_kubeconfig
         logger.info { "Fetching kubectl config ..." }
         config_data = @ssh.file("/etc/kubernetes/admin.conf").read
+        config_data = config_data.gsub(%r{(server: https://)(.+)(:6443)}, "\\1#{@host.api_address}\\3")
+      end
+
+      def install_local_kubeconfig(config_data)
+        Dir.mkdir(config_dir, 0o700) unless Dir.exist?(config_dir)
+
+        config_file = File.join(config_dir, @host.api_address)
```
SpComb: This should be `@host.address`, I think...

jakolehm: Why?

SpComb: I suppose if we're using the `api.endpoint` from the config for the CLI => Master requests, then we can't run the `ConfigureClient` phase for all master hosts... need to move the remote kubeconfig thing back out of this phase.

SpComb: See #278 that changes this to use `@config.api_endpoint` instead.

SpComb:
The `ConfigureMaster` specs were only testing the `generate_config` methods, so they also work for the new `Pharos::Kubeadm::Config` with only minor changes.

SpComb:
Still planning on working on better spec coverage, but this is now green and reviewable.

lib/pharos/cluster_manager.rb
```diff
@@ -42,25 +42,28 @@ def validate
 
     def apply_phases
       apply_phase(Phases::ValidateHost, config.hosts, ssh: true, parallel: true)
-      apply_phase(Phases::MigrateMaster, config.master_hosts, ssh: true, parallel: true)
       apply_phase(Phases::ConfigureHost, config.hosts, ssh: true, parallel: true)
+
+      apply_phase(Phases::MigrateMaster, config.master_hosts, ssh: true, parallel: true)
       apply_phase(Phases::ConfigureCfssl, config.etcd_hosts, ssh: true, parallel: true)
       apply_phase(Phases::ConfigureEtcdCa, config.etcd_hosts[0...1], ssh: true, parallel: false)
```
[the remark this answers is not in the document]
SpComb: Yes, this is intentional trickery from  #276... if the `etcd_hosts` is empty, then this is an empty array and the phase is skipped; otherwise it's equivalent to `[config.etcd_hosts.first]`. This should really be better written out more explicitly.

lib/pharos/cluster_manager.rb
```diff
@@ -74,9 +74,13 @@ def apply_phases
...
 
       apply_phase(Phases::ConfigureSecretsEncryption, master_hosts, ssh: true, parallel: false)
+      apply_phase(Phases::SetupMaster, master_hosts, ssh: true, parallel: true)
+      apply_phase(Phases::UpgradeMaster, master_hosts, ssh: true, parallel: false)
+
+      apply_phase(Phases::MigrateWorker, config.worker_hosts, ssh: true, parallel: true)
+      apply_phase(Phases::ConfigureKubelet, config.hosts, ssh: true, parallel: true)
+
       apply_phase(Phases::ConfigureMaster, master_hosts, ssh: true, parallel: false)
-      apply_phase(Phases::MigrateWorker, config.worker_hosts, ssh: true, parallel: true, master: master_hosts.first)
```
SpComb: The `master: ...` was already removed as part of #302, but snuck back in during the #293 merge.

lib/pharos/phases/setup_master.rb
```diff
@@ -56,9 +48,12 @@ def push_authentication_token_webhook_certs(webhook_config)
...
         webhook_config = @config.authentication.token_webhook.config
 
-        auth_token_webhook_config = generate_authentication_token_webhook_config(webhook_config)
+        logger.info { "Pushing token authentication webhook config ..." }
+        auth_token_webhook_config = kubeadm.generate_authentication_token_webhook_config(webhook_config)
+
+        @ssh.exec!('sudo mkdir -p /etc/kubernetes/authentication')
+        @ssh.file('/etc/kubernetes/authentication/token-webhook-config.yaml').write(auth_token_webhook_config.to_yaml)
 
-        push_authentication_token_webhook_config(auth_token_webhook_config)
```
SpComb: Needs specs to pick these kinds of things up

```
lib/pharos/phases/setup_master.rb:56:7: W: Lint/DuplicateMethods: Method Pharos::Phases::SetupMaster#push_authentication_token_webhook_config is defined at both lib/pharos/phases/setup_master.rb:39 and lib/pharos/phases/setup_master.rb:56.
      def push_authentication_token_webhook_config
      ^^^
```

SpComb:
> Need to resolve conflicts.

Hesitant to keep merging this again and again... but it doesn't look like it will conflict with #360 or other pending PRs ATM, so I'll go ahead..

SpComb:
Noticed that the upgrade case now does both `Phases::UpgradeMaster#upgrade` followed by `Phases::ConfigureMaster#reconfigure`... it shouldn't hurt, but it's extra work... could be avoided by using `host_context` (?) to mark the node as upgraded and skip the reconfigure.

Tested with
* `KUBE_VERSION=1.10.0 pharos-cluster up` (master install)
* `KUBE_VERSION=1.10.0 pharos-cluster up` (master reconfigure)
* `KUBE_VERSION=1.10.2 pharos-cluster up` (master upgrade+reconfigure)
* `KUBE_VERSION=1.10.2 pharos-cluster up` (master reconfigure)

PR #278, opened 2018-04-19
Refactor addons, phases to use Pharos::Kube::Session

SpComb:
This should allow running kube phases/addons in parallel in the future, by using a different `Pharos::Kube::Session` for each thread.

* Refactor `Pharos::Kube::Stack#prune` to explicitly ignore the checksum when uninstalling an addon
* Refactor `Pharos::Kube::Stack` vars to be passed to `#apply` instead of `new`, because they are not needed/available for `prune`
* Replace `Pharos::Kube.client(<String>)` with `Pharos::Kube::Session.client(<Pharos::Configuration::Host>)` using (threadsafe) `Pharos::Kube::Client.for_host`
* <s>Fix the `MigrateWorker` phase to be independent of the `@master`</s> #302 
* Replace all usages of `@master` and `Pharos::Kube` class methods with with `@kube` (`Pharos::Kube::Session`) instance methods throughout `Pharos::Phases` and `Pharos::Addons`
* <s>Remove the non-threadsafe `Pharos::Kube` class methods, and explicitly pass the `Pharos::Kube::Session` from the `ClusterManager` to the `PhaseManager` and `AddonManager` for non-threaded use</s> TODO: `Pharos::Kube::Manager` with some kind of threadsafe session pool for parallel kube phases/addons


kke:
Maybe make a `Pharos::Session` or such?

I was looking at [dry-container](http://dry-rb.org/gems/dry-container/) for something like this.

Basically it's a container for composition:

```ruby
container.register(:kube_session, Pharos::Kube::Session.new)
```

lib/pharos/kube.rb
```diff
@@ -10,61 +10,33 @@ module Kube
...
 
-    # @param host [String]
+    # @return [String]
+    def self.config_dir
+      File.join(Dir.home, '.pharos')
+    end
+
+    # @param host [Pharos::Configuration::Host]
     # @return [String]
     def self.host_config_path(host)
```
kke: Maybe this stuff should actually be in `Pharos::Config::Host`

lib/pharos/kube/stack.rb
```diff
@@ -60,15 +60,13 @@ def prune(checksum = nil)
           end
 
           if checksum
-            prunables = objects.select do |obj|
+            objects = objects.select { |obj|
```
kke:
```ruby
objects.delete_if do |obj|
  ...
end
```

lib/pharos/kube.rb
SpComb: I dont't need to touch it here, we can refactor the config harder later if we want to.

SpComb: Also, the configs are not actually per-host... if you configure with `api.endpoint`, then you'll only have one config file for that endpoint... so it's a little illogical to actually involve the `Pharos::Configuration::Host` here, because it's not actually host-specific. Instead, the cluster config has an api endpoint that defaults to the address of the first master host...

SpComb: Flipped the `api_endpoint` logic around, it's now defined on the top-level config, not on the host.

SpComb:
This should be good for review now.

lib/pharos/kube/client.rb
```diff
@@ -10,7 +10,7 @@ class Client < ::Kubeclient::Client
       def self.for_host(host, version)
         config = Pharos::Kube.host_config(host)
         path_prefix = version == 'v1' ? 'api' : 'apis'
-        api_version, api_group = version.split('/').reverse
+        api_group, api_version = version.split('/')
```
kke: I think it was intentional, I was confused by this line too. If version does not have '/', `api_group` would be nil. With this change api_version will be nil.

SpComb: Odd, this does behave differently, but it did still work when testing... needs some specs.

lib/pharos/config.rb
```diff
@@ -47,29 +47,45 @@ def master_hosts
...
+      return [] unless etcd_hosts?
+
+      # XXX: sort master hosts by etcd_sort_score
+      @etcd_hosts ||= hosts.select { |h| h.role == 'etcd' }.sort_by(&:etcd_sort_score)
 
-      etcd_hosts = hosts.select { |h| h.role == 'etcd' }
-      if etcd_hosts.empty?
-        master_hosts
+      if @etcd_hosts.empty?
+        sorted_master_hosts
```
SpComb: This needs to use `master_hosts.sort_by(&:etcd_sort_score)` instead

lib/pharos/cluster_manager.rb
```diff
@@ -45,51 +48,34 @@ def validate
...
+      if config.etcd_hosts?
+        apply_phase(Phases::ConfigureCfssl, config.etcd_hosts, ssh: true, parallel: true)
+        apply_phase(Phases::ConfigureEtcdCa, [config.etcd_hosts.first], ssh: true, parallel: false)
+        apply_phase(Phases::ConfigureEtcdChanges, [config.etcd_hosts.first], ssh: true, parallel: false)
+        apply_phase(Phases::ConfigureEtcd, config.etcd_hosts, ssh: true, parallel: true)
       end
 
-      apply_phase(Phases::ConfigureSecretsEncryption, master_hosts, ssh: true, parallel: false)
-      apply_phase(Phases::ConfigureMaster, master_hosts, ssh: true, parallel: false)
-      apply_phase(Phases::MigrateWorker, config.worker_hosts, ssh: true, parallel: true, master: master_hosts.first)
```
SpComb: (This `master: ...` was already removed as part of #302, but snuck back in during the #293 merge)

lib/pharos/cluster_manager.rb
```diff
@@ -22,15 +29,21 @@ def ssh_manager
...
       )
     end
 
     # @return [Pharos::AddonManager]
     def addon_manager
-      @addon_manager ||= Pharos::AddonManager.new(@config, @context)
+      @addon_manager ||= Pharos::AddonManager.new(
+        @config,
+        cluster_context: @context,
+        kube_session: kube_session
```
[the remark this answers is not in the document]
SpComb: > Need to delay the Pharos::Kube::Session creation to until it's actually needed by the PhaseManager / AddonManager... I suppose maybe the answer here could be to already introduce the Pharos::Kube::Manager that could later also be used for managing the per-thread sessions.

Didn't go for the `Pharos::Kube::Manager` yet, that would be over-engineering at this point... kept the non-threadsafe `Pharos::Kube#session` instead, it's simplest, and easy to continue improving that later because it's only used by the `PhaseManager`/`AddonManager` now.

SpComb:
Initial smoketest of `pharos-cluster up` passed now, should be reviewable again after being updated for 1.0 changes.

lib/pharos/kube/client.rb
SpComb: Finally found a case where this breaks... seems like these usually only end up getting interpolated into the path as `".../#{api_group}/#{api_version}"`, and in that case it doesn't matter which is which... except the `create_entity` tries to detect the `@api_group` from the given URL, and then automatically set `hash[:apiVersion] = @api_group + @api_version`:
```
no implicit conversion of nil into String
/usr/local/bundle/gems/kubeclient-3.0.0/lib/kubeclient/common.rb:324:in `+'
/usr/local/bundle/gems/kubeclient-3.0.0/lib/kubeclient/common.rb:324:in `create_entity'
/usr/local/bundle/gems/kubeclient-3.0.0/lib/kubeclient/common.rb:185:in `block (2 levels) in define_entity_methods'
/app/lib/pharos/phases/configure_weave.rb:33:in `rescue in ensure_passwd'
/app/lib/pharos/phases/configure_weave.rb:20:in `ensure_passwd'
/app/lib/pharos/phases/configure_weave.rb:16:in `call'
/app/lib/pharos/phase_manager.rb:80:in `block in apply'
/app/lib/pharos/phase_manager.rb:47:in `block in run_serial'
/app/lib/pharos/phase_manager.rb:46:in `map'
/app/lib/pharos/phase_manager.rb:46:in `run_serial'
/app/lib/pharos/phase_manager.rb:58:in `run'
/app/lib/pharos/phase_manager.rb:77:in `apply'
/app/lib/pharos/cluster_manager.rb:88:in `apply_phase'
/app/lib/pharos/cluster_manager.rb:71:in `apply_phases'
/app/lib/pharos/up_command.rb:104:in `configure'
/app/lib/pharos/up_command.rb:47:in `block in execute'
/app/lib/pharos/up_command.rb:45:in `chdir'
/app/lib/pharos/up_command.rb:45:in `execute'
/app/lib/pharos/root_command.rb:14:in `run'
```

SpComb:
Outdated after #475

PR #286, opened 2018-04-20
[WiP] Automatic phase building

kke:
The DSL is still ugly and seems mostly unnecessary.

For example `LabelNode` does not actually run on hosts, it runs on kube api. Maybe it should actually do `@config.hosts.each` instead of being run individually for each host.


kke:
closing because of #278 and rethink for now

PR #289, opened 2018-04-23
Allow to configure cloud-config

### LINKED ISSUE #287: Option to define --cloud-config
```yaml
cloud:
  provider: openstack
  config: ./path/to/cloud.conf
```

nevalla:
Fixes #287

### DIFF
--- a/lib/pharos/phases/configure_master.rb
+++ b/lib/pharos/phases/configure_master.rb
@@ -14,6 +14,8 @@ module Pharos
       AUDIT_CFG_DIR = (PHAROS_DIR + '/audit').freeze
       SECRETS_CFG_DIR = (PHAROS_DIR + '/secrets-encryption').freeze
       SECRETS_CFG_FILE = (SECRETS_CFG_DIR + '/config.yml').freeze
+      CLOUD_CFG_DIR = (PHAROS_DIR + '/cloud').freeze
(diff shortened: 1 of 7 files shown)


nevalla:
Releted documention PR: https://github.com/kontena/pharos-docs/pull/8

lib/pharos/phases/configure_master.rb
```diff
@@ -53,6 +55,8 @@ def install
           copy_external_etcd_certs
         end
 
+        push_cloud_config if @config.cloud&.config
```
SpComb: The master also runs `configure_kubelet` => `ConfigureKubelet`, won't that also upload the same file into place? It doesn't hurt to upload the same file twice, but it seems a little odd.

Maybe this should be at a different path, so that the two phases don't interfere with eachother... or just skip the file write here and rely on the file put into place by the `ConfigureKubelet` phase, but those kinds of inter-phase dependencies are annoying to figure out later on.

jakolehm:
@nevalla conflicts...

lib/pharos/phases/configure_master.rb
nevalla: Changed to use `/etc/pharos/kubelet` for kubelet's cloud-config.

nevalla:
Rubocop should be happy now

PR #290, opened 2018-04-23
Scale nginx default-backends by number of workers

### LINKED ISSUE #40: Scale ingress-nginx default backends
Based on number of workers.

kke:
Fixes #40

Something like this?

### DIFF
--- a/lib/pharos/addons/ingress_nginx.rb
+++ b/lib/pharos/addons/ingress_nginx.rb
@@ -38,7 +38,8 @@ module Pharos
         apply_stack(
           configmap: config.configmap || {},
           node_selector: config.node_selector || {},
-          image: image_name
+          image: image_name,
+          worker_hosts: addon_manager.config.worker_hosts.size
         )
       end
     end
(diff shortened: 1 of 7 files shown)


lib/pharos/addons/ingress_nginx.rb
```diff
@@ -38,7 +38,8 @@ def install
         apply_stack(
           configmap: config.configmap || {},
           node_selector: config.node_selector || {},
-          image: image_name
+          image: image_name,
+          worker_hosts: addon_manager.config.worker_hosts.size
```
SpComb: The addons already have `config.worker_hosts.size`, no need to pass the `addon_manager`

jakolehm:
Issue description was not the best one. I meant that we should scale default backend to more than one replica. IMHO it does not make sense to scale this to every node (and then we should use daemonset).

Maybe something similar to kube-dns?

lib/pharos/addons/ingress_nginx.rb
kke: Addon's `config` is the struct/schema thing for the current addon:

```ruby
config = configs[addon_class.name]
```

SpComb: Bleh, right. The addons should probably also get the top-level config... passing the `addon_manager` doesn't feel right.

PR #293, opened 2018-04-23
Handle add/remove of master hosts

### LINKED ISSUE #240: Adding new masters to existing cluster does not work
At least we should block control-plane scaling until we have proper logic in place.

jakolehm:
Adds logic for adding/removing masters. It allows to grow number of masters one-by-one (reason for this is internal etcd clustering). Multiple masters can be removed  at the same time from cluster if majority is kept in place. These rules do not apply if cluster.yml is configured to use external etcd endpoints.

Fixes #240

### DIFF
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -42,12 +42,12 @@ module Pharos
 
     # @return [Array<Pharos::Configuration::Host>]
     def sorted_master_hosts
-      config.master_hosts.sort_by { |h| h.master_sort_score }
+      config.master_hosts.sort_by(&:master_sort_score)
     end
 
     # @return [Array<Pharos::Configuration::Host>]
     def sorted_etcd_hosts
-      config.etcd_hosts.sort_by { |h| h.etcd_sort_score }
+      config.etcd_hosts.sort_by(&:etcd_sort_score)
     end
 
     def apply_phases
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -44,9 +44,12 @@ module Pharos
       apply_phase(Phases::ValidateHost, config.hosts, ssh: true, parallel: true)
       apply_phase(Phases::MigrateMaster, config.master_hosts, ssh: true, parallel: true)
       apply_phase(Phases::ConfigureHost, config.hosts, ssh: true, parallel: true)
-      apply_phase(Phases::ConfigureCfssl, config.etcd_hosts, ssh: true, parallel: true)
-      apply_phase(Phases::ConfigureEtcdCa, config.etcd_hosts[0...1], ssh: true, parallel: false)
-      apply_phase(Phases::ConfigureEtcd, config.etcd_hosts, ssh: true, parallel: true)
+      unless @config.etcd&.hosts
+        apply_phase(Phases::ConfigureCfssl, config.etcd_hosts, ssh: true, parallel: true)
+        apply_phase(Phases::ConfigureEtcdCa, config.etcd_hosts[0...1], ssh: true, parallel: false)
+        apply_phase(Phases::ConfigureEtcdChanges, config.etcd_hosts[0...1], ssh: true, parallel: false)
+        apply_phase(Phases::ConfigureEtcd, config.etcd_hosts, ssh: true, parallel: true)
+      end
 
       apply_phase(Phases::ConfigureSecretsEncryption, config.master_hosts, ssh: true, parallel: false)
       apply_phase(Phases::ConfigureMaster, config.master_hosts, ssh: true, parallel: false)
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -40,35 +40,74 @@ module Pharos
       addon_manager.validate
     end
 
+    # @return [Array<Pharos::Configuration::Host>]
+    def sorted_master_hosts
+      @sorted_master_hosts ||= config.master_hosts.sort_by { |h| master_sort_score(h) }
+    end
+
+    # @param host [Pharos::Configuration::Host]
+    # @return [Integer]
+    def master_sort_score(host)
+      score = 0
+      return score if host.checks['api_healthy']
+      score += 1
+      return score if host.checks['kubelet_configured']
+
+      score + 1
+    end
+
+    # @return [Array<Pharos::Configuration::Host>]
+    def sorted_etcd_hosts
+      @sorted_etcd_hosts ||= config.etcd_hosts.sort_by { |h| etcd_sort_score(h) }
+    end
+
+    # @param host [Pharos::Configuration::Host]
+    # @return [Integer]
+    def etcd_sort_score(host)
+      score = 0
+      return score if host.checks['etcd_healthy']
+      score += 1
+      return score if host.checks['etcd_ca_exists']
+
+      score + 1
+    end
+
     def apply_phases
       apply_phase(Phases::ValidateHost, config.hosts, ssh: true, parallel: true)
-      apply_phase(Phases::MigrateMaster, config.master_hosts, ssh: true, parallel: true)
+
+      master_hosts = sorted_master_hosts
+
+      apply_phase(Phases::MigrateMaster, master_hosts, ssh: true, parallel: true)
       apply_phase(Phases::ConfigureHost, config.hosts, ssh: true, parallel: true)
+
       unless @config.etcd&.hosts
-        apply_phase(Phases::ConfigureCfssl, config.etcd_hosts, ssh: true, parallel: true)
-        apply_phase(Phases::ConfigureEtcdCa, config.etcd_hosts[0...1], ssh: true, parallel: false)
-        apply_phase(Phases::ConfigureEtcdChanges, config.etcd_hosts[0...1], ssh: true, parallel: false)
-        apply_phase(Phases::ConfigureEtcd, config.etcd_hosts, ssh: true, parallel: true)
+        etcd_hosts = sorted_etcd_hosts
+        apply_phase(Phases::ConfigureCfssl, etcd_hosts, ssh: true, parallel: true)
+        apply_phase(Phases::ConfigureEtcdCa, [etcd_hosts.first], ssh: true, parallel: false)
+        apply_phase(Phases::ConfigureEtcdChanges, [etcd_hosts.first], ssh: true, parallel: false)
+        apply_phase(Phases::ConfigureEtcd, etcd_hosts, ssh: true, parallel: true)
       end
 
-      apply_phase(Phases::ConfigureSecretsEncryption, config.master_hosts, ssh: true, parallel: false)
-      apply_phase(Phases::ConfigureMaster, config.master_hosts, ssh: true, parallel: false)
-      apply_phase(Phases::MigrateWorker, config.worker_hosts, ssh: true, parallel: true, master: config.master_host)
-      apply_phase(Phases::ConfigureKubelet, config.worker_hosts, ssh: true, parallel: true) # TODO: also run this phase in parallel for the master nodes, if not doing an upgrade?
-      apply_phase(Phases::ConfigureClient, [config.master_host], ssh: true, parallel: false)
+      apply_phase(Phases::ConfigureSecretsEncryption, master_hosts, ssh: true, parallel: false)
+      apply_phase(Phases::ConfigureMaster, master_hosts, ssh: true, parallel: false)
+      apply_phase(Phases::MigrateWorker, config.worker_hosts, ssh: true, parallel: true, master: master_hosts.first)
+      apply_phase(Phases::ConfigureKubelet, config.worker_hosts, ssh: true, parallel: true)
+      apply_phase(Phases::ConfigureClient, [master_hosts.first], ssh: true, parallel: false)
 
       # master is now configured and can be used
-      apply_phase(Phases::ConfigureDNS, [config.master_host], master: config.master_host)
-      apply_phase(Phases::ConfigureNetwork, [config.master_host], master: config.master_host)
-      apply_phase(Phases::ConfigureMetrics, [config.master_host], master: config.master_host)
-      apply_phase(Phases::StoreClusterYAML, [config.master_host], master: config.master_host, config_content: @config_content)
-      apply_phase(Phases::ConfigureBootstrap, [config.master_host], ssh: true) # using `kubeadm token`, not the kube API
+      apply_phase(Phases::ConfigureDNS, [master_hosts.first], master: master_hosts.first)
+      apply_phase(Phases::ConfigureNetwork, [master_hosts.first], master: master_hosts.first)
+      apply_phase(Phases::ConfigureMetrics, [master_hosts.first], master: master_hosts.first)
+      apply_phase(Phases::StoreClusterYAML, [master_hosts.first], master: master_hosts.first, config_content: @config_content)
+      apply_phase(Phases::ConfigureBootstrap, [master_hosts.first], ssh: true) # using `kubeadm token`, not the kube API
 
       apply_phase(Phases::JoinNode, config.worker_hosts, ssh: true, parallel: true)
 
-      apply_phase(Phases::LabelNode, config.hosts, master: config.master_host, ssh: false, parallel: false) # NOTE: uses the @master kube API for each node, not threadsafe
+      apply_phase(Phases::LabelNode, config.hosts, master: master_hosts.first, ssh: false, parallel: false) # NOTE: uses the @master kube API for each node, not threadsafe
     end
 
+    # @param phase_class [Pharos::Phase]
+    # @param hosts [Array<Pharos::Configuration::Host>]
     def apply_phase(phase_class, hosts, **options)
       return if hosts.empty?
 
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -53,12 +53,16 @@ module Pharos
     def apply_phases
       apply_phase(Phases::ValidateHost, config.hosts, ssh: true, parallel: true)
 
+      # we need to use sorted masters because phases expects that first one has
+      # ca etc config files
       master_hosts = sorted_master_hosts
 
       apply_phase(Phases::MigrateMaster, master_hosts, ssh: true, parallel: true)
       apply_phase(Phases::ConfigureHost, config.hosts, ssh: true, parallel: true)
 
       unless @config.etcd&.hosts
+        # we need to use sorted etcd hosts because phases expects that first one has
+        # ca etc config files
         etcd_hosts = sorted_etcd_hosts
         apply_phase(Phases::ConfigureCfssl, etcd_hosts, ssh: true, parallel: true)
         apply_phase(Phases::ConfigureEtcdCa, [etcd_hosts.first], ssh: true, parallel: false)
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -42,34 +42,12 @@ module Pharos
 
     # @return [Array<Pharos::Configuration::Host>]
     def sorted_master_hosts
-      @sorted_master_hosts ||= config.master_hosts.sort_by { |h| master_sort_score(h) }
-    end
-
-    # @param host [Pharos::Configuration::Host]
-    # @return [Integer]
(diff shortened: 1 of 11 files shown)


lib/pharos/phases/configure_etcd_changes.rb
```diff
@@ -0,0 +1,63 @@
...
+module Pharos
+  module Phases
+    class ConfigureEtcdChanges < Pharos::Phase
+      title 'Configure etcd member changes'
+
+      def call
+        store_initial_cluster_state
+
+        if etcd.healthy?
+          add_new_members if initial_cluster_state == 'existing'
```
jakolehm: Does not yet handle removal.

lib/pharos/phases/configure_master.rb
```diff
@@ -346,6 +346,11 @@ def push_authentication_token_webhook_config
       end
 
       def configure
+        cfg = generate_config
```
jakolehm: This probably fixes some other issues as well. Anyway we need/should rewrite control plane config always.

lib/pharos/scripts/configure-etcd.sh
```diff
@@ -82,7 +82,12 @@ fi
 
 echo "Waiting etcd to launch on port 2380..."
 
-while ! nc -z ${PEER_IP} 2380; do
+etcd_healthy() {
```
jakolehm: Had to remove netcat because it gave false positives.

lib/pharos/etcd/client.rb
```diff
@@ -0,0 +1,34 @@
...
+    class Client
+      CURL = 'sudo curl -s --connect-timeout 2 --cacert /etc/pharos/pki/ca.pem --cert /etc/pharos/pki/etcd/client.pem --key /etc/pharos/pki/etcd/client-key.pem https://localhost:2379'
+
+      def initialize(ssh)
+        @ssh = ssh
+      end
+
+      def healthy?
+        response = @ssh.exec!("#{CURL}/health")
+        response.stdout == '{"health": "true"}'
```
SpComb: Better to `JSON.parse` instead of comparing the output verbatim?

lib/pharos/etcd/client.rb
```diff
@@ -0,0 +1,34 @@
...
+      rescue
+        false
+      end
+
+      # @return [Array<Hash>]
+      def members
+        result = @ssh.exec("#{CURL}/v2/members")
+        JSON.parse(result.stdout)['members']
+      rescue JSON::ParserError
+        []
```
SpComb: Bad idea, just let it raise instead? Is there some known case where this happens?

lib/pharos/etcd/client.rb
```diff
@@ -0,0 +1,34 @@
...
+      end
+
+      # @param host [Pharos::Configuration::Host]
+      def add_member(host)
+        @ssh.exec!("#{CURL}/v2/members -X POST -H 'Content-Type: application/json' -d '{\"peerURLs\":[\"https://#{host.peer_address}:2380\"]}'")
```
SpComb: The quoting is going to get funky.... I'd recommend `curl -d @-` and then `stdin: JSON.dumps(...)`

lib/pharos/phases/configure_master.rb
SpComb: This is going to fail for things like enabling the `@config.audit`, unless you also call `push_audit_config` to write those referenced config files as well.

lib/pharos/scripts/configure-etcd.sh
```diff
@@ -82,7 +82,12 @@ fi
...
+  response=$(curl -s --cacert /etc/pharos/pki/ca.pem --cert /etc/pharos/pki/etcd/client.pem --key /etc/pharos/pki/etcd/client-key.pem https://${PEER_IP}:2379/health)
+  [ "${response}" = '{"health": "true"}' ]
```
SpComb: Something like `jq -r .health <<<"$response"` would be less whitespace-sensitive, but I don't think `jq` is actually installed by default.

jakolehm:
> The ConfigureMaster => kubeadm alpha phase ... change is probably a good idea, but is it strictly necessary for this PR?

Cluster will be left in a weird state if the control plane configuration is not updated. I'll reconsider if we can get #277 merged sooner than later.

lib/pharos/phases/configure_etcd_changes.rb
```diff
@@ -0,0 +1,63 @@
...
+      # @param peer [Pharos::Configuration::Host]
+      # @return [String]
+      def peer_name(peer)
+        peer_index = @config.etcd_hosts.find_index { |h| h == peer }
+        "etcd#{peer_index + 1}"
+      end
+
+      # @return [Boolean]
+      def first_peer?
+        @config.etcd_hosts[0] == @host
```
SpComb: Not used, yet?

SpComb:
> The ConfigureMaster => kubeadm alpha phase ... change is probably a good idea, but is it strictly necessary for this PR?
> 
> Cluster will be left in a weird state if the control plane configuration is not updated. I'll reconsider if we can get #277 merged sooner than later.

No need, it's fine to do this here if it's needed, I just want to understand why.

Seems like the `--apiserver-count` and `etcd.endpoints` change.

lib/pharos/etcd/client.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+require 'json'
+
+module Pharos
+  module Etcd
+    class Client
+      CURL = 'sudo curl -s --connect-timeout 2 --cacert /etc/pharos/pki/ca.pem --cert /etc/pharos/pki/etcd/client.pem --key /etc/pharos/pki/etcd/client-key.pem https://localhost:2379'
```
SpComb: `curl -sSf` would be best, in order to handle HTTP API errors as SSH exec errors.

lib/pharos/phases/configure_etcd_changes.rb
```diff
@@ -0,0 +1,77 @@
...
+      end
+
+      def add_new_members
+        member_list = etcd.members
+        new_members = @config.etcd_hosts.select { |h|
+          !member_list.find { |m|
+            m['name'] == peer_name(h) && m['peerURLs'] == ["https://#{h.peer_address}:2380"]
+          }
+        }
+        if new_members.size > (member_list.size / 2.0).ceil
```
jakolehm: This is not perfect.. suggestions?

kke: `new_members.size / member_list.size.to_f >= 0.5` ?

SpComb: Yeah, the `new_members.size > (member_list.size / 2.0).ceil` would allow you to add three new members to an existing cluster of two members: `['c', 'd', 'e'].size > (['a', 'b', 'c', 'd', 'e'].size / 2.0).ceil => false`

The `new_members.size / member_list.size.to_f >= 0.5` is also wrong, it would refuse to add 1 new member bring the cluster to two members: `['a'].size / ['a', 'b'].size.to_f >= 0.5 => true`

jakolehm: It's still a bit unclear if we can actually add more than one peer at once.

SpComb: However, I figure the existing condition is probably correct... if you want to be able to scale the cluster up from one member to two members, then you will have to allow breaking the quorum by one node when adding the new member.

If you wanted the cluster to retain quorum with the new members added, then it would be something like `unless member_list.size - new_members.size >= member_list.size / 2 + 1`, but that would make it impossible to scale up a cluster from one node.

SpComb: > It's still a bit unclear if we can actually add more than one peer at once.

Yes, the best approach would be to add one new peer at a time, waiting for the new peer to successfully join the cluster before adding the next one.

lib/pharos/cluster_manager.rb
```diff
@@ -40,30 +40,47 @@ def validate
...
+      }
+      if master
+        master
+      else
+        config.master_host
+      end
+    end
+
     def apply_phases
+      master = preferred_master_host
```
jakolehm: @SpComb wdyt? This allows to replace the "first" master host.

jakolehm: I mean: first master is not always the best one...

SpComb: Sure, although if you're using `api.endpoint` then it doesn't really matter which host you select... this is more explicit in #278

:-1: for using SSH to check for the `/etc/kubernetes/kubelet.conf` though... why not just use `https://#{host.address}:6443/healthz` to select a working master instead?

jakolehm: We could check api health.. but it's not guaranteed to have any healthy master. Still we should prefer a master that has generated config (because of etcd, ca, etc..)

SpComb: > Still we should prefer a master that has generated config (because of etcd, ca, etc..)

But the `master` is not used for that, that uses whatever host the `ConfigureSecretsEncryption`, `ConfigureEtcdCa` etc phases run on first.

SpComb: > We could check api health.. but it's not guaranteed to have any healthy master

Having the kube-apiserver `/healthz` return HTTP 200 ok is more likely to indicate a healthy master than checking for the existence of `/etc/kubernetes/kubelet.conf`?

lib/pharos/cluster_manager.rb
```diff
@@ -40,30 +40,47 @@ def validate
...
+        apply_phase(Phases::ConfigureCfssl, config.etcd_hosts, ssh: true, parallel: true)
+        apply_phase(Phases::ConfigureEtcdCa, config.etcd_hosts[0...1], ssh: true, parallel: false)
+        apply_phase(Phases::ConfigureEtcdChanges, config.etcd_hosts[0...1], ssh: true, parallel: false)
+        apply_phase(Phases::ConfigureEtcd, config.etcd_hosts, ssh: true, parallel: true)
+      end
 
       apply_phase(Phases::ConfigureSecretsEncryption, config.master_hosts, ssh: true, parallel: false)
       apply_phase(Phases::ConfigureMaster, config.master_hosts, ssh: true, parallel: false)
-      apply_phase(Phases::MigrateWorker, config.worker_hosts, ssh: true, parallel: true, master: config.master_host)
+      apply_phase(Phases::MigrateWorker, config.worker_hosts, ssh: true, parallel: true, master: master)
```
SpComb: This is going to bug out if the master changes... I already fixed this as part of #278, I'll split that out as a separate fix PR.

SpComb: See #302

jakolehm:
@SpComb @jnummelin PTAL

lib/pharos/cluster_manager.rb
```diff
@@ -40,33 +40,76 @@ def validate
...
+    def sorted_master_hosts
+      @sorted_master_hosts ||= config.master_hosts.sort_by { |h| master_sort_score(h) }
+    end
+
+    # @param host [Pharos::Configuration::Host]
+    # @return [Integer]
+    def master_sort_score(host)
+      score = 0
+      return score if host.checks['api_healthy']
+      score += 1
```
SpComb: This control flow is really weird

```ruby
class Pharos::Configuration::Host
  def master_sort_score
    if checks['api_healthy']
      0
    elsif checks['kubelet_configured']
      1
    else
      2
    end
  end
end
```

lib/pharos/cluster_manager.rb
```diff
@@ -40,33 +40,76 @@ def validate
       addon_manager.validate
     end
 
+    # @return [Array<Pharos::Configuration::Host>]
+    def sorted_master_hosts
+      @sorted_master_hosts ||= config.master_hosts.sort_by { |h| master_sort_score(h) }
```
SpComb: I'm pretty sure that this should be an instance method on the `Pharos::Config` class?

SpComb: The `@sorted_master_hosts` memoization is also risky, because the sort scoring changes... although it would crash on the uninitialized `checks` in that case, so I guess that's good.

lib/pharos/etcd/client.rb
```diff
@@ -0,0 +1,47 @@
...
+        return false if result.error?
+
+        data = JSON.parse(result.stdout)
+        data['health'] == 'true'
+      end
+
+      # @return [Array<Hash>]
+      def members
+        result = @ssh.exec("#{CURL}/v2/members")
+        raise Error, "Cannot fetch etcd members" if result.error?
```
SpComb: This loses the `curl` `stderr` output? Why not just use `exec!` like the other methods?

jakolehm: I can change this.. reason for not using `exec!` is more clear error message.

SpComb: Yes, I imagine that having the complete `curl` command in the error message would look ugly... maybe have a `curl` helper method that raises `Error` and includes the requested URL + `result.stderr` from curl?

lib/pharos/cluster_manager.rb
```diff
@@ -40,33 +40,76 @@ def validate
...
 
       # master is now configured and can be used
-      apply_phase(Phases::ConfigureDNS, [config.master_host], master: config.master_host)
-      apply_phase(Phases::ConfigureWeave, [config.master_host], master: config.master_host) if config.network.provider == 'weave'
-      apply_phase(Phases::ConfigureCalico, [config.master_host], master: config.master_host) if config.network.provider == 'calico'
-      apply_phase(Phases::ConfigureMetrics, [config.master_host], master: config.master_host)
-      apply_phase(Phases::StoreClusterYAML, [config.master_host], master: config.master_host, config_content: @config_content)
-      apply_phase(Phases::ConfigureBootstrap, [config.master_host], ssh: true) # using `kubeadm token`, not the kube API
+      apply_phase(Phases::ConfigureDNS, [master_hosts.first], master: master_hosts.first)
+
```
SpComb: spurious empty line?

lib/pharos/scripts/configure-etcd.sh
```diff
@@ -83,9 +94,7 @@ EOF
 fi
 
 echo "Waiting etcd to launch on port 2380..."
-
-while ! nc -z ${PEER_IP} 2380; do
+while ! etcd_healthy; do
```
jnummelin: I think we should put some timeouts (5min or something long enough) around these sleep loops

lib/pharos/cluster_manager.rb
jnummelin: I agree, pretty hard to read as-is.

lib/pharos/cluster_manager.rb
```diff
@@ -40,33 +40,76 @@ def validate
...
     end
 
+    # @return [Array<Pharos::Configuration::Host>]
+    def sorted_master_hosts
+      @sorted_master_hosts ||= config.master_hosts.sort_by { |h| master_sort_score(h) }
+    end
+
+    # @param host [Pharos::Configuration::Host]
+    # @return [Integer]
+    def master_sort_score(host)
```
jnummelin: Some commentary why we sort the masters would be nice, I'm not sure if I can 100% understand why...

jnummelin:
Is there a docs PR already in place?

lib/pharos/phases/validate_host.rb
```diff
@@ -76,6 +77,30 @@ def cpu_arch
...
+      def master_checks
+        data = {}
+        result = @ssh.exec("sudo curl -sSf --connect-timeout 1 --cacert /etc/kubernetes/pki/ca.crt https://localhost:6443/healthz")
+        data['api_healthy'] = (result.success? && result.stdout == 'ok')
+        data['ca_exists'] = @ssh.file('/etc/kubernetes/pki/ca.key').exist?
+
+        unless @config.etcd&.endpoints
+          etcd = Pharos::Etcd::Client.new(@ssh)
+          data['etcd_healthy'] = etcd.healthy?
+          data['etcd_ca_exists'] = @ssh.file('/etc/pharos/pki/etcd/ca-key.pem').exist?
```
SpComb: Wrong path, should be `/etc/pharos/pki/ca-key.pem`


```
    $ sudo sh -c 'test -e /etc/pharos/pki/etcd/ca-key.pem && echo true || echo false'
    false
    ! 0
```

SpComb:
This PR also has a new requirement that (master) hosts machines must have unique hostnames... that should probably be explicitly validated before the etcd cluster setup fails.

jakolehm:
> This PR also has a new requirement that (master) hosts machines must have unique hostnames... that should probably be explicitly validated before the etcd cluster setup fails.

Pharos (Kubernetes) already has this requirement.. it's not specific to this PR. But I agree that we should validate it.

lib/pharos/etcd/client.rb
jakolehm: Etcd client is now refactored.

lib/pharos/cluster_manager.rb
jakolehm: I can remove memoization, it's not really needed.

lib/pharos/cluster_manager.rb
jakolehm: Fixed.

jnummelin:
> Not sure what kind of docs this PR needs. This fixes several cases that user expects to work.

Maybe something around the limitations, e.g. the case of role changes. Myself as a user stumbled on that. ;)

lib/pharos/cluster_manager.rb
jakolehm: Added some commentary.

jakolehm:
> Maybe something around the limitations, e.g. the case of role changes. Myself as a user stumbled on that. ;)

It's a different issue that we need to fix (at least have validation).

lib/pharos/phases/configure_etcd_changes.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      def remove_old_members
+        member_list = etcd.members
+        remove_members = member_list.reject { |m|
+          @config.etcd_hosts.find { |h|
+            m['peerURLs'] == ["https://#{h.peer_address}:2380"]
+          }
+        }
+        if remove_members.size / member_list.size.to_f >= 0.5
+          fail "Cannot remove majority of etcd peers"
```
SpComb: Not sure about why this restriction is needed... does this prevent you from replacing either node if you have two master hosts?

jakolehm: Yes, it should prevent it. If you remove 50% from etcd cluster then it goes into unhealthy state... not sure if we should allow it.

lib/pharos/phases/configure_etcd_changes.rb
```diff
@@ -0,0 +1,79 @@
...
+
+module Pharos
+  module Phases
+    class ConfigureEtcdChanges < Pharos::Phase
+      title 'Configure etcd member changes'
+
+      def call
+        store_initial_cluster_state
+
+        return if initial_cluster_state != 'existing' || !etcd.healthy?
```
SpComb: This can't be used to repair an unhealthy etcd cluster by adding a new member or replacing failed members?

jakolehm: Currently no, I have not tested recovery from unhealthy state.

SpComb: Right now the "recover from broken etcd cluster" case isn't working... but that might be because we have a `GET /v2/keys/foo` healthcheck on the etcd pod, so the kubelet ends up killing all the etcd members once the cluster goes unhealthy...

Once the `etcd` restarts, it seems like there's some kind of startup delay where it just times out all client requests if the cluster is unhealthy cluster... but I did eventually seem to get a response for `GET /v2/members` before the kubelet killed the etcd pod again, so I think it *might* be possible to recover an unhealthy cluster by replacing the failed member?

jakolehm: I tried this but it seems that broken cluster peer does not respond at all. I think it's out of scope of this PR to fix this broken etcd cluster.

SpComb: Still thinking this might behave better without the `etcd.healthy?` check... then it would fail right on the `Configure etcd member changes` => get member list, rather than continuing on to the `Configure etcd` and then just hanging because the new member can't join the cluster.

PR #294, opened 2018-04-23
Extra methods for testing various file/path features to Pharos::Transport::TransportFile

kke:
Adds question-mark methods for testing various file/path features:

```ruby
file = host.transport.file('/tmp/foo')
file.exist?
file.directory?
file.file?
file.writable?
file.socket?
file.symlink?
file.pipe?
file.readable?
file.executable?
file.non_zero?
```


lib/pharos/ssh/remote_file.rb
```diff
@@ -50,10 +50,10 @@ def read
         @client.exec!("sudo cat #{escaped_path}")
       end
 
-      # True if the file exists. Assumes a bash-like shell.
+      # True if the file exists
       # @return [Boolean]
       def exist?
-        @client.exec?("[ -e #{escaped_path} ]")
+        @client.exec!("sudo test -e #{escaped_path}")
```
jakolehm: why `!` instead of `?`

SpComb:
Note that this does not fix #295, it only workarounds it for `EACCES`... although I can't immediately think of how `stat` would otherwise realistically fail.

Although, this is now a false-negative in case `sudo ...` fails for whatever reason... still not ideal.

lib/pharos/ssh/remote_file.rb
kke: yeah it was not working either.

SpComb:
Could we have the trivial one-liner fix for backporting to 0.6?

kke:
Yes.

lib/pharos/ssh/remote_file.rb
```diff
@@ -147,7 +147,7 @@ def each_line
       private
 
       def test?(flag)
-        @client.exec?("sudo test -#{flag} #{escaped_path}")
+        @client.exec!("(sudo true || exit 1) && (sudo test -#{flag} #{escaped_path} && echo true || exit 0)").strip == "true"
```
SpComb: TOCTOU applies, running `sudo` twice won't cancel out random errors.

Something like `sudo sh -c 'test -e ... && echo true || echo false'` would actually help, though... I'm fairly confident that `test`/`stat` itself is unlikely to fail in ways that leads to destructive false-negatives.

kke:
close, merge or keep keeping up to date?

kke:
Still up to date, close or merge? @jakolehm

kke:
Got to be impressed by git, just merge master, no conflicts, everything seems to still be working even the whole ssh stuff has been refactored and moved around.

jakolehm:
Extra methods are good if they are used.. maybe this pr should refactor codebase so that they are actually used? (otherwise we just have unnecessary code)

PR #297, opened 2018-04-23
Avoid false-negatives when checking if file exists on remote host

### LINKED ISSUE #295: Pharos::SSH::RemoteFile#exists? returns false-negative for errors
If the `[ -e $path ]` fails for any other reason than `EEXIST`, this will still return `false`. That also includes `EACCES`, for cases where the file is not readable by the SSH user:

```
$ ls -l /etc/pharos/secrets-encryption/config.yml
ls: cannot access '/etc/pharos/secrets-encryption/config.yml': Permission denied
$ test -e /etc/pharos/secrets-encryption/config.yml && echo yay
```

This can cause bad side effects for phases that replace the file if it does not "exist".

Workaround for `EACCES` is to use `sudo test -e ...`, but this does not fix it for any other temporary `stat` errors...?

### LINKED ISSUE #296: pharos-cluster up corrupts encrypted secrets unless running as SSH root@
Affects 0.6.x

The `ConfigureSecretsEncryption` phase will attempt to read the `/etc/pharos/secrets-encryption/config.yml` file on the first master node, and use those keys on all master nodes for consistent encryption of secrets in etcd. If the file does not exist, such as when doing the initial install, then it will generate and write new secrets encryption keys.

However, #295 means that the `file.exists?` check will always incorrectly return `false` for the non-world-readable `/etc/pharos/secrets-encryption` dir, unless using SSH with the `root` user. This means that each run of `pharos-cluster up` will rewrite the `/etc/pharos/secrets-encryption/config.yml` file, which effectively corrupts all encrypted etcd secrets once the apiserver is restarted with the new keys.

```
==> Configure secrets encryption @ 167.99.36.141
    [167.99.36.141] Checking if secrets encryption is already configured ...
    $ [ -e /etc/pharos/secrets-encryption/config.yml ]
    ! 1
    [167.99.36.141] Generating new encryption keys ...
    $ [ -e /etc/pharos/secrets-encryption/config.yml ]
    ! 1
    [167.99.36.141] Creating secrets encryption configuration ...
    $ test -d /etc/pharos/secrets-encryption || sudo install -m 0700 -d /etc/pharos/secrets-encryption
    ! 0
    $ cat > /tmp/config.yml.b72612a37c7878b2792df05531eb6710 && (sudo mv /tmp/config.yml.b72612a37c7878b2792df05531eb6710 /etc/pharos/secrets-encryption/config.yml || (rm /tmp/config.yml.b72612a37c7878b2792df05531eb6710; exit 1))
    ! 0
    $ sudo chmod 0700 /etc/pharos/secrets-encryption/config.yml
    ! 0
    Completed Configure secrets encryption @ 167.99.36.141 in 0.956s
```

```
$ ls -l /etc/pharos/secrets-encryption/config.yml
ls: cannot access '/etc/pharos/secrets-encryption/config.yml': Permission denied
$ test -e /etc/pharos/secrets-encryption/config.yml && echo yay
$ sudo test -e /etc/pharos/secrets-encryption/config.yml && echo yay
yay
```

kke:
Fixes #296

### DIFF
--- a/lib/pharos/ssh/remote_file.rb
+++ b/lib/pharos/ssh/remote_file.rb
@@ -53,7 +53,7 @@ module Pharos
       # True if the file exists. Assumes a bash-like shell.
       # @return [Boolean]
       def exist?
-        @client.exec?("[ -e #{escaped_path} ]")
+        @client.exec!("(sudo true || exit 1) && (sudo test -e #{escaped_path} && echo true || exit 0)").strip == "true"
       end
 
       # Performs the block if the remote file exists, otherwise returns false
--- a/lib/pharos/ssh/remote_file.rb
+++ b/lib/pharos/ssh/remote_file.rb
@@ -53,7 +53,7 @@ module Pharos
       # True if the file exists. Assumes a bash-like shell.
       # @return [Boolean]
       def exist?
-        @client.exec!("(sudo true || exit 1) && (sudo test -e #{escaped_path} && echo true || exit 0)").strip == "true"
+        @client.exec!("sudo sh -c 'test -e #{escaped_path} && echo true || echo false'").strip == "true"
       end
 
       # Performs the block if the remote file exists, otherwise returns false


lib/pharos/ssh/remote_file.rb
```diff
@@ -53,7 +53,7 @@ def read
       # True if the file exists. Assumes a bash-like shell.
       # @return [Boolean]
       def exist?
-        @client.exec?("[ -e #{escaped_path} ]")
+        @client.exec!("(sudo true || exit 1) && (sudo test -e #{escaped_path} && echo true || exit 0)").strip == "true"
```
SpComb: No, the `sudo true && sudo test ...` is a bad idea...

Maybe use something like this to fail on `sudo` errors but trap the `test` error case:

```ruby
@client.exec!('sudo sh -c "test -e #{escaped_path} && echo true || echo false") == "true"` 
```

I'm fairly confident that `test`/`stat` itself is unlikely to fail without something else also breaking, so it should avoid false-positives.

PR #299, opened 2018-04-24
Update crio to 1.10 and fix it

jnummelin:
cri-o setup was totally broken, caused by etcd & kubelet proxy dropins.

Moved the kubelet args logic into `Configuration::Host` so that it can easily be re-used in different phases doing kubelet configs & "one-shot" runs.

### DIFF
--- a/lib/pharos/configuration/host.rb
+++ b/lib/pharos/configuration/host.rb
@@ -29,6 +29,34 @@ module Pharos
       def peer_address
         private_address || address
       end
+
+      def kubelet_args(local_only: false)
+        args = []
+        node_ip = private_address.nil? ? address : private_address
+
+        if crio?
+          args << '--container-runtime=remote'
+          args << '--runtime-request-timeout=15m'
+          args << '--container-runtime-endpoint=/var/run/crio/crio.sock'
+        end
+
+        if local_only
+          args << "--pod-manifest-path=/etc/kubernetes/manifests/"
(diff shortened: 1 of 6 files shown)


lib/pharos/configuration/host.rb
```diff
@@ -29,6 +29,34 @@ def api_address
       def peer_address
         private_address || address
       end
+
+      def kubelet_args(local_only: false)
```
SpComb: I don't think the `Pharos::Configuration::Host` is the right place to put kubelet command-line args.

jnummelin: My thinking was like:
each host runs kubelet -> host knows container runtime -> host knows needed kubelet args

Another option would be some sort of `KubeletHelper` that handles this.

Third option is to have the logic baked in the bash scripts. That would mean lot of duplication.

SpComb: Case in point: the `-cloud-provider=` is still in the `ConfigureKubelet` because it needs access to the `@config` - although this could be solved by copying the `@cloud.provider` configuration down to each host config.

I'd say this should be in some `KubeletHelper`, although with the #277 refactoring there shouldn't be any need to handle this outside of the `ConfigureKubelet` phase. We can move this code later, but I personally wouldn't encourage moving more and more of the kubeadm etc config options into the `Pharos::Configuration::Host` either.

PR #301, opened 2018-04-24
Verify that user and root default shell are bash

kke:


kke:
This probably unnecessarily breaks everything when root shell is dash.

PR #303, opened 2018-04-24
Discover host private interface address

### LINKED ISSUE #127: Add option to define private network interface
This way user does not need to know private_address.

SpComb:
Fixes #127 

Support `private_interface` as an alternative to `private_address`.

The `ValidateHost` phase discovers the private Ip using `ip addr`. This will fail if the interface does not exist, or continue without a private IP if the interface is not configured.

### DIFF
--- a/lib/pharos/phases/validate_host.rb
+++ b/lib/pharos/phases/validate_host.rb
@@ -36,6 +36,7 @@ module Pharos
         @host.os_release = os_release
         @host.cpu_arch = cpu_arch
         @host.hostname = hostname
+        @host.private_interface_address = private_interface_address(@host.private_interface) if @host.private_interface
       end
 
       # @return [String]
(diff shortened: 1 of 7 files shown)


lib/pharos/phases/validate_host.rb
```diff
@@ -74,6 +75,19 @@ def cpu_arch
...
         )
       end
+
+      # @param interface [String]
+      # @return [String]
+      def private_interface_address(interface)
+        @ssh.exec!("ip -o addr show dev #{interface} scope global | awk '{ print $4 }'").each_line do |line|
+          ip, prefixlen = line.split('/')
+
+          next if ip == @host.address
```
SpComb: Show allow this to work on the likes of `packet.net`, which configure both the public IP and private IP on the same `bond0` interface.

SpComb:
https://github.com/kontena/pharos-docs/pull/10

> Looks good. Should we prefer the use of private_interface in docs/examples?

There's room for both in the `cluster.example.yml`

lib/pharos/phases/validate_host.rb
```diff
@@ -74,6 +75,19 @@ def cpu_arch
           id: cpu['Architecture']
         )
       end
+
+      # @param interface [String]
+      # @return [String]
+      def private_interface_address(interface)
+        @ssh.exec!("ip -o addr show dev #{interface} scope global | awk '{ print $4 }'").each_line do |line|
```
jakolehm: I would prefer to do output parsing completely in the code (easier to write specs).

SpComb: spec'd

PR #306, opened 2018-04-25
Run all commands through bash

kke:
An alternative to #301 

Changes all execs to go through `bash --noprofile --norc -c `.


kke:
#590 pretty much replaced this

PR #308, opened 2018-04-25
Install crictl if using crio

jnummelin:
`crictl` is pretty low level, but allows some level of tooling to debug running containers on top of cri-o.

There's no deb packages available AFAIK, hence the direct download.


jnummelin:
@jakolehm would this be good for 1.1?

lib/pharos/scripts/configure-cri-o.sh
```diff
@@ -14,3 +14,8 @@ systemctl enable crio
 # remove unnecessary cni plugins
 rm /etc/cni/net.d/100-crio-bridge.conf /etc/cni/net.d/200-loopback.conf || true
 systemctl start crio
+
+# Install crictl binary
+curl -sSL https://github.com/kubernetes-incubator/cri-tools/releases/download/v1.0.0-beta.0/crictl-v1.0.0-beta.0-linux-amd64.tar.gz | tar xz
```
jakolehm: Can we verify this download somehow?

jnummelin: We could verify it with a known (== hardcoded) sha or something. I mean verify it manually  first and grab a sha of it and verify against that.

Other option is to package it in our own bintray repo as deb package.

lib/pharos/scripts/configure-cri-o.sh
```diff
@@ -14,3 +14,10 @@ systemctl enable crio
 # remove unnecessary cni plugins
 rm /etc/cni/net.d/100-crio-bridge.conf /etc/cni/net.d/200-loopback.conf || true
 systemctl start crio
+
+# Install crictl binary
+CRICTL_DOWNLOAD_SHA="597a4db0289870d81d0377396ddaf4c23725a47b33b30856e6291c2b958786f3"
+curl -sSL https://github.com/kubernetes-incubator/cri-tools/releases/download/v1.0.0-beta.0/crictl-v1.0.0-beta.0-linux-${CPU_ARCH}.tar.gz -o /tmp/crictl.tar.gz
```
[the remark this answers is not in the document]
jnummelin: Not really sure, the version matrix they have indicates that the `beta.1` would be for kube 1.11.x. https://github.com/kubernetes-incubator/cri-tools#current-status

lib/pharos/scripts/configure-cri-o.sh
```diff
@@ -14,3 +14,10 @@ systemctl enable crio
 # remove unnecessary cni plugins
 rm /etc/cni/net.d/100-crio-bridge.conf /etc/cni/net.d/200-loopback.conf || true
 systemctl start crio
+
+# Install crictl binary
+CRICTL_DOWNLOAD_SHA="597a4db0289870d81d0377396ddaf4c23725a47b33b30856e6291c2b958786f3"
+curl -sSL https://github.com/kubernetes-incubator/cri-tools/releases/download/v1.0.0-beta.0/crictl-${CRICTL_VERSION}-linux-${CPU_ARCH}.tar.gz -o /tmp/crictl.tar.gz
```
SpComb: This re-downloads/installs `crictl` for every `pharos-cluster up` run? Needs a conditional to check for the existence of `crictl`... and possibly also something for version changes?

lib/pharos/scripts/configure-cri-o.sh
```diff
@@ -14,3 +14,10 @@ systemctl enable crio
...
+echo "$CRICTL_DOWNLOAD_SHA  /tmp/crictl.tar.gz" | shasum -a256 -c
+tar xzf /tmp/crictl.tar.gz
+install -m 755 -o root -g root crictl /usr/bin/crictl && rm crictl
```
SpComb: Belongs in `/usr/local/bin`, I think.

lib/pharos/scripts/configure-cri-o.sh
jnummelin: done

lib/pharos/scripts/configure-cri-o.sh
jnummelin: fixed

lib/pharos/scripts/configure-cri-o.sh
```diff
@@ -14,3 +14,16 @@ systemctl enable crio
 # remove unnecessary cni plugins
 rm /etc/cni/net.d/100-crio-bridge.conf /etc/cni/net.d/200-loopback.conf || true
 systemctl start crio
+
+# Install crictl binary if needed
+CRICTL_DOWNLOAD_SHA="597a4db0289870d81d0377396ddaf4c23725a47b33b30856e6291c2b958786f3"
+CURRENT_VERSION=$(which crictl 1>/dev/null && criversion=$(crictl -v) && echo ${criversion##* })
+CRICTL_EXISTS=$?
```
SpComb: This could be simplified to something like

```sh
if ! which crictl > /dev/null || [ "$(crictl -v)" != "$CRICTL_VERSION" ]; then
  ...
fi
```

SpComb:
The pr build is passing, but the branch build is failing on #364, needs merge with master to go green... should be fine to merge regardless.

jnummelin:
merged with master, builds green :)

jnummelin:
Binary now taken from bintray. Also had to disable `cri` pre-flight checks as per https://github.com/kubernetes/kubeadm/issues/814 as it gives false negative.

@jakolehm @SpComb PTAL again.

lib/pharos_cluster.rb
```diff
@@ -9,6 +9,7 @@
 
 module Pharos
   CRIO_VERSION = '1.10'
+  CRICTL_VERSION = 'v1.0.0-beta.0'
```
jakolehm: Should be without `v`.

lib/pharos/scripts/configure-cri-o.sh
```diff
@@ -14,3 +14,14 @@ systemctl enable crio
...
 rm /etc/cni/net.d/100-crio-bridge.conf /etc/cni/net.d/200-loopback.conf || true
 systemctl start crio
+
+# Install crictl binary if needed
+CRICTL_DOWNLOAD_SHA="597a4db0289870d81d0377396ddaf4c23725a47b33b30856e6291c2b958786f3"
+
+if ! which crictl > /dev/null || [ "$(crictl -v)" != "$CRICTL_VERSION" ]; then
+    # Not installed or wrong version
+    curl -sSL https://bintray.com/kontena/pharos-bin/download_file?file_path=crictl-${CRICTL_VERSION}-linux-${CPU_ARCH}.tar.gz -o /tmp/crictl.tar.gz
+    echo "$CRICTL_DOWNLOAD_SHA  /tmp/crictl.tar.gz" | shasum -a256 -c
```
SpComb: I'm confused, wasn't this supposed to be using a signed `.deb`? The bintray repo has `.asc` signatures for the `.tar.gz' files, but this isn't checking those?

SpComb: NVM, I misread the earlier discussion... this is fine with a SHA256SUM, although a custom `.deb` in a signed APT repo would be ideal.

jnummelin: I'll make it use the .asc for verification

jnummelin: done

lib/pharos_cluster.rb
jnummelin: done

lib/pharos/phases/join_node.rb
```diff
@@ -22,7 +22,9 @@ def call
           join_command << '--cri-socket /var/run/crio/crio.sock'
         end
         join_command << "--node-name #{@host.hostname}"
-        join_command << "--ignore-preflight-errors DirAvailable--etc-kubernetes-manifests"
+        # CRI related pre-flight checks ignored as they cause false negative:
+        # https://github.com/kubernetes/kubeadm/issues/814
+        join_command << "--ignore-preflight-errors DirAvailable--etc-kubernetes-manifests,cri"
```
SpComb: Confirm that this broken in kubeadm 1.10.2 per https://github.com/kubernetes/kubernetes/pull/63297, but it looks like it's fixed in 1.10.3:

* https://github.com/kubernetes/kubernetes/blob/v1.10.2/cmd/kubeadm/app/preflight/checks.go#L107
* https://github.com/kubernetes/kubernetes/blob/v1.10.3/cmd/kubeadm/app/preflight/checks.go#L107

jnummelin: so, now that kube 1.10.3 is merged to master, I'll merge/rebase and remove the cri pre-flight check ignore

SpComb:
Note that per https://github.com/kubernetes/kubeadm/issues/814#issuecomment-392510572 having `crictl` installed will break `kubeadm join` for the default Docker runtime... shouldn't be a problem outside of testing setups and switching from `container_runtime: cri-o` back to the default.

PR #311, opened 2018-04-25
Validate hostname uniqueness

### LINKED ISSUE #307: pharos-cluster up should validate that hosts have unique hostnames
Requirement for #293 etcd clustering, and presumably kube itself as well.

jnummelin:
Added as separate phase to allow host fact gathering to be run in parallel.

fixes #307

### DIFF
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -52,7 +52,7 @@ module Pharos
 
     def apply_phases
       apply_phase(Phases::ValidateHost, config.hosts, ssh: true, parallel: true)
-      apply_phase(Phases::ValidateHostname, config.hosts, ssh: true, parallel: false, hosts: config.hosts)
+      apply_phase(Phases::ValidateHostname, config.hosts, ssh: true, parallel: false)
       # we need to use sorted masters because phases expects that first one has
       # ca etc config files
       master_hosts = sorted_master_hosts
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -52,7 +52,7 @@ module Pharos
 
     def apply_phases
       apply_phase(Phases::ValidateHost, config.hosts, ssh: true, parallel: true)
-
+      apply_phase(Phases::ValidateHostname, config.hosts, ssh: true, parallel: false, hosts: config.hosts)
       # we need to use sorted masters because phases expects that first one has
       # ca etc config files
       master_hosts = sorted_master_hosts
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -52,7 +52,7 @@ module Pharos
 
     def apply_phases
       apply_phase(Phases::ValidateHost, config.hosts, ssh: true, parallel: true)
-      apply_phase(Phases::ValidateHostname, config.hosts, ssh: true, parallel: false)
+      apply_phase(Phases::ValidateHostname, config.hosts, ssh: false, parallel: false)
       # we need to use sorted masters because phases expects that first one has
       # ca etc config files
       master_hosts = sorted_master_hosts
--- a/lib/pharos/phases/validate_hostname.rb
+++ b/lib/pharos/phases/validate_hostname.rb
@@ -5,19 +5,13 @@ module Pharos
     class ValidateHostname < Pharos::Phase
       title "Validate unique hostnames for all hosts"
 
-      # @param hosts Array<Pharos::Configuration::Host>
-      def initialize(host, hosts:, **options)
-        super(host, **options)
-        @hosts = hosts
-      end
-
       def call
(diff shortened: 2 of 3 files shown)


lib/pharos/phases/validate_hostname.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Pharos
+  module Phases
+    class ValidateHostname < Pharos::Phase
+      title "Validate unique hostnames for all hosts"
+
+      # @param hosts Array<Pharos::Configuration::Host>
```
kke: Phases already have `@config` so you could just use `@config.hosts` from the phase and ignore whatever is in `@host`.

lib/pharos/cluster_manager.rb
```diff
@@ -52,7 +52,7 @@ def sorted_etcd_hosts
 
     def apply_phases
       apply_phase(Phases::ValidateHost, config.hosts, ssh: true, parallel: true)
-
+      apply_phase(Phases::ValidateHostname, config.hosts, ssh: true, parallel: false)
```
SpComb: Doesn't need `ssh`?

spec/pharos/phases/validate_hostname_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+        config.hosts[0].hostname = "host-0"
+        config.hosts[1].hostname = "host-1"
+        config.hosts[2].hostname = "host-2"
+
+        subject.call
+      end
+    end
+
+    context 'duplicate hostnames' do
+      it 'raises if no duplicates' do
```
SpComb: raises if duplicates

spec/pharos/phases/validate_hostname_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+        ),
+      ],
+  ) }
+
+  subject { described_class.new(config.hosts[0], config: config, ssh: ssh) }
+
+
+  describe '#call' do
+    context 'no duplicate hostnames' do
+      it 'raises if no duplicates' do
```
SpComb: does not raise if no duplicates

lib/pharos/cluster_manager.rb
jnummelin: removed

spec/pharos/phases/validate_hostname_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+      end
+    end
+
+    context 'duplicate hostnames' do
+      it 'raises if no duplicates' do
+        config.hosts[0].hostname = "foo"
+        config.hosts[1].hostname = "foo"
+        config.hosts[2].hostname = "foo"
+
+        expect{ subject.call }.to raise_error(Pharos::InvalidHostError, "Duplicate hostname foo for hosts 192.0.2.2,192.0.2.3")
```
SpComb: Hmm... shouldn't this also include `192.0.2.1`?

jnummelin: IMO no. `192.0.2.1` is the host we're currently checking, so the list only shows other hosts that have same hostname with that.

spec/pharos/phases/validate_hostname_spec.rb
jnummelin: copy-pasta strikes again :) fixed now

spec/pharos/phases/validate_hostname_spec.rb
SpComb: Hmm... but if running with `parallel: false`, the `PhaseManager` currently doesn't actually log what host failed, I think... maybe that's something that should be fixed up there. With `parallel: true`, the `PhaseManager` does log the error with the failing host before letting it get raisd.

SpComb: So the actual error output ATM looks like this, it doesn't say which two hosts had duplicate hostnames:

```
==> Validate unique hostnames for all hosts @ 167.99.36.141 167.99.36.158 167.99.36.223
    [167.99.36.141] Validating hostname uniqueness ...
    [167.99.36.158] Validating hostname uniqueness ...
Pharos::InvalidHostError : Duplicate hostname terom-kupo-node for hosts 167.99.36.223
```

But I'd consider that a `PhaseManager` bug.

jnummelin: I'd interpret that as `167.99.36.158`  having same hostname with `167.99.36.223`. Is that not true in this case?

SpComb: Yes, you need to match up the error with the previous line... it's not obvious IMO, particularly since it's not the same for the parallel phase output/error case.

jnummelin: For me it was obvious. :) Yes, the error reporting could be enhanced at `PhaseManager` level.

PR #317, opened 2018-04-26
Store complete cluster.yml to configmap

jakolehm:
Stored cluster.yml is currently missing `hosts` if they are read from terraform json file. This PR fixes it.


jakolehm:
@kke PTAL

PR #320, opened 2018-04-26
Scale default backend

### LINKED ISSUE #40: Scale ingress-nginx default backends

jnummelin:
One per two workers might be bit overkill?

fixes #40

### DIFF
--- a/lib/pharos/addons/ingress_nginx.rb
+++ b/lib/pharos/addons/ingress_nginx.rb
@@ -37,7 +37,7 @@ module Pharos
       # ~One replica per two workers
       # @return [Integer]
       def replicas
-        1 + (@cluster_config.worker_hosts.size / 2)
+        2 + (@cluster_config.worker_hosts.size / 10)
       end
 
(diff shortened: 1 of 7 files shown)


jakolehm:
> One per two workers might be bit overkill?

Probably yes, maybe 1 per 10 or so workers (with minimum of 2)?

lib/pharos/addons/ingress_nginx.rb
```diff
@@ -34,11 +34,21 @@ def image_name
         end
       end
 
+      # ~One replica per 10 workers, min 2
+      # @return [Integer]
+      def replicas
```
jakolehm: Naming is a bit misleading.. this is default backend replicas, right?

jnummelin: yes, I'll rename it

jnummelin: fixed

PR #322, opened 2018-04-26
Configurable calico ipip_mode

### LINKED ISSUE #315: Calico IP-IP tunneling should be optional
Some deployments (AWS, packet.net layer2) will be able to use direct routing instead, and avoid the CPU overhead of IP-in-IP.

https://docs.projectcalico.org/v3.1/usage/configuration/ip-in-ip

https://docs.projectcalico.org/v3.1/reference/node/configuration `CALICO_IPV4POOL_IPIP`

https://docs.projectcalico.org/v3.1/reference/calicoctl/resources/ippool `ipipMode`

SpComb:
Fixes #315

Allow configuring `network.calico.ipip_mode` to `Never`, `CrossSubnet`, `Always` (default).

If configured to `Never`, then the `ipip` setup (`tunl0` iface) is disabled completely.

Changing the `ipip_mode` is not supported... it requires resetting the calico ipPool (`kubectl delete ippools.crd.projectcalico.org default-ipv4-ippool`) and restarting the calico-node pods.

### DIFF
--- a/lib/pharos/phases/configure_calico.rb
+++ b/lib/pharos/phases/configure_calico.rb
@@ -31,6 +31,7 @@ module Pharos
           @master.api_address, 'calico',
           ipv4_pool_cidr: @config.network.pod_network_cidr,
           ipip_mode: @config.network.calico&.ipip_mode || 'Always',
+          ipip_enabled: @config.network.calico&.ipip_mode != 'Never',
           master_ip: @config.master_host.peer_address,
           version: CALICO_VERSION
         )
--- a/lib/pharos/phases/configure_calico.rb
+++ b/lib/pharos/phases/configure_calico.rb
@@ -30,6 +30,7 @@ module Pharos
         Pharos::Kube.apply_stack(
           @master.api_address, 'calico',
           ipv4_pool_cidr: @config.network.pod_network_cidr,
+          ipip_mode: @config.network.calico&.ipip_mode || 'Always',
           master_ip: @config.master_host.peer_address,
           version: CALICO_VERSION
         )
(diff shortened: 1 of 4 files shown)


SpComb:
Tested on packet.net layer2 using both `ipip_mode: CrossSubnet` and `ipip_mode: Never`:

```
$ ip ro
10.32.0.0/24 via 10.255.0.1 dev eno2  proto bird 
```

```
$ kubectl -n kube-system exec -c calico-node calico-node-vj2qp cat /etc/calico/confd/config/bird_ipam.cfg
# Generated by confd
filter calico_pools {
  calico_aggr();

  if ( net ~ 10.32.0.0/12 ) then {
    accept;
  }

  reject;
}


filter calico_ipip {

  if ( net ~ 10.32.0.0/12 ) then {

    if defined(bgp_next_hop) && ( bgp_next_hop ~ 10.255.0.0/24 ) then
      krt_tunnel = "";                     
    else
      krt_tunnel = "tunl0";       
    accept;
  }  

  accept;                                  
}
```

lib/pharos/phases/configure_calico.rb
```diff
@@ -30,6 +30,8 @@ def call
         Pharos::Kube.apply_stack(
           @master.api_address, 'calico',
           ipv4_pool_cidr: @config.network.pod_network_cidr,
+          ipip_mode: @config.network.calico&.ipip_mode || 'Always',
```
SpComb: This is annoying, but the `.default(...)` value defined on the struct attribute is mostly useless, because the struct itself doesn't have a default value... dealing with the configs would be so much easier if the nested structs also had default values.

PR #323, opened 2018-04-26
Disable only previously enabled addons

jakolehm:
It was annoying to see (and wait) "disabling addon ..." messages on every run. This PR adds a bit more logic to `AddonManager` so that we only disable addon if it was previously enabled.

PR also includes a bit more clean cluster context implementation.

### DIFF
--- a/lib/pharos/phases/configure_client.rb
+++ b/lib/pharos/phases/configure_client.rb
@@ -5,18 +5,54 @@ module Pharos
     class ConfigureClient < Pharos::Phase
       title "Configure kube client"
 
+      REMOTE_FILE = "/etc/kubernetes/admin.conf"
+
       def call
+        save_config_locally
+        config_map = previous_config_map
+        if config_map
+          cluster_context['previous-config-map'] = config_map
+          cluster_context['previous-config'] = build_config(config_map)
+        end
+      end
+
+      def save_config_locally
         Dir.mkdir(config_dir, 0o700) unless Dir.exist?(config_dir)
-        config_file = File.join(config_dir, @host.api_address)
+
         logger.info { "Fetching kubectl config ..." }
-        config_data = @ssh.file("/etc/kubernetes/admin.conf").read
+        config_data = remote_config_file.read
         File.chmod(0o600, config_file) if File.exist?(config_file)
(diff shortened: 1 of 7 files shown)


jakolehm:
@SpComb @kke PTAL

SpComb:
The new `pharos-config` should only be written after the addons have been cleaned up... otherwise any failure between the `Store cluster YAML` and the `Disabling addon ...` will cause those addons to be left running.

lib/pharos/phases/configure_client.rb
```diff
@@ -5,19 +5,55 @@ module Phases
...
+      end
+
       def config_dir
         File.join(Dir.home, '.pharos')
       end
+
+      # @param configmap [Kubeclient::Resource]
+      # @return [Pharos::Config]
+      def build_config(configmap)
+        cluster_config = YAML.safe_load(configmap.data['cluster.yml'])
```
kke: `Pharos::YamlFile.new(StringIO.new(configmap.data['cluster.yml']), override_filename: "#{@host}:cluster.yml").load`

jakolehm:
> The new pharos-config should only be written after the addons have been cleaned up... otherwise any failure between the Store cluster YAML and the Disabling addon ... will cause those addons to be left running.

I agree, now config is saved after addons have completed.

```
==> Label nodes @ 46.101.146.73 46.101.96.46
    [46.101.146.73] No labels set ... 
    [46.101.96.46] No labels set ... 
==> Configuring addons ...
==> Enabling addon ingress-nginx
==> Store cluster YAML @ 46.101.146.73
    [46.101.146.73] Storing cluster configuration to configmap
==> Cluster has been crafted! (took 1 minute)
```

lib/pharos/phases/configure_client.rb
jakolehm: why? 😃

kke: not any prettier but will use the same yaml-loading rules as everything else (altough the config is from `YAML.dump(JSON.parse(JSON.dump(config_hash)))` and shouldn't therefore include any trickery)

jakolehm: Fixed.

PR #326, opened 2018-04-26
Fix stack prune

### LINKED ISSUE #18: Prune does not remove all pods
Although it has `propagationPolicy: 'Foreground'`

SpComb:
Fixes  #18

* `Pharos::Kube::Stack#prune` was not enumerating the core `v1` api resources
* `Pharos::Kube::Resource#delete` was not setting any `DeleteOptions` `propagationPolicy` if the resource had the `metadata.selfLink`

Change the `Pharos::Kube::Stack#prune` to enumerate each API resource type in turn, fixing up the resource list items to include the correct `apiVersion` and `kind`, such that the `selfLink` is no longer required.

The stack prune now also enumerates all of the core API resource types at the end, which also means that stack prunes now takes even longer.

### DIFF
--- a/lib/pharos/kube/client.rb
+++ b/lib/pharos/kube/client.rb
@@ -71,6 +71,17 @@ module Pharos
 
         get_entity(definition.resource_name, resource.metadata.name, resource.metadata.namespace)
       end
+
+      # @param resource [Kubeclient::Resource]
+      def delete_resource(resource, propagationPolicy: 'Foreground')
+        definition = entity_for_resource(resource)
+
+        delete_entity(definition.resource_name, resource.metadata.name, resource.metadata.namespace, delete_options: {
+          apiVersion: 'v1',
+          kind: 'DeleteOptions',
+          propagationPolicy: propagationPolicy
+        })
+      end
     end
   end
 end
--- a/lib/pharos/kube/client.rb
+++ b/lib/pharos/kube/client.rb
@@ -73,14 +73,19 @@ module Pharos
       end
 
       # @param resource [Kubeclient::Resource]
-      def delete_resource(resource, propagationPolicy: 'Foreground')
+      def delete_resource(resource, propagation_policy: 'Foreground')
         definition = entity_for_resource(resource)
 
-        delete_entity(definition.resource_name, resource.metadata.name, resource.metadata.namespace, delete_options: {
-          apiVersion: 'v1',
-          kind: 'DeleteOptions',
-          propagationPolicy: propagationPolicy
-        })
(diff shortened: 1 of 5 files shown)


lib/pharos/kube/stack.rb
```diff
@@ -53,23 +52,29 @@ def apply
...
 
-        @session.api_groups.each do |api_group|
-          group_client = @session.resource_client(api_group.preferredVersion.groupVersion)
+        @session.api_versions.each do |api_version|
+          client = @session.resource_client(api_version)
+          client.entities.each do |method_name, entity|
+            next if method_name.end_with?('_review')
+            next if api_version == 'v1' && entity.resource_name == 'bindings' # XXX: the entity definition does not have the list verb... but kubeclient does not expose that
+            next if api_version == 'v1' && entity.resource_name == 'componentstatuses' # apiserver ignores the ?labelSelector query
+            next if api_version == 'v1' && entity.resource_name == 'endpoints' # inherits stack labels from service, does not have any ownerReference...
```
SpComb: Only way I can think of avoiding having to hardcode this would be to skip resources that have an `ownerReference` that would cause them to get deleted later during the prune... but service endpoints don't have any `ownerReference`, they must get GC'd magically :(

SpComb: Maybe prune should not delete resources that do not have the `pharos.kontena.io/stack-checksum` annotation? At least not when pruning as part of a stack apply, maybe as part of an addon uninstall => full prune?

lib/pharos/kube/client.rb
```diff
@@ -71,6 +71,17 @@ def get_resource(resource)
 
         get_entity(definition.resource_name, resource.metadata.name, resource.metadata.namespace)
       end
+
+      # @param resource [Kubeclient::Resource]
+      def delete_resource(resource, propagationPolicy: 'Foreground')
+        definition = entity_for_resource(resource)
+
+        delete_entity(definition.resource_name, resource.metadata.name, resource.metadata.namespace, delete_options: {
+          apiVersion: 'v1',
```
jakolehm: Does the hardcoded `v1` work here?

SpComb: This is from the original [`Pharos::Kube::Resource#delete`](https://github.com/kontena/pharos-cluster/pull/326/files#diff-2582468926bd28a4186bc42425aa7b42L68)... AFAIK it's the version of the `DeleteOptions` itself, not of the resource.

jakolehm: Right.

PR #331, opened 2018-04-27
Prevent role change for existing node

### LINKED ISSUE #309: Don't allow to change host role
Because we don't really support it.

jnummelin:
Prevent changing existing nodes role:
```
==> Reading instructions ...
==> Sharpening tools ...
==> Starting to craft cluster ...
==> Validate hosts @ 188.166.90.177 178.62.212.223 188.166.36.138
    [188.166.36.138] Checking sudo access ...
    [188.166.90.177] Checking sudo access ...
    [178.62.212.223] Checking sudo access ...
    [178.62.212.223] Gathering host facts ...
    [188.166.90.177] Gathering host facts ...
    [188.166.36.138] Gathering host facts ...
    [178.62.212.223] Validating current role matches ...
 [Validate hosts @ 178.62.212.223] Pharos::InvalidHostError: Cannot change role of an existing node
    [188.166.90.177] Validating current role matches ...
    [188.166.90.177] Validating distro and version ...
    [188.166.90.177] Validating host configuration ...
Pharos::InvalidHostError : Cannot change role of an existing node
```

fixes #309

### DIFF
--- a/lib/pharos/phases/validate_host.rb
+++ b/lib/pharos/phases/validate_host.rb
@@ -10,6 +10,8 @@ module Pharos
         check_sudo
         logger.info { "Gathering host facts ..." }
         gather_host_facts
+        logger.info { "Validating current role matches ..." }
+        check_role
         logger.info { "Validating distro and version ..." }
         check_distro_version
         logger.info { "Validating host configuration ..." }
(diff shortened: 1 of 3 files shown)


jnummelin:
> Can't we use the stored cluster yaml to figure out the previous roles?

We could, all these checks were already in place and provides reliable way to determine this so didn't want to go there. :)

lib/pharos/phases/validate_host.rb
```diff
@@ -36,10 +38,17 @@ def gather_host_facts
...
         @host.cpu_arch = cpu_arch
         @host.hostname = hostname
-        @host.checks = @host.role == 'master' ? master_checks : worker_checks
+        @host.checks = host_checks
         @host.private_interface_address = private_interface_address(@host.private_interface) if @host.private_interface
       end
 
+      def check_role
+        return if @host.master? && @host.checks['ca_exists'] == true
+        return if @host.worker? && @host.checks['kubelet_configured'] && @host.checks['ca_exists'] != true
```
SpComb: I think the check for `@host.checks['kubelet_configured']` here is the wrong way around - this only accepts if the kubelet is configured, which is the broken part.

Issue #305, opened 2018-04-25
Host role change does not really work


jnummelin:
With the master re-size coming in from #293 the change or host roles does not really work. There's two issues:

1. Etcd gets left behind on host changed from master -> worker
```
etcd-jussi-worker-1                         0/1       CrashLoopBackOff   6          23m       178.62.212.223   jussi-worker-1
```

It's in `CrashLoopBackOff` since that member has been "evicted" from etcd cluster but still the static manifest for is remains in `/etc/kubernetes/manifests/pharos-proxy.yaml`

2. Apiserver won't start on host changed from worker -> master

```
root@jussi-worker-2:~# ./crictl logs b76f848f5afbb
Flag --admission-control has been deprecated, Use --enable-admission-plugins or --disable-admission-plugins instead. Will be removed in a future version.
Flag --insecure-port has been deprecated, This flag will be removed in a future version.
I0425 07:13:21.059001       1 server.go:135] Version: v1.10.1
I0425 07:13:21.059166       1 server.go:717] external host was not specified, using 188.166.36.138
failed to create listener: failed to listen on 0.0.0.0:6443: listen tcp 0.0.0.0:6443: bind: address already in use
```

That's because the workers kubelet-proxy is still running and hogging the port.

---

First step would be to disable the possibility to do role changes in validation phase.

Second step would be to run some sort of "reset" and remove all things not needed for the role.

SpComb:
Ref #309, I suppose this still remains open as a feature request for resetting the existing node configuration when changing roles?

jnummelin:
I think we can close this and open a new feature request on node role changes if that seems to be needed.

PR #336, opened 2018-04-27
Fix host role checks to skip unconfigured hosts

### LINKED ISSUE #332: Cannot add a fresh master
Regression from #331.

```
Pharos::InvalidHostError : Cannot change role of an existing node
```

SpComb:
Fixes #332

Minimal fix to skip the role checks if the kubelet was not yet configured. Tested to worker for `role` changes after a succesfull `pharos-cluster up`.

### DIFF
--- a/lib/pharos/phases/validate_host.rb
+++ b/lib/pharos/phases/validate_host.rb
@@ -43,8 +43,9 @@ module Pharos
       end
 
       def check_role
+        return if !@host.checks['kubelet_configured']
         return if @host.master? && @host.checks['ca_exists'] == true
-        return if @host.worker? && @host.checks['kubelet_configured'] && @host.checks['ca_exists'] != true
+        return if @host.worker? && @host.checks['ca_exists'] != true
 
         raise Pharos::InvalidHostError, "Cannot change role of an existing node"
       end
--- a/lib/pharos/phases/validate_host.rb
+++ b/lib/pharos/phases/validate_host.rb
@@ -44,10 +44,12 @@ module Pharos
 
       def check_role
         return if !@host.checks['kubelet_configured']
-        return if @host.master? && @host.checks['ca_exists'] == true
-        return if @host.worker? && @host.checks['ca_exists'] != true
 
-        raise Pharos::InvalidHostError, "Cannot change role of an existing node"
+        if @host.master? && !@host.checks['ca_exists']
+          raise Pharos::InvalidHostError, "Cannot change worker host role to master"
+        elsif @host.worker? && @host.checks['ca_exists']
+          raise Pharos::InvalidHostError, "Cannot change master host role to worker"
+        end
       end
 
       # @return [String]
--- a/lib/pharos/phases/validate_host.rb
+++ b/lib/pharos/phases/validate_host.rb
@@ -45,11 +45,10 @@ module Pharos
       def check_role
         return if !@host.checks['kubelet_configured']
 
-        if @host.master? && !@host.checks['ca_exists']
-          raise Pharos::InvalidHostError, "Cannot change worker host role to master"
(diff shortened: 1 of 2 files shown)


lib/pharos/phases/validate_host.rb
```diff
@@ -89,15 +92,18 @@ def cpu_arch
...
         data = {}
-        result = @ssh.exec("sudo curl -sSf --connect-timeout 1 --cacert /etc/kubernetes/pki/ca.crt https://localhost:6443/healthz")
-        data['api_healthy'] = (result.success? && result.stdout == 'ok')
-        data['ca_exists'] = @ssh.file('/etc/kubernetes/pki/ca.key').exist?
         data['kubelet_configured'] = @ssh.file('/etc/kubernetes/kubelet.conf').exist?
+        data['ca_exists'] = @ssh.file('/etc/kubernetes/pki/ca.key').exist?
+        data['etcd_ca_exists'] = @ssh.file('/etc/pharos/pki/ca-key.pem').exist?
+
+        if data['ca_exists']
+          result = @ssh.exec("sudo curl -sSf --connect-timeout 1 --cacert /etc/kubernetes/pki/ca.crt https://localhost:6443/healthz")
```
SpComb: Unrelated change, but we don't need to bother with the `curl` if the CA cert that it's using doesn't exist?

lib/pharos/phases/validate_host.rb
```diff
@@ -43,10 +43,13 @@ def gather_host_facts
       end
 
       def check_role
-        return if @host.master? && @host.checks['ca_exists'] == true
-        return if @host.worker? && @host.checks['kubelet_configured'] && @host.checks['ca_exists'] != true
+        return if !@host.checks['kubelet_configured']
```
SpComb: This will not work correctly if the cluster setup failed after the `ConfigureKubelet`/`ConfigureEtcd` phases but before the `ConfigureMaster`/`JoinNode` phases, and the role was then changed... Might be better to check for the `/etc/kubernetes/manifests/*` files instead ( `pharos-proxy`, `pharos-etcd` and `kube-apiserver`), since AFAIK those are what actually cause the actual role incompatibilities per #305.

lib/pharos/phases/validate_host.rb
```diff
@@ -43,10 +43,13 @@ def gather_host_facts
...
 
-        raise Pharos::InvalidHostError, "Cannot change role of an existing node"
+        if @host.master? && !@host.checks['ca_exists']
```
SpComb: I'm guessing the rubocop suggestino to use a "guard clause instead" isn't just referring to `return`, because I don't see how to transform this code with the two different `raise` exceptions to use `return`... but `raise ... if ...` does work, and I guess that also coutns as a "guard clause"

```
lib/pharos/phases/validate_host.rb:48:9: C: Style/GuardClause: Use a guard clause instead of wrapping the code inside a conditional expression.
        if @host.master? && !@host.checks['ca_exists']
        ^^

```

lib/pharos/phases/validate_host.rb
jakolehm: True.

spec/pharos/phases/validate_host_spec.rb
```diff
@@ -53,55 +53,99 @@
...
+          'ca_exists' => false,
+          'api_healthy' => false,
+          'kubelet_configured' => false
+        } }
 
-      subject = described_class.new(worker, config: config, ssh: ssh)
-      expect{ subject.check_role }.not_to raise_error
-    end
+        it 'does not raise' do
+          expect{ subject.check_role }.to_not raise_error(Pharos::InvalidHostError)
```
SpComb: And `expect{...}.to_not raise_error(...)` is also a bad idea...

PR #339, opened 2018-04-28
Install nfs-common as part of configure-kube

### LINKED ISSUE #338: nfs-common is missing
Without nfs-common it's not possible to mount NFS persistent volumes.

jakolehm:
Fixes #338


lib/pharos/scripts/configure-kube.sh
```diff
@@ -17,4 +17,9 @@ fi
 export DEBIAN_FRONTEND=noninteractive
 apt-mark unhold kubelet kubectl kubeadm
 apt-get install -y kubelet=${KUBE_VERSION}-00 kubectl=${KUBE_VERSION}-00 kubeadm=${KUBEADM_VERSION}-00
-apt-mark hold kubelet kubectl kubeadm
\ No newline at end of file
+apt-mark hold kubelet kubectl kubeadm
+
+if ! dpkg -s nfs-common > /dev/null; then
+    systemctl mask rpcbind
```
kke: Doesn't this prevent running an nfs-server?

jakolehm: I don't think this prevents running a nfs-server in a container? (which should be "the right way")

PR #345, opened 2018-05-03
[WIP] Add openebs add-on

### LINKED ISSUE #171: OpenEBS addon
https://github.com/openebs/openebs/

jnummelin:
Still bit of WIP. 

Need to figure out how to make the stack apply to withstand re-runs. Currently it fails with:
```
Kubeclient::HttpError : StorageClass.storage.k8s.io "openebs-standard" is invalid: parameters: Forbidden: updates to parameters are forbidden.
```

Not sure on the namespace we should put this into. `default` feels just plain wrong, maybe `kube-system` would be better for this kind of "infra" things.

fixes #171

### DIFF
--- a/lib/pharos/addons/open_ebs.rb
+++ b/lib/pharos/addons/open_ebs.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+module Pharos
+  module Addons
+    class OpenEbs < Pharos::Addon
+      name 'openebs'
+      version '0.5.3'
+      license 'Apache License 2.0'
+
+      struct {
+        attribute :default_replicas, Pharos::Types::Int
+        attribute :default_capacity, Pharos::Types::String
+      }
+
+      schema {
+        optional(:default_replicas).filled(:int?)
+        optional(:default_capacity).filled(:str?)
+      }
+
+      def validate
+        super
+        raise Pharos::InvalidAddonError, "Cannot set more replicas than workers" if config.default_replicas && config.default_replicas > cluster_config.worker_hosts.count
+      end
+
+      def default_replica_count
+        cluster_config.worker_hosts.count < 3 ? cluster_config.worker_hosts.count : 3
+      end
+
+      def install
+        apply_stack(
+          default_replicas: config.default_replicas || default_replica_count,
+          default_capacity: config.default_capacity || '5G'
+        )
+      end
+    end
+  end
+end
--- a/lib/pharos/resources/openebs/12-provisioner-deployment.yml.erb
+++ b/lib/pharos/resources/openebs/12-provisioner-deployment.yml.erb
@@ -1,0 +1,28 @@
+apiVersion: apps/v1beta1
+kind: Deployment
+metadata:
+  name: openebs-provisioner
+  namespace: default
+spec:
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        name: openebs-provisioner
+    spec:
+      serviceAccountName: openebs-maya-operator
+      containers:
+      - name: openebs-provisioner
+        imagePullPolicy: Always
+        image: openebs/openebs-k8s-provisioner:<%= version %>
+        env:
+        - name: NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: OPENEBS_MONITOR_URL
+          value: "http://127.0.0.1:32515/dashboard/db/openebs-volume-stats?orgId=1"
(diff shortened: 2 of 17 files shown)


lib/pharos/resources/openebs/22-storage-class.yml.erb
```diff
@@ -0,0 +1,10 @@
+apiVersion: storage.k8s.io/v1
+kind: StorageClass
```
jakolehm: Maybe we should have option to set this as the default storage class.

jnummelin: done

jakolehm:
> Not sure on the namespace we should put this into. default feels just plain wrong, maybe kube-system would be better for this kind of "infra" things.

IMHO `storage` (or similar) namespace would make more sense here.. not good idea to dump everything into `kube-system`.

jnummelin:
Changed now to use `storage` namespace.

Also added option to change the default storage pool path.

jnummelin:
docs PR: https://github.com/kontena/pharos-docs/pull/27

lib/pharos/resources/openebs/12-provisioner-deployment.yml.erb
```diff
@@ -0,0 +1,32 @@
...
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        name: openebs-provisioner
+    spec:
+      serviceAccountName: openebs-maya-operator
+      containers:
+      - name: openebs-provisioner
+        imagePullPolicy: Always
```
jakolehm: always?

lib/pharos/resources/openebs/30-default-storage-pool.yml.erb
```diff
@@ -0,0 +1,12 @@
+---
+apiVersion: openebs.io/v1alpha1
+kind: StoragePool
+metadata:
+  name: default
+  type: hostdir
+spec:
+  path: <%= default_storage_pool_path %>
+  nodes: [ "all" ]
+  #maxVolSize: 10G #Max Volume Size
```
jakolehm: Why are these commented out?

jnummelin: I not 100% sure how they affect, so had them still left commented out there. Probably there should be some config option for these.

lib/pharos/resources/openebs/30-default-storage-pool.yml.erb
```diff
@@ -0,0 +1,12 @@
...
+  #maxAllocatedSize: 100G #Max Total Space allocated from this directory/mount
+---
```
kke: Trailing document separator

lib/pharos/resources/openebs/12-provisioner-deployment.yml.erb
jnummelin: just blindly copied from their examples, probably `IfNotExist` would be better

lib/pharos/resources/openebs/30-default-storage-pool.yml.erb
jnummelin: removed them now. we could add more options in later PR's, IMO no need to put everything into this initial PR.

lib/pharos/resources/openebs/30-default-storage-pool.yml.erb
jnummelin: removed, leftover from initial try to put many resources into same yaml file.

lib/pharos/addons/open_ebs.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      struct {
+        attribute :default_replicas, Pharos::Types::Int
+        attribute :default_capacity, Pharos::Types::String
+        attribute :default_class, Pharos::Types::Bool
+        attribute :default_storage_path, Pharos::Types::String
+      }
+
+      schema {
+        optional(:default_replicas).filled(:int?)
```
jakolehm: I don't like repetition of `default_`. These are all `StorageClass` options?

jnummelin: Yes, they're mostly for the "default" `StorageClass` that gets created during the add-on setup. `default_replicas`, but I _think_ it's a value that gets used if the storage class does not provide amount of replicas.

Would it make more sense to move these to something like:
```
openebs:
  enabled: true
  storage_class:
    size: 50G
    replicas: 3
    ...
```
??

PR #347, opened 2018-05-04
Add config option to enable kubelet read-only port

### LINKED ISSUE #344: Option to enable kubelet read only port
There are some monitoring tools that do not cope at all with secured kubelet. To enable their usage, we should add an option to enable the kubelet read-only port.

Extra care should be used in documentation so that people really understand they must secure the access to that port in other means if we leave it running open.

jnummelin:
Introduces new top level `kubelet` config key to collec all possible future kubelet specific configs.

With
```
kubelet:
  read_only_port: true
```
it enables kubelet read-only port (`10255`), disables it by default.

fixes #344

### DIFF
--- a/lib/pharos/config_schema.rb
+++ b/lib/pharos/config_schema.rb
@@ -89,6 +89,9 @@ module Pharos
           optional(:mode).filled(included_in?: %w(userspace iptables ipvs))
         end
         optional(:addons).value(type?: Hash)
+        optional(:kubelet).schema do
+          optional(:read_only_port).filled(:bool?)
+        end
 
         validate(network_dns_replicas: [:network, :hosts]) do |network, hosts|
           if network && network[:dns_replicas]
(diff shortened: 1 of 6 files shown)


lib/pharos/config_schema.rb
```diff
@@ -89,6 +89,9 @@ def self.messages
           optional(:mode).filled(included_in?: %w(userspace iptables ipvs))
         end
         optional(:addons).value(type?: Hash)
+        optional(:kubelet).schema do
+          optional(:read_only_port).filled(:bool?)
```
jakolehm: Should this be an integer (default `0`)?

jakolehm: Right, maybe it's easier to just have a boolean for now.

jnummelin: yeah, that's the reason I went with boolean for now :)

PR #348, opened 2018-05-04
Fix horizontal-pod-autoscaler-use-rest-clients

jakolehm:
Regression from #330


jnummelin:
needs rebase/merge though to make build pass with the fixed dry stuff

PR #349, opened 2018-05-04
Pin dry-versions

jnummelin:
Latest 0.5.0 dry stuff breaks the API, we need to pin the versions. dry-validation and dry-struct seem to have incompatible version dependency to dry-types, hence it's pinned separately.

Later a proper 0.5.x compatible change PR should be made, there seems to be some handy new features.


jnummelin:
@kke PTAL

PR #358, opened 2018-05-08
Refactor addons folder structure

jakolehm:


jakolehm:
@kke PTAL

lib/pharos/kube/session.rb
```diff
@@ -20,10 +20,11 @@ def resource(resource)
 
       # Returns a new stack associated with this session
       # @param name [String]
+      # @param resource_path [String]
       # @param vars [Hash]
       # @return [Pharos::Kube::Stack]
-      def stack(name, vars = {})
-        Stack.new(self, name, vars)
+      def stack(name, resource_path, vars = {})
+        Stack.new(self, name, resource_path, vars)
```
kke: If the resource path is required, I think the name could be derived from `File.basename(resource_path)`

The name I think was mostly used for finding the subdir under `resources/`

jakolehm: Not sure if we can make it work.. path points to a sub-folder `resources`.

kke: Right. I guess it could then be just stack path and the resource_path would be that + resources/

lib/pharos/kube.rb
```diff
@@ -56,15 +58,15 @@ def self.config_exists?(host)
     # Shortcuts / compatibility:
```
kke: (These are not used much anymore, could be taken out)

SpComb: The `apply_stack` is still used in `ConfigureMetrics` 

FWIW the `apply_resource` is broken and never used, and the `prune_stack` is now unused, so those could be removed.

lib/pharos/kube.rb
```diff
@@ -56,15 +58,15 @@ def self.config_exists?(host)
...
 
     def self.apply_stack(host, name, vars = {})
-      session(host).stack(name, vars).apply
```
SpComb: Needs to be`File.join(RESOURCE_PATH, name)`?

lib/pharos/kube.rb
```diff
@@ -56,15 +58,15 @@ def self.config_exists?(host)
...
+      session(host).stack(name, RESOURCE_PATH, vars).apply
```
SpComb: Needs to be`File.join(RESOURCE_PATH, name)`?

jakolehm: Yes.

lib/pharos/kube.rb
jakolehm: Removed `apply_resource` & `prune_stack` (they are not used anymore).

lib/pharos/kube.rb
jakolehm: Fixed.

PR #359, opened 2018-05-09
Fix kubelet config defaulting

jnummelin:


lib/pharos/config_schema.rb
```diff
@@ -10,7 +10,8 @@ class ConfigSchema
       'api' => {},
       'network' => {},
       'authentication' => {},
-      'kube_proxy' => {}
+      'kube_proxy' => {},
+      'kubelet' => {}
```
jakolehm: Specs would be nice.

jnummelin: added

Issue #241, opened 2018-04-16
Pharos::Addon monkeypatches class name


kke:
`self.class.name` returns the plain addon name

Not necessarily a bug but perhaps slightly fishy.

PR #360, opened 2018-05-09
Allow to define external addons

jakolehm:


jakolehm:
Extras folder location is still hardcoded (relative to `cluster.yml`). Any ideas how it could be configured?

kke:
Maybe add some `ENV['PHAROS_EXTRAS'] || File.join(Dir.home, '.pharos', 'extras')` in there?

jakolehm:
@jnummelin @SpComb WDYT

lib/pharos/cluster_manager.rb
```diff
@@ -36,7 +36,8 @@ def addon_manager
     # load phases/addons
     def load
       Pharos::PhaseManager.load_phases(__dir__ + '/phases/')
-      Pharos::AddonManager.load_addons(__dir__ + '/addons/')
+      addon_dirs = [__dir__ + '/addons/', File.join(Dir.pwd, 'extras')]
+      Pharos::AddonManager.load_addons(*addon_dirs)
```
SpComb: The `*addon_dirs` indirection seems unnecessary, just `load_addons(__dir__ + '/addons/', File.join(Dir.pwd, 'extras'))`?

lib/pharos/addon.rb
```diff
@@ -26,56 +26,70 @@ def self.messages
...
-      if name
-        @name = name
-      else
-        @name
+    class << self
+      attr_writer :addon_location
+
+      # @param subclass [Class]
+      def inherited(subclass)
+        subclass.addon_location = File.dirname(caller(1..1).first[/(.+\.rb):\d+:in/, 1])
```
SpComb: This is a pretty awful hack, but I suppose it's well-enough contained that it's okay so long as it works for now.

jakolehm: Idk if there is any other way to achieve this.

SpComb: If the addons were files using some DSL rather than using the standard `require`, then it would be part of the `AddonManager.load_addons(...)`... but that's a bit further out.

jakolehm: DSL is now there.

jakolehm:
> do we want to support user-defined addons yet? 

Yes.

lib/pharos/addon.rb
SpComb: #361 isn't exactly what I meant by a DSL, it's still using `require`... but the `block.source_location.first` is slightly neater than the `caller(1..1).first`.

jakolehm:
Documentation PR: https://github.com/kontena/pharos-docs/pull/32

lib/pharos/addons/kubernetes-dashboard/addon.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+Pharos.addon 'kubernetes-dashboard' do
+  name 'kubernetes-dashboard'
+  version '1.8.3'
+  license 'Apache License 2.0'
+
+  install {
```
kke: Maybe use the `do` in these (and the schemas/structs) instead of mixing styles.

lib/pharos/addon_manager.rb
```diff
@@ -10,11 +10,19 @@ class AddonManager
     class InvalidConfig < Pharos::Error; end
     class UnknownAddon < Pharos::Error; end
 
+    # @return [Array<Pharos::Addon>]
```
SpComb: `@return [Array<Class<Pharos::Addon>>]`

lib/pharos/addon_manager.rb
```diff
@@ -38,7 +46,7 @@ def prev_configs
 
     # @return [Array<Pharos::Addon>]
```
SpComb: `@return [Array<Class<Pharos::Addon>>]`

lib/pharos/addon_manager.rb
```diff
@@ -10,11 +10,19 @@ class AddonManager
     class InvalidConfig < Pharos::Error; end
     class UnknownAddon < Pharos::Error; end
 
+    # @return [Array<Pharos::Addon>]
+    def self.addons
+      @addons ||= []
+    end
+
     # @param dirs [Array<String>]
+    # @return [Array<Pharos::Addon>]
```
SpComb: `@return [Array<Class<Pharos::Addon>>]`

lib/pharos/addons/cert-manager/addon.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+Pharos.addon 'cert-manager' do
+  version '0.2.3'
+  license 'Apache License 2.0'
+
+  issuer = custom_type {
```
SpComb: With newer version of `dry-struct` (?) this would just be a nested struct: http://dry-rb.org/gems/dry-struct/nested-structs/

```ruby
config do
  attribute :issuer do
    attribute :name, Pharos::Types::String
    ...
  end
end
```

lib/pharos/addons/ingress-nginx/addon.rb
```diff
@@ -0,0 +1,51 @@
...
+  config_schema {
+    optional(:configmap).filled(:hash?)
+    optional(:node_selector).filled(:hash?)
+    optional(:default_backend).schema {
+      optional(:image).filled(:str?)
+    }
+  }
+
+  install {
+    install(
```
SpComb: Overriding the name `install` for two different things here isn't good, this should probably be `install_stack`.

lib/pharos/addon.rb
```diff
@@ -5,6 +5,20 @@
 require_relative 'logging'
 
 module Pharos
+  # @param name [String]
+  # @return [Pharos::Addon]
+  def self.addon(name, &block)
+    klass = Class.new(Pharos::Addon, &block).tap do |addon|
+      addon.addon_location = block.source_location.first
+      addon.name(name)
```
SpComb: This overrides whatever `name ...` is within the addon itself, making it meaningless... risks having the wrong `name ...` within the addon itself, making it misleading.

lib/pharos/addons/kubernetes-dashboard/addon.rb
```diff
@@ -0,0 +1,19 @@
...
+    install
+
+    logger.info { "~~> kubernetes-dashboard can be accessed via kubectl proxy at http://localhost:8001/ui" }
+
+    service_account = kube_client.get_service_account('dashboard-admin', 'kube-system')
+    token_secret = service_account.secrets[0]
+    return unless token_secret
```
kke: Hmm.  I think this (or any other `return` for that matter) will give you a `LocalJumpError`.

```ruby
class Addon
  def self.hooks
    @hooks ||= {}
  end

  def self.install(&block)
    hooks[:install] = block
  end

  def run_install
    puts "Installing.."
    self.class.hooks[:install].call
    puts "Installed!"
  end
end

class FooAddon < Addon
  install do
    return "hello"
  end
end

FooAddon.new.run_install
```

==> 

```console
Installing..
Traceback (most recent call last):
	2: from foo.rb:23:in `<main>'
	1: from foo.rb:12:in `run_install'
foo.rb:19:in `block in <class:FooAddon>': unexpected return (LocalJumpError)
```

lib/pharos/addon.rb
```diff
@@ -26,62 +40,98 @@ def self.messages
...
+      def validate(config)
+        if @schema
+          @schema.call(config)
+        else
+          validation {}.call(config)
+        end
+      end
+
+      def descendants
+        ObjectSpace.each_object(Class).select { |klass| klass < self }
```
SpComb: No longer used for the `Pharos::AddonManager`, but still used by the `version` command... this should be removed and have both use the same code paths.

lib/pharos/addons/ingress-nginx/addon.rb
jnummelin: totally agree, having two `install`s makes it really hard to grok out

jakolehm: Ok, kinda agree that it's confusing.. but so is `apply_stack`. What is a stack (this is an addon) and why should user know about it?

lib/pharos/addon_manager.rb
```diff
@@ -10,11 +10,19 @@ class AddonManager
...
     def self.load_addons(*dirs)
       dirs.each do |dir|
-        Dir.glob(File.join(dir, '**', '*.rb')).each { |f| require(f) }
+        Dir.glob(File.join(dir, '**', 'addon.rb')).each { |f| require(f) }
```
kke: I dislike this `addon.rb` thing a little because now when editing multiple files, all the tabs will just say "addon.rb". Also makes any fancier file fuzzyfinding thing annoying, you can only complete up to the dirname.

jakolehm: Fixed name makes it more consistent? If filename can be anything it's a bit misleading signal for users.. can they have multiple `.rb` files within a same folder etc...

lib/pharos/addons/ingress-nginx/addon.rb
SpComb: If we assume that addon is 1:1 with a stack, then I think the high-level API addon API shold look like:

```ruby
let(:configmap) { config.configmap || {} }
let(:node_selector) { config.node_selector || {} }
let(:image_name) { config.default_backend.image || IMAGES[cpu_arch.name] }`
...
```

SpComb: My suggestion is to just keep the `def install ...` for this PR, and then have a separate PR for making the DSL less `def`-ish.

lib/pharos/addon_manager.rb
kke: Yep, probably simpler.

lib/pharos/addon.rb
```diff
@@ -26,62 +40,98 @@ def self.messages
...
+    attr_reader :config, :cpu_arch, :cluster_config, :master
 
+    # @param config [Hash]
+    # @param enabled [Boolean]
+    # @param master [Pharos::Configuration::Host,NilClass]
+    # @param cpu_arch [String, NilClass]
+    # @param cluster_config [Pharos::Config, NilClass]
     def initialize(config = nil, enabled: true, master:, cpu_arch:, cluster_config:)
-      @config = self.class.struct.new(config)
+      @config = self.class.config? ? self.class.config.new(config) : RecursiveOpenStruct.new(config)
```
kke: If we're directly using some dependency of a dependency like `RecursiveOpenStruct`, maybe it should be in our gemspec.

lib/pharos/addons/kubernetes-dashboard/addon.rb
jakolehm: Does it throw `LocalJumpError` also if `instance_eval` is used (it's using instance_eval currently)?

lib/pharos/addon.rb
jakolehm: So what is your suggestion? Try to make `.name` private or?

lib/pharos/addons/kubernetes-dashboard/addon.rb
kke: The problem is that a return from a proc returns from the thing that is calling the proc, not from the proc itself. Just like:

```ruby
def run
  foo.each do |xxx|
    return "foo" 
  end
end
```

does not returrn from the `.each` block but the whole `run`.

jakolehm: Makes sense, need to fix kubernetes-dashboard addon.

lib/pharos/addon.rb
SpComb: I don't know, but repeating the name in both the `Pharos.addon '...'` and within `name '...'` doesn't make sense. Maybe keep the `Pharos.addon '...'` name and remove the set-form of `def name(name)`... maybe also remove the entire `def name` to fix #241 at the same time.

lib/pharos/addons/ingress-nginx/addon.rb
jakolehm: Is there something else "wrong" with the current `install` block except that we call also method with the same name? If no, then I would keep it here and just try to come up with better name.

Options:
- `apply_stack` (bad, as I said we don't have concept of stacks explained)
- `install` (bad, it has the same name as the block)
- `apply_resources` (better? we have thing called resources folder)
- `install_resources` (better? we have thing called resources folder)

jnummelin: my vote would go to `apply_resources`, that maps it to kube API usage well too. we essentially `apply` bunch of `resources`

jakolehm:
```ruby
  install {
    install_resources(
      image: image_name,
      default_backend_replicas: default_backend_replicas
    )
  }
```

kke: It's a bit unclear what happens and when. 

If you have an addon without `install` I suppose it just does the resource applying.

We could have something like:


```ruby
hooks[:install] = proc do
  hooks.values_at(:before, :apply, :after).compact.map(&:call)
end
```

So normally you would then do:
```
before do
 puts "installing foofoo"
end

after do
  puts "you can use foofoo by foofooing"
end
```

If you don't have any resources or somehow want to mess with the applying you use:

```ruby
install do
  ssh { puts file('/etc/motd`).read }
end
```

lib/pharos/addons/cert-manager/addon.rb
jakolehm: Yes, it is much better.

lib/pharos/addon.rb
jakolehm: Fixed.

lib/pharos/addons/kubernetes-dashboard/addon.rb
jakolehm: Fixed.

lib/pharos/addons/ingress-nginx/addon.rb
jakolehm: > If you have an addon without install I suppose it just does the resource applying.

Yes. We need to document this.

I think we can add before/after_install and before/after_uninstall later.

lib/pharos/version_command.rb
```diff
@@ -21,15 +21,15 @@ def phases
...
+      Pharos::AddonManager.addons.sort_by(&:name)
     end
 
     def load_phases
       Pharos::PhaseManager.load_phases(__dir__ + '/phases')
     end
 
     def load_addons
-      Pharos::AddonManager.load_addons([__dir__ + '/addons'])
+      Pharos::AddonManager.load_addons([File.join(__dir__, '..', '..', 'addons'),])
```
SpComb: Still failing on rubocop

```
lib/pharos/version_command.rb:32:81: C: Style/TrailingCommaInArrayLiteral: Avoid comma after the last item of an array. (https://github.com/bbatsov/ruby-style-guide#no-trailing-array-commas)
      Pharos::AddonManager.load_addons([File.join(__dir__, '..', '..', 'addons'),])
^
```

How does the `load_addons([...])` even work when it's defined as `def self.load_addons(*paths)`? Shouldn't it be `Pharos::AddonManager.load_addons(File.join(...))`, or does ruby somehow implicitly flatten the array?

PR #362, opened 2018-05-11
Host operating system configurers

jakolehm:
Refactors all operating system specific "things" to `Configurer` class. First implementation is `UbuntuXenial` which is currently the only supported linux distro.


lib/pharos/host/configurer.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module Pharos
+  module Host
+    class Configurer
```
kke: `Configurator` would sound better to me

SpComb:
At this point this kinda makes sense as a way to specialize the `Pharos::Phases::ConfigureHost` per OS, but a lot of the other phases also invoke scripts that use `apt-get` (`configure-etcd.sh`, `configure-kube.sh`, `configure-kubelet-proxy.sh`, `install-kubeadm.sh`)... moving all of those scripts used by different phases into a single class feels like it would defeat the point of having separate phase classes in this first place, and having this kind of "phase class is a thin wrapper around a composed OS-specific class" pattern for all of those phases will quickly get unwieldy...

Looks like at this point this is basically equivalent to having the phases use `exec_script("#{@host.os}/configure-foo.sh")`? The interesting part would probably be the component selections/versions varying by host OS? Could we somehow handle this directly in the `Phase` base class instead, by making `Pharos::Phase#exec_script` use OS-specific scripts, and having optional OS-specific overrides for component versions?

```ruby
      register_component(
        name: 'docker', license: 'Apache License 2.0',
        enabled: proc { |c| c.hosts.any? { |h| h.container_runtime == 'docker' } }
      ) do
        override os: 'ubuntu/xenial' do
          { version: '1.13.1' }
        end
        override os: 'ubuntu/bionic' do
          { version: '17.12.1' }
        end
      end
```

Perhaps using something like #269 to lookup the per-host component version (`VERSION: components['docker'].version`)?

SpComb:
So :-1: on a design level in favor of allowing the `Phase` classes themselves to be specialized by OS somehow, because this essentially leaves the phase classes as nothing more than thin wrappers for instantiating a `Configurer` class variant and invoking some methods on it...

jakolehm:
> So 👎 on a design level in favor of allowing the Phase classes themselves to be specialized by OS somehow, because this essentially leaves the phase classes as nothing more than thin wrappers for instantiating a Configurer class variant and invoking some methods on it... the Pharos::Host::Configurer class is confusingly similar to Pharos::Phase.

Refactored more, now all OS specific bits are moved to `Pharos::Host::Configurer` ... IMO there is still plenty of functionality for phases because they handle the generic flow (manifests etc..).

Biggest reason for this design is to make adding new OS as easy as possible. If OS specific pieces are separate from the actual control flow then OS (at least in theory) could be pluggable.

jakolehm:
@kke @jnummelin PTAL

lib/pharos/scripts/wait-etcd.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -e
+
+etcd_healthy() {
+  response=$(curl -s --noproxy "*" --cacert /etc/pharos/pki/ca.pem --cert /etc/pharos/pki/etcd/client.pem --key /etc/pharos/pki/etcd/client-key.pem https://${PEER_IP}:2379/health)
+  [ "${response}" = '{"health": "true"}' ]
+}
+
+echo "Waiting etcd to launch on port 2380..."
```
jnummelin: The check is made for `2379` port... :)

PR #363, opened 2018-05-15
Better config specs

SpComb:
Move all config specs into one place, and fix the specs to cover the complete hash => schema => struct codepath, also taking into account the `DEFAULT_DATA`.


SpComb:
The CLI error handling should still behave exactly the same with the new `Pharos::ConfigError`:

```
$ docker-compose run --rm -e DEBUG=true pharos-cluster up || echo $?
==> KONTENA PHAROS v1.0.1 (Kubernetes v1.10.2)
==> Reading instructions ...
==> Invalid configuration:
---
:network:
  :provider:
  - 'must be one of: weave calico'
11
```

SpComb:
Rubocop is complaining about existing code:

```
SplatKeywordArguments: Do not use splat keyword arguments as a single Hash.
        @client.delete_resource(@resource, **options)
^^^^^^^^^ 
```

lib/pharos/config.rb
```diff
@@ -18,6 +18,21 @@ class Config < Dry::Struct
 
     constructor_type :schema
 
+    # @param data [Hash]
```
jnummelin: <nitpicking>
param name should be `raw_data` :)
</nitpicking>

jnummelin:
maybe merge/rebase with  master to make rubocop happy for this PR

PR #374, opened 2018-05-18
Addons DSL

SpComb:
Moves the built-in addons out of `lib/pharos/addons` and `spec/pharos/addons` into their own top-level `addons` dir in the git repo / gem. Each addon dir is loaded as a `Pharos::Addon` subclass, with the contents of each `*.rb` file being evaluated in the class body as a kind of DSL.

The addon dirs can also contain their own specs, using `Pharos::Addon.load` to load the addon class.

Future work:
* Improve the addon DSL to get rid of all `def` overrides, using something like `validate do raise unless ...`, `stack_vars do { default_backend_replicas: ... } end` (or `let(:default_backend_replicas) do ...`?) instead.
* Load user-defined addons from alongside the `cluster.yml`


lib/pharos/addon.rb
```diff
@@ -1,13 +1,52 @@
...
 module Pharos
   class Addon
     include Pharos::Logging
 
+    # Load addon classes from filesystem path
+    #
+    # @param path [String] path to dir containing */*.rb addon dirs
+    # @return [Array<Class<Kontena::Pharos::Addon>>]
+    def self.loads(path)
+      paths = Dir.glob("#{path}/*")
```
kke: No file type restriction

lib/pharos/addon.rb
```diff
@@ -55,27 +91,24 @@ def self.to_h
     end
 
     def self.schema(&block)
-      @schema = Dry::Validation.Form(Schema, &block)
+      if block
```
kke: `if block_given?`

lib/pharos/addon.rb
```diff
@@ -55,27 +91,24 @@ def self.to_h
...
-    end
-
-    def self.validation
-      Dry::Validation.Form(Schema) { yield }
-    end
-
-    def self.validate(config)
-      if @schema
-        @schema.call(config)
+      if block
```
kke: `if block_given?`

lib/pharos/addon.rb
```diff
@@ -1,13 +1,52 @@
...
+require 'dry-struct'
 require 'dry-validation'
-require_relative 'addons/struct'
 require_relative 'logging'
 
 module Pharos
   class Addon
     include Pharos::Logging
 
+    # Load addon classes from filesystem path
```
kke: Yardoc mismatch

lib/pharos/addon.rb
```diff
@@ -1,13 +1,52 @@
...
+    #
+    # @param path [String] path to dir containing */*.rb addon dirs
+    # @return [Array<Class<Kontena::Pharos::Addon>>]
+    def self.loads(path)
```
kke: Undescriptive method name

jakolehm:
This probably can be closed?

PR #376, opened 2018-05-21
Always disable apt unattended-upgrade

### LINKED ISSUE #375: Unattended-upgrade is not always disabled
Depending on used image/configuration unattended-upgrade might be enabled. We should disable it so that `host-upgrades` addon is the only way to apply host level upgrades.

jakolehm:
Fixes #375

### DIFF
--- a/lib/pharos/scripts/configure-essentials.sh
+++ b/lib/pharos/scripts/configure-essentials.sh
@@ -7,3 +7,8 @@ if ! dpkg -l apt-transport-https software-properties-common > /dev/null; then
     apt-get update -y
     apt-get install -y apt-transport-https software-properties-common
 fi
+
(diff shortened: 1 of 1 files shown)


lib/pharos/scripts/configure-essentials.sh
```diff
@@ -7,3 +7,8 @@ if ! dpkg -l apt-transport-https software-properties-common > /dev/null; then
     apt-get update -y
     apt-get install -y apt-transport-https software-properties-common
 fi
+
+cat <<EOF >/etc/apt/apt.conf.d/20auto-upgrades
+APT::Periodic::Update-Package-Lists "1";
```
SpComb: Note that this is still required for `unattended-upgrade` to work, I don't think it will do the equivalent of `apt-get update` by itself.

PR #382, opened 2018-05-26
Allow to pass addon config from terraform

jakolehm:
For example:
```tf
output "pharos_addons" {
  value = {
    efs-provisioner = {
      efs_id = "${aws_efs_file_system.default.id}"
      dns_name = "${aws_efs_file_system.default.dns_name}"
      region = "${var.aws_region}"
    }
  }
}
```
will result in:
```yaml
addons:
  ingress-nginx:
    efs_id: "..."
    dns_name: "..."
    region: "..."
```

### DIFF
--- a/lib/pharos/addon.rb
+++ b/lib/pharos/addon.rb
@@ -114,13 +114,13 @@ module Pharos
 
     attr_reader :config, :cpu_arch, :cluster_config, :master
 
-    # @param config [Hash]
+    # @param config [Hash,Dry::Validation::Result]
     # @param enabled [Boolean]
     # @param master [Pharos::Configuration::Host,NilClass]
     # @param cpu_arch [String, NilClass]
     # @param cluster_config [Pharos::Config, NilClass]
     def initialize(config = nil, enabled: true, master:, cpu_arch:, cluster_config:)
-      @config = self.class.config? ? self.class.config.new(config) : RecursiveOpenStruct.new(config)
+      @config = self.class.config? ? self.class.config.new(config) : RecursiveOpenStruct.new(config.to_hash)
       @enabled = enabled
       @master = master
       @cpu_arch = cpu_arch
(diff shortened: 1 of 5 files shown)


lib/pharos/up_command.rb
```diff
@@ -72,8 +72,10 @@ def load_terraform(file, config)
       tf_parser = Pharos::Terraform::JsonParser.new(File.read(file))
       config['hosts'] ||= []
       config['api'] ||= {}
+      config['addons'] ||= {}
       config['hosts'] += tf_parser.hosts
       config['api'].merge!(tf_parser.api) if tf_parser.api
+      config['addons'].merge!(tf_parser.addons) if tf_parser.addons
```
jnummelin: Isn't the `if` completely unnecessary as `tf_parse.addons` will always return a hash. Yes, it'll be an empty hash if no addons in the output but empty hash is not falsey for the if

jakolehm: > Isn't the if completely unnecessary

Yes.

jakolehm:
@SpComb PTAL

lib/pharos/addon.rb
```diff
@@ -114,13 +114,13 @@ def validate(config)
...
 
-    # @param config [Hash]
+    # @param config [Hash,Dry::Validation::Result]
     # @param enabled [Boolean]
     # @param master [Pharos::Configuration::Host,NilClass]
     # @param cpu_arch [String, NilClass]
     # @param cluster_config [Pharos::Config, NilClass]
     def initialize(config = nil, enabled: true, master:, cpu_arch:, cluster_config:)
-      @config = self.class.config? ? self.class.config.new(config) : RecursiveOpenStruct.new(config)
+      @config = self.class.config? ? self.class.config.new(config) : RecursiveOpenStruct.new(config.to_hash)
```
kke: Potential for `NoMethodError (undefined method `to_hash' for nil:NilClass)` because config defaults to nil.

Can be avoided by `RecursiveOpenStruct.new(Hash(config))` (returns an empty hash or tries .to_hash)

PR #383, opened 2018-05-26
Update pharos-kubelet-proxy to 0.3.6

jakolehm:


jakolehm:
@jnummelin PTAL

jakolehm:
This is also fixed in #385

PR #385, opened 2018-05-28
Configurable image_repository

jakolehm:
By default pulls all the "core" images from `quay.io/kontena/`.

Todo:

- [x] kubernetes control plane
- [x] kubedns
- [x] etcd
- [x] heapster
- [x] pause
- [x] pharos-kubelet-proxy
- [x] weave
- [x] calico


jakolehm:
@SpComb @jnummelin PTAL

lib/pharos_cluster.rb
```diff
@@ -13,5 +13,5 @@ module Pharos
   KUBEADM_VERSION = ENV.fetch('KUBEADM_VERSION') { KUBE_VERSION }
   ETCD_VERSION = ENV.fetch('ETCD_VERSION') { '3.1.12' }
   DOCKER_VERSION = '1.13.1'
-  KUBELET_PROXY_VERSION = '0.3.5'
+  KUBELET_PROXY_VERSION = '0.3.6'
```
SpComb: Merge from #383?

jakolehm: Yes, it did not make sense to keep these separate (that's why I closed #383)

SpComb: Ah, version bump is for the image push to `quay.io`?

jakolehm: Yes.

jnummelin:
Should we handle all the built-in addons from Quay.io too?

jakolehm:
> Should we handle all the built-in addons from Quay.io too? 

Yes, but it should be a separate PR (needs more image plumbing)

PR #387, opened 2018-05-28
Fix openebs to use qualified images

### LINKED ISSUE #386: OpenEBS addon fails on crio
All the OpenEBS configs use "un-qualified" image names like `openebs/m-apiserver:<%= version %>`. This fails in image pull with cri-o as it will NOT default to use docker.io as the registry. Need to ensure all the images are specified with fully qualified names.

jnummelin:
fixes #386


jnummelin:
Tested to work with simple Redis setup that actually uses OpenEBS volumes

PR #389, opened 2018-05-29
Configurable http proxy per host

### LINKED ISSUE #388: Support http proxies

jakolehm:
Fixes #388

### DIFF
--- a/examples/vagrant/Vagrantfile
+++ b/examples/vagrant/Vagrantfile
@@ -32,7 +32,7 @@ Vagrant.configure("2") do |config|
         vb.cpus = 1
       end
       host.vm.network "private_network", ip: "192.168.100.#{i + 100}"
-      # host.vm.provision "shell", path: 'proxy-only.sh'
+      host.vm.provision("shell", path: 'proxy-only.sh') if ENV['VAGRANT_HTTP_PROXY']
     end
   end
 end
--- a/examples/vagrant/Vagrantfile
+++ b/examples/vagrant/Vagrantfile
@@ -32,6 +32,7 @@ Vagrant.configure("2") do |config|
         vb.cpus = 1
       end
       host.vm.network "private_network", ip: "192.168.100.#{i + 100}"
+      #host.vm.provision "shell", path: 'proxy-only.sh'
     end
   end
 end
--- a/examples/vagrant/Vagrantfile
+++ b/examples/vagrant/Vagrantfile
@@ -32,7 +32,7 @@ Vagrant.configure("2") do |config|
         vb.cpus = 1
       end
       host.vm.network "private_network", ip: "192.168.100.#{i + 100}"
-      #host.vm.provision "shell", path: 'proxy-only.sh'
+      # host.vm.provision "shell", path: 'proxy-only.sh'
     end
   end
 end
--- a/lib/pharos/phase.rb
+++ b/lib/pharos/phase.rb
@@ -53,6 +53,7 @@ module Pharos
     end
 
     # @param script [String] name of file under ../scripts/
+    # @param vars [Hash]
     def exec_script(script, vars = {})
       @ssh.exec_script!(
         script,
@@ -61,6 +62,9 @@ module Pharos
       )
     end
 
+    # @param path [String]
+    # @param vars [Hash]
+    # @return [Pharos::YamlFile]
     def parse_resource_file(path, vars = {})
       Pharos::YamlFile.new(resource_path(path)).read(vars)
     end
--- a/lib/pharos/phases/configure_host.rb
+++ b/lib/pharos/phases/configure_host.rb
@@ -22,6 +22,7 @@ module Pharos
 
       def call
         logger.info { "Configuring essential packages ..." }
+        @ssh.file('/usr/local/lib/pharos.sh').write(File.read(File.join(__dir__, '..', 'scripts', 'pharos.sh')))
         exec_script(
           'configure-essentials.sh',
           HTTP_PROXY: @host.http_proxy.to_s,
--- a/lib/pharos/phases/configure_host.rb
+++ b/lib/pharos/phases/configure_host.rb
@@ -24,7 +24,8 @@ module Pharos
         logger.info { "Configuring essential packages ..." }
         exec_script(
           'configure-essentials.sh',
-          HTTP_PROXY: @host.http_proxy
+          HTTP_PROXY: @host.http_proxy.to_s,
+          SET_HTTP_PROXY: @host.http_proxy.nil? ? 'false' : 'true'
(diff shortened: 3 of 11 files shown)


examples/vagrant/Vagrantfile
```diff
@@ -32,6 +32,7 @@ Vagrant.configure("2") do |config|
         vb.cpus = 1
       end
       host.vm.network "private_network", ip: "192.168.100.#{i + 100}"
+      # host.vm.provision "shell", path: 'proxy-only.sh'
```
kke: I suppose this is intentionally here, but maybe it could be something like `if ENV['VAGRANT_HTTP_PROXY']`?

lib/pharos/phase.rb
```diff
@@ -53,6 +53,7 @@ def resource_path(*path)
     end
 
     # @param script [String] name of file under ../scripts/
+    # @param vars [Hash]
```
kke: or `env`

lib/pharos/phases/configure_host.rb
```diff
@@ -22,7 +22,11 @@ class ConfigureHost < Pharos::Phase
 
       def call
         logger.info { "Configuring essential packages ..." }
-        exec_script('configure-essentials.sh')
+        exec_script(
+          'configure-essentials.sh',
+          HTTP_PROXY: @host.http_proxy.to_s,
+          SET_HTTP_PROXY: @host.http_proxy.nil? ? 'false' : 'true'
```
kke: This could be handled in the script instead

jakolehm: How?

lib/pharos/phases/configure_host.rb
```diff
@@ -35,7 +39,8 @@ def call
           exec_script(
             'configure-docker.sh',
             DOCKER_PACKAGE: 'docker.io',
-            DOCKER_VERSION: "#{Pharos::DOCKER_VERSION}-0ubuntu1~16.04.2"
+            DOCKER_VERSION: "#{Pharos::DOCKER_VERSION}-0ubuntu1~16.04.2",
+            HTTP_PROXY: @host.http_proxy.to_s
```
kke: These seem to be everywhere. `Host` could have something like `default_env` that would be merged in `exec_script`

jakolehm: I'm not sure if we actually need these anymore. `HTTP_PROXY` should be already set via `/etc/environment`.

lib/pharos/scripts/configure-etcd.sh
```diff
@@ -3,7 +3,7 @@
 set -e
 
 etcd_healthy() {
-  response=$(curl -s --cacert /etc/pharos/pki/ca.pem --cert /etc/pharos/pki/etcd/client.pem --key /etc/pharos/pki/etcd/client-key.pem https://${PEER_IP}:2379/health)
+  response=$(curl -s --noproxy "*" --cacert /etc/pharos/pki/ca.pem --cert /etc/pharos/pki/etcd/client.pem --key /etc/pharos/pki/etcd/client-key.pem https://${PEER_IP}:2379/health)
```
kke: This could be something like `--no-proxy "${PEER_IP}"`. Theoretically I suppose the `https://${PEER_IP}:2379/health` could redirect somewhere. Maybe far fetched.

lib/pharos/scripts/configure-cri-o.sh
```diff
@@ -2,13 +2,31 @@
 
 set -ue
 
+reload_daemon() {
+    systemctl daemon-reload
+    systemctl is-active --quiet crio && systemctl restart crio
```
SpComb: `systemctl try-restart crio`

lib/pharos/scripts/configure-docker.sh
```diff
@@ -17,6 +17,25 @@ cat <<EOF >/etc/docker/daemon.json
 }
 EOF
 
+reload_daemon() {
+    systemctl daemon-reload
+    systemctl is-active --quiet docker && systemctl restart docker
```
SpComb: `systemctl try-restart docker`

lib/pharos/scripts/configure-docker.sh
```diff
@@ -17,6 +17,25 @@ cat <<EOF >/etc/docker/daemon.json
...
+}
+
+if [ -n "$HTTP_PROXY" ]; then
+    mkdir -p /etc/systemd/system/docker.service.d
+    cat <<EOF >/etc/systemd/system/docker.service.d/http-proxy.conf
+[Service]
+Environment="HTTP_PROXY=${HTTP_PROXY}"
+EOF
+    reload_daemon
```
SpComb: We're now restarting the docker daemon for every `pharos-cluster up` run. but only if the http proxy is configured...

Does that cause problems? It's something we've been avoiding until now.

Ideally we should only be reloading if the systemd unit dropin has changed...

lib/pharos/scripts/configure-essentials.sh
```diff
@@ -2,6 +2,19 @@
 
 set -e
 
+if [ "${SET_HTTP_PROXY}" = "true" ]; then
+    cat <<EOF >/etc/environment
+PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games"
```
SpComb: It would be nice to avoid hardcoding the `PATH`, but that would need some kind of `lineinfile` helper using sed instead...

lib/pharos/scripts/configure-docker.sh
jakolehm: It should not cause too many problems because it leaves containers running.

SpComb: Still feels risky to restart docker/crio, particularly if it only happens in relatively niche configurations (with http_proxy). It could be avoided if we were smarter about it... like ansible `notify` or Puppet `File[...] ˝> Service[...]`

lib/pharos/scripts/configure-essentials.sh
jakolehm: Fixed.

jakolehm:
I think I fixed most complaints (except restarting of docker/cri-o).

lib/pharos/scripts/configure-essentials.sh
```diff
@@ -2,13 +2,26 @@
 
 set -e
 
+. /usr/local/lib/pharos.sh
```
kke: My `/usr/local/lib` just has a bunch of `.so` and `.a` files and subdirs, maybe `/usr/local/lib/pharos/util.sh`?

PR #393, opened 2018-05-31
Update host-upgrades addon to use pharos-host-upgrades

SpComb:
See https://github.com/kontena/pharos-host-upgrades

Currently using the `quay.io/kontena/pharos-host-upgrades:edge` build, pending testing and a tagged release.

The defaults run upgrades at midnight, <s>waiting up to two hours for the kube lock</s>, without rebooting the host.

Replaces the `kured` addon, which needs to be deprecated somehow... This functionality is optional, it needs `reboot: true` to enable it.

# TODO
- [x] Test
- [x] Don't run upgrades at exactly midnight by default
- [x] Use tagged `pharos-host-upgrades` release images
- [ ] #420 Deprecate the `kured` addon
- [x] Add arm64 support
  - [x] https://github.com/kontena/pharos-host-upgrades/issues/24


addons/host-upgrades/resources/40-configmap.yml
```diff
@@ -0,0 +1,159 @@
...
+  yum-cron.conf: |
+    [commands]
+    #  What kind of update to use:
+    # default                            = yum upgrade
+    # security                           = yum --security upgrade
+    # security-severity:Critical         = yum --sec-severity=Critical upgrade
+    # minimal                            = yum --bugfix update-minimal
+    # minimal-security                   = yum --security update-minimal
+    # minimal-security-severity:Critical =  --sec-severity=Critical update-minimal
+    update_cmd = default
```
SpComb: I believe this will install all available updated package from all yum repos... my understanding is that CentOS is missing the required repo metadata for things like `minimal-security` to work, those are only available in RHEL:

* https://www.reddit.com/r/CentOS/comments/7wrjcp/how_to_handle_security_upgrades/
* https://updateinfo.cefs.steve-meier.de/
* https://www.caseylabs.com/centos-automatic-security-updates-do-not-work/

addons/host-upgrades/resources/40-configmap.yml
```diff
@@ -0,0 +1,159 @@
...
+    # Uncomment to auto-import new gpg keys (dangerous)
+    # assumeyes = True
+
+  unattended-upgrades.conf: |
+    // Override system /etc/apt/apt.conf.d/50unattended-upgrades
+    #clear "Unattended-Upgrade::Allowed-Origins";
+
+    // Automatically upgrade packages from these (origin:archive) pairs
+    Unattended-Upgrade::Allowed-Origins {
+    // "${distro_id}:${distro_codename}";
```
SpComb: The `Ubuntu:xenial` origin is whitelisted in the default configuration, AFAIK that corresponds to LTS point releases (16.04.x):

```
// Automatically upgrade packages from these (origin:archive) pairs
Unattended-Upgrade::Allowed-Origins {
	"${distro_id}:${distro_codename}";
	"${distro_id}:${distro_codename}-security";
	// Extended Security Maintenance; doesn't necessarily exist for
	// every release and this system may not have it installed, but if
	// available, the policy for updates is such that unattended-upgrades
	// should also install from here by default.
	"${distro_id}ESM:${distro_codename}";
//	"${distro_id}:${distro_codename}-updates";
//	"${distro_id}:${distro_codename}-proposed";
//	"${distro_id}:${distro_codename}-backports";
};
```

addons/host-upgrades/addon.rb
```diff
@@ -1,20 +1,25 @@
 # frozen_string_literal: true
 
 Pharos.addon 'host-upgrades' do
-  version '0.1.0'
+  version '0.0.0'
```
jakolehm: Just a reminder that this need to be updated.

addons/host-upgrades/addon.rb
```diff
@@ -1,20 +1,29 @@
...
   license 'Apache License 2.0'
 
   config {
-    attribute :interval, Pharos::Types::String
+    attribute :image_tag, Pharos::Types::String.default("edge")
+    attribute :schedule, Pharos::Types::String
```
SpComb: This absolutely needs to have some regexp for validation, because anyone configuring this is going to be more likely to get the syntax wrong rather than right, and then be left debugging their `CrashLoopBackoff` pods... the feedback cycle for fixing this configuration is too long and painful otherwise.

jakolehm: I would prefer to merge this sooner than later.. if regexp takes time then we should open a new issue and mark it to the 1.2.0 milestone.

SpComb: No need, I already have something, it just needs specs to test it.

addons/host-upgrades/addon.rb
```diff
@@ -1,20 +1,33 @@
...
+    attribute :schedule_window, Pharos::Types::String
```
SpComb: Note: The `--schedule-window` defaults to `1h` within `pharos-host-upgrades` itself, which might be a mistake... it should still be possible to disable the `--schedule-window` by using `schedule_window: 0` - or something that makes sense given the `schedule`, like `schedule_window: 24h`.

SpComb:
Leaving #420 as a separate issue for the `kured` deprecation... having both `kured` and `host-upgrades' running with `reboot: true` may or may not behave badly.

addons/host-upgrades/addon.rb
```diff
@@ -5,16 +5,48 @@
...
   }
 
   config_schema {
-    required(:interval).filled(:str?, :duration?)
+    configure do
+      SCHEDULE_RE = %r(^([*?]|[A-Z0-9/,-]+)(\s+([*?]|[A-Z0-9/,-]+)){5}|@\w+$)
+      DURATION_RE = /^(\d+(\.\d+)?(ns|us|ms|s|m|h|))+$/
+
+      def schedule?(value)
+        !!SCHEDULE_RE.match?(value)
```
jakolehm: Just wondering should we use valid cron schedule (we have fugit gem as a dep)?

addons/host-upgrades/resources/40-configmap.yml
jakolehm: Yes, CentOS is missing security metadata. We lock kube/docker packages with `yum-lockversion` which should make this a bit more safe. For RHEL we could use `security` or similar?

SpComb: Should probably still consider testing the combination of `yum-lockversion` and the `yum-cron` configuration to make sure it doesn't upgrade pinned packages.

addons/host-upgrades/addon.rb
SpComb: The `pharos-cluster` itself doesn't need to actually parse the cron string, but `Fugit::Cron.parse` would probably be better than a regexp... unfortunately, the github.com/robfig/cron package used by `pharos-host-upgrades` doesn't accept the standard five-field cron syntax, it expects an extended cron syntax with a leading seconds field... which is a bad idea IMO, it would be better if this used a standard cron syntax, we don't really need per-second granularity here.

Maybe we should change `pharos-host-upgrades` to accept a standard five-field cron syntax before releasing this?

addons/host-upgrades/addon.rb
```diff
@@ -5,16 +5,48 @@
   license 'Apache License 2.0'
 
   config {
-    attribute :interval, Pharos::Types::String
```
SpComb: If we wanted to make the 1.2 upgrade smoother, we could continue to accept existing addon configurations with `interval`, and just translate that to `schedule: "@every ..."`.

addons/host-upgrades/addon.rb
jakolehm: I meant that we could use `Fugit::Cron.parse` to parse standard five-field cron and then addon could convert it to something that `pharos-host-upgrades` understands.

addons/host-upgrades/addon.rb
jakolehm: Not sure if it's needed if the new addon complains when schedule is missing?

SpComb: Everyone using the existing `host-upgrades` addon will need to edit their `cluster.yml` when upgrading to 1.2. That could be avoided.

addons/host-upgrades/addon.rb
SpComb: I think that conversion should happen within `pharos-host-upgrades`.

addons/host-upgrades/addon.rb
jakolehm: I think it's ok to edit cluster.yml in this case.

addons/host-upgrades/addon.rb
SpComb: Looks like `Fugit::Cron.parse` also supports the alternative six-field cron syntax, which confuses things a bit now...

```
irb(main):003:0> Fugit::Cron.parse('0 0 0 * * *')
=> #<Fugit::Cron:0x0000000001c172c0 @original="0 0 0 * * *", @cron_s=nil, @seconds=[0], @minutes=[0], @hours=[0], @monthdays=nil, @months=nil, @weekdays=nil, @zone=nil, @timezone=nil>
```

spec/pharos/addons/host_upgrades_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+
+        expect(result).to be_success, result.errors.inspect
+      end
+    end
+
+    describe 'duration' do
+      it "fails with invalid duration" do
+        result = described_class.validate({enabled: true, schedule: '0 0 * * *', schedule_window: '1'})
+
+        expect(result).to_not be_success
```
SpComb: Having a hard time figuring out how to get `Fugit::Cron.parse` to reject or normalize the extended seconds syntax without monkey-patching it...

```

  1) Pharos::Addons::HostUpgrades#validate duration fails with invalid duration
     Failure/Error: expect(result).to_not be_success
       expected `#<Dry::Validation::Result output={:enabled=>true, :schedule=>"0 0 * * *", :schedule_window=>"1"} errors={}>.success?` to return false, got true
     # ./spec/pharos/addons/host_upgrades_spec.rb:75:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:121:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:111:in `block (2 levels) in <top (required)>'
```

spec/pharos/addons/host_upgrades_spec.rb
```diff
@@ -0,0 +1,120 @@
...
+        result = described_class.validate({enabled: true, schedule: '0 * * *'})
+
+        expect(result).to_not be_success
+        expect(result.errors.dig(:schedule)).to match [/is not a valid crontab/]
+      end
+
+      it "fails with long cron schedule" do
+        result = described_class.validate({enabled: true, schedule: '0 0 0 * * *'})
+
+        expect(result).to_not be_success
```
SpComb: Having a hard time figuring out how to get `Fugit::Cron.parse` to reject or normalize the extended seconds syntax without monkey-patching it...

```
  1) Pharos::Addons::HostUpgrades#validate schedule fails with long cron schedule
     Failure/Error: expect(result).to_not be_success
       expected `#<Dry::Validation::Result output={:enabled=>true, :schedule=>"0 0 0 * * *"} errors={}>.success?` to return false, got true
     # ./spec/pharos/addons/host_upgrades_spec.rb:49:in `block (4 levels) in <top (required)>'
     # ./spec/spec_helper.rb:121:in `block (2 levels) in <top (required)>'
     # ./spec/spec_helper.rb:111:in `block (2 levels) in <top (required)>'

```

SpComb: Hacked this in with a trivial monkeypatch to add a missing `attr_reader`... validating/normalizing the cron schedule was more difficult than expected this way.

addons/host-upgrades/addon.rb
SpComb: This will now get overridden with `--schedule-window=0` by default, which is IMO good behavior.

lib/pharos/addon.rb
```diff
@@ -1,6 +1,14 @@
 # frozen_string_literal: true
 
 require 'dry-validation'
+require 'fugit'
+
+module Fugit
+  class Cron
+    attr_reader :seconds # XXX: missing from upstream
```
jakolehm: We should try to push this also to the upstream.

SpComb: https://github.com/floraison/fugit/pull/3

jakolehm:
@SpComb ok to merge?

lib/pharos/addon.rb
SpComb: ^ this got merged and released upstream as 1.1.2, so I'll still update this PR to use that instead.

SpComb:
> @SpComb ok to merge?

Good to go now.

jakolehm:
@SpComb docs pr is still missing...

PR #403, opened 2018-06-01
Ensure that /etc/apt/apt.conf.d/20auto-upgrades exist

jakolehm:


lib/pharos/scripts/configure-essentials.sh
```diff
@@ -23,5 +23,8 @@ if ! dpkg -l apt-transport-https software-properties-common > /dev/null; then
 fi
 
 autoupgrade_file="/etc/apt/apt.conf.d/20auto-upgrades"
+if [ ! -f $autoupgrade_file ]; then
```
kke: less bracketful oneliner version: `test -f $autoupgrade_file || touch $autoupgrade_file`

PR #407, opened 2018-06-04
Initial CoreOS Container Linux / Flatcar Linux support

jakolehm:
Requires #362

### DIFF
--- a/lib/pharos/host/coreos/coreos.rb
+++ b/lib/pharos/host/coreos/coreos.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+
+module Pharos
+  module Host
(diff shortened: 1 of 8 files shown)


SpComb:
#362 is merged, so this can be rebased against master, and the PR base branch updated?

jakolehm:
Yes, but this is not in any milestone yet... not going to included in 1.2.x.

SpComb:
If the problem here is that some configurations (like the Docker version) are only configurable via cloud-config/ignition, what about a `pharos-cluster cloud-config` / `pharos-cluster ignition` approach, where we would generate an appropriate init config instead of configuring the host post-install via SSH?

PR #412, opened 2018-06-06
Upgrade net-ssh dependency to 5.0.1

kke:
Changes since 5.0.0beta2:

```
=== 5.0.1

  * default_keys were not loaded even if no keys or key_data options specified [#607]

=== 5.0.0

 * Verify_host_key options rename (true, false, :very, :secure depreacted new equivalents are :never, :accept_new_or_local_tunnel :accept_new :always) [Jared Beck, #595]
```

Doesn't look like breaking changes for pharos.


kke:
Rubocop fails until #411

PR #424, opened 2018-06-12
Handle migration from 1.1.x to 1.2.x

### LINKED ISSUE #422: Handle migration of changed systemd unit file names

jakolehm:
Fixes #422

### DIFF
--- a/lib/pharos/phases/configure_kubelet.rb
+++ b/lib/pharos/phases/configure_kubelet.rb
@@ -14,7 +14,7 @@ module Pharos
         enabled: proc { |c| !c.worker_hosts.empty? }
       )
 
-      DROPIN_PATH = "/etc/systemd/system/kubelet.service.d/05-pharos.conf"
+      DROPIN_PATH = "/etc/systemd/system/kubelet.service.d/11-pharos.conf"
       CLOUD_CONFIG_DIR = "/etc/pharos/kubelet"
       CLOUD_CONFIG_FILE = (CLOUD_CONFIG_DIR + '/cloud-config')
 
--- a/lib/pharos/phases/migrate_master.rb
+++ b/lib/pharos/phases/migrate_master.rb
@@ -6,19 +6,20 @@ module Pharos
       title "Migrate master"
 
       def call
-        migrate_0_5_to_0_6 if migrate_0_5_to_0_6?
+        if migrate_1_1_to_1_2?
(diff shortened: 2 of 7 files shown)


lib/pharos/phases/configure_kubelet.rb
```diff
@@ -14,7 +14,7 @@ class ConfigureKubelet < Pharos::Phase
         enabled: proc { |c| !c.worker_hosts.empty? }
       )
 
-      DROPIN_PATH = "/etc/systemd/system/kubelet.service.d/05-pharos.conf"
+      DROPIN_PATH = "/etc/systemd/system/kubelet.service.d/11-pharos.conf"
```
jakolehm: 11 because it needs to be loaded after kubeadm generated dropin (it's 10).

lib/pharos/phases/migrate_master.rb
```diff
@@ -6,19 +6,20 @@ class MigrateMaster < Pharos::Phase
       title "Migrate master"
 
       def call
-        migrate_0_5_to_0_6 if migrate_0_5_to_0_6?
```
jakolehm: Removing old migration because we don't really support direct upgrade from 0.5.x to 1.2.x.

lib/pharos/scripts/wait-kubelet-proxy.sh
```diff
@@ -4,7 +4,7 @@ set -e
 
 echo "Waiting kubelet-proxy to launch on port 6443..."
 
-while ! (echo > /dev/tcp/localhost/6443) >/dev/null 2>&1; do
+while ! ( bash -c 'echo > /dev/tcp/localhost/6443') >/dev/null 2>&1; do
```
jakolehm: Fixes regression from CentOS/RHEL feature.. needs to call explicitly bash because Ubuntu has dash as sh.

lib/pharos/ssh/remote_file.rb
```diff
@@ -20,7 +20,7 @@ def initialize(client, path)
 
       # Removes the remote file
       def unlink
-        @client.exec!("rm #{escaped_path}")
+        @client.exec!("sudo rm #{escaped_path}")
```
jakolehm: Fixes bug when ssh user is not root.

jakolehm:
@jnummelin PTAL

PR #432, opened 2018-06-14
Cluster reset command

### LINKED ISSUE #128: Reset command
Should remove all traces of pharos installation.

jakolehm:
Fixes #128 

### TODO
- [x] ubuntu
- [x] centos/rhel


jakolehm:
CentOS/RHEL functionality is missing proper unlocking of packages.

jakolehm:
Still leaves zombie containers running.. we need to remove containers before removing docker/cri-o.

jakolehm:
Tested: Ubuntu (docker & cri-o), CentOS (docker)

jakolehm:
@kke PTAL

SpComb:
IMO the ideal way to implement this would be to have an (optional) `reset` method on each phase, and then `pharos-cluster reset` could run the `reset` method across the same phases as it would otherwise run... but this is functionally equivalent, and it can be refactored later.

lib/pharos/host/ubuntu/scripts/configure-cri-o.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-set -ue
```
SpComb: Surprising, wny drop the `set -u`?

lib/pharos/host/ubuntu/scripts/reset.sh
```diff
@@ -0,0 +1,37 @@
...
+    systemctl disable docker
+elif which crictl ; then
+    systemctl stop crio
+    systemctl disable crio
+    crictl rm $(crictl ps -a -q)
+fi
+
+kubeadm reset
+
+apt-get purge -y --allow-change-held-packages --purge kubeadm kubelet kubectl docker.io cri-o-${CRIO_VERSION}
```
SpComb: `apt-get purge 'cri-o-*'` should also work

lib/pharos/reset_command.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module Pharos
+  class ResetCommand < UpCommand
```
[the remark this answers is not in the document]
kke: It will have all the `--tf-json`things and whatever is added to `pharos up` too then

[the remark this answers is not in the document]
SpComb: Agreed... the `load_config` etc that's common to both `UpCommand` and `ResetCommand` should come from some common `ClusterCommand` class instead?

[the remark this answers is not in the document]
jakolehm: I think we can refactor this later?

lib/pharos/host/ubuntu/scripts/reset.sh
jakolehm: Good catch, will fix this.

jakolehm:
Added cri-o (see #449) cleanup.

SpComb:
Needs merge

lib/pharos/host/ubuntu/ubuntu_xenial.rb
```diff
@@ -52,6 +52,13 @@ def configure_container_runtime
           raise Pharos::Error, "Unknown container runtime: #{host.container_runtime}"
         end
       end
+
+      def reset
+        exec_script(
+          "reset.sh",
+          CRIO_VERSION: CRIO_VERSION
```
SpComb: The `CRIO_VERSION` is no longer used, don't need to override.

lib/pharos/host/el7/scripts/reset.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/sh
+
+rm -rf /etc/kubernetes/manifests
+sleep 5
+systemctl stop kubelet
+systemctl disable kubelet
+
+if which docker ; then
```
SpComb: CentOS `/usr/bin/docker` is a wrapper script provided by `docker-common`, not the `docker` package

SpComb:
The `apt-get purge docker.io` is broken, but that's an upstream ubuntu packaging bug...

```
    /var/lib/docker/nuke-graph-directory.sh: 64: /var/lib/docker/nuke-graph-directory.sh:     shopt: not found    
    dpkg: error processing package docker.io (--purge):
     subprocess installed post-removal script returned error exit status 127
```

SpComb:
CRI-O cleanup seems to leave the `/pause` containers running...

jakolehm:
> CRI-O cleanup seems to leave the /pause containers running...

Fixed.

lib/pharos/phases/gather_facts.rb
```diff
@@ -0,0 +1,141 @@
+# frozen_string_literal: true
+
+require 'ipaddr'
+
+module Pharos
+  module Phases
+    class GatherFacts < Pharos::Phase
```
jakolehm: Refactored to a separate phase because `reset` should not validate hosts.

lib/pharos/phases/validate_cluster_hosts.rb
```diff
@@ -2,10 +2,14 @@
 
 module Pharos
   module Phases
-    class ValidateHostname < Pharos::Phase
-      title "Validate unique hostnames for all hosts"
+    class ValidateClusterHosts < Pharos::Phase
```
jakolehm: All cluster wide host validations should be under this phase. New title should make it more clear.

jakolehm:
Output:
image

lib/pharos/cluster_manager.rb
```diff
@@ -44,10 +44,15 @@ def load
...
+    def gather_facts
+      apply_phase(Phases::GatherFacts, config.hosts, ssh: true, parallel: true)
+    end
+
     def validate
       addon_manager.validate
+      gather_facts
       apply_phase(Phases::ValidateHost, config.hosts, ssh: true, parallel: true)
-      apply_phase(Phases::ValidateHostname, config.hosts, ssh: false, parallel: false)
+      apply_phase(Phases::ValidateClusterHosts, config.hosts, ssh: false, parallel: false)
```
SpComb: If we `GatherFacts` first to get the `@host.hostname`, then the `ValidateHostname` logic can also be done in `ValidateHost`.

SpComb: IOW, the reason for the separate `ValidateHostClusters` phase was that the hostnames were gathered in parallel as part of `ValidateHost`... so there's no more need to separate the two validation phases anymore.

jakolehm: Good catch, I'll refactor this more.

jakolehm:
@nevalla PTAL

PR #436, opened 2018-06-18
Fix output error when color is disabled

### LINKED ISSUE #435: Problems running on CI server
I have a cluster up and running just fine. Working now on automating the deployment via our CI server (teamcity), and have hit a few snags:

1.  Running in teamcity, I get an error after gathering host facts. Traces to [this line](https://github.com/kontena/pharos-cluster/blob/v1.2.0-beta.1/lib/pharos/up_command.rb#L144), with something like "yaml not defined". Probably has something to do with colors not being supported, and that code path has a bug or something.
2. I am not sure if I would be able to proceed if that were fixed, because the console would not be interactive, and couldn't hit `y`. Perhaps a flag for that would be useful.

Is this a target use case, or is it assumed that pharos will always be run by a person?

jakolehm:
Fixes #435

### DIFF
--- a/lib/pharos/command.rb
+++ b/lib/pharos/command.rb
@@ -24,5 +24,13 @@ module Pharos
     def rouge
       @rouge ||= Rouge::Formatters::Terminal256.new(Rouge::Themes::Github.new)
     end
+
+    def tty?
+      $stdin.tty?
+    end
+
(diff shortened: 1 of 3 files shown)


jakolehm:
@kke PTAL

PR #441, opened 2018-06-20
Fix CentOS/RHEL7 kubelet cgroups settings

### LINKED ISSUE #440: CentOS kubelet complains about systemd slices
We should probably set `--runtime-cgroups=/systemd/system.slice --kubelet-cgroups=/systemd/system.slice` as kubelet args on CentOS/RHEL.

jakolehm:
Fixes #440

### DIFF
--- a/lib/pharos/host/configurer.rb
+++ b/lib/pharos/host/configurer.rb
@@ -33,6 +33,11 @@ module Pharos
         abstract_method!
       end
 
+      # @return [Array<String>]
+      def kubelet_extra_args
+        abstract_method!
+      end
+
       # @param args [Hash]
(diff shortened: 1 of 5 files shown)


jakolehm:
@SpComb PTAL

SpComb:
> @SpComb PTAL

No idea what the issue or fix is about, let me bring up a CentOS machine and take a look...

SpComb:
NACK, see https://github.com/kontena/pharos-cluster/issues/440#issuecomment-399014418 for why this workaround is a terrible idea, and https://github.com/kontena/pharos-cluster/pull/446 for a better alternative that actually fixes the issue.

jakolehm:
Fixed in kubelet rpm package.

PR #442, opened 2018-06-20
Show deprecation notice in kured output

jakolehm:


jakolehm:
> would be better if these warnings were in yellow text

Addon does not provide good interface for handling colors (pastel?).

PR #444, opened 2018-06-20
Don't add --node-ip if cloud.provider is set

### LINKED ISSUE #368: --node-ip should not be set if cloud.provider is enabled
Because cloud provider should set all ip's. Setting this manually causes cloud provider to not set externalIP.

jakolehm:
Fixes #368

### DIFF
--- a/spec/pharos/phases/configure_kubelet_spec.rb
+++ b/spec/pharos/phases/configure_kubelet_spec.rb
@@ -102,6 +102,12 @@ describe Pharos::Phases::ConfigureKubelet do
           '--cloud-config=/etc/pharos/kubelet/cloud-config'
         )
       end
+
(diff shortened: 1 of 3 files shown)


spec/pharos/phases/configure_kubelet_spec.rb
```diff
@@ -102,6 +102,12 @@
           '--cloud-config=/etc/pharos/kubelet/cloud-config'
         )
       end
+
+      it 'does not add node ip' do
+        expect(subject.kubelet_extra_args).not_to include(
+          "--node-ip=#{host.peer_address}"
```
SpComb: Weak spec, `expect(...).to_not include(/--node-ip/)` would be better.

PR #450, opened 2018-06-25
Detect systemd-resolved stub resolver at localhost and bypass using kubelet --resolv-conf

### LINKED ISSUE #448: Ubuntu 18.04 dns problems in pods
### Symptoms:

kubectl forward doens't work, Example:

    kubectl -n kube-system port-forward service/tiller-deploy 44134:44134

Results in 

> E0624 12:06:20.664091   34312 portforward.go:331] an error occurred forwarding 42399 -> 44134: error forwarding port 44134 to pod 255e06439c2da94a4b6a8b1ad2d3d7f4d6d1ba1f82ab6eb2ae519133b1f2bc58, uid : exit status 1: 2018/06/24 15:06:20 socat[22114] E getaddrinfo("localhost", "NULL", {1,2,1,6}, {}): Temporary failure in name resolution

Geting the log in kubelet on the worker node 

    journalctl -u kubelet

Results in something like:

> 7797 httpstream.go:251] error forwarding port 44134 to pod 255e06439c2da94a4b6a8b1ad2d3d7f4d6d1ba1f82ab6eb2ae519133b1f2bc58, uid : exit status 1: 2018/06/24 15:06:21 socat[22130] E getaddrinfo("localhost", "NULL", {1,2,1,6}, {}): Temporary failure in namJun 24 15:09:53 worker0 kubelet[7797]: E0624 15:09:53.133330    7797 httpstream.go:251] error forwarding port 44134 to pod 255e06439c2da94a4b6a8b1ad2d3d7f4d6d1ba1f82ab6eb2ae519133b1f2bc58, uid : exit status 1: 2018/06/24 15:09:53 socat[22418] E getaddrinfo("localhost", "NULL", {1,2,1,6}, {}): Temporary failure in namJun 24 15:15:56 worker0 kubelet[7797]: E0624 15:15:56.303891    7797 httpstream.go:251] error forwarding port 44134 to pod 

Which points to **getaddrinfo("localhost")...**, that means pod is not able to resolve localhost.

Other commands that use port-forward like *helm version* or other helm commands (helm package interact with tiller server using port forwards) has the same symptoms

### Cause

Ubuntu 18.04 uses **systemd-resolve** which change /etc/resolv.conf to use a local dns. Kubelet needs to be started with the flag **--resolv-conf=/run/systemd/resolve/resolv.conf** on this systems.

### Possible Solutions

This have been addressed by https://github.com/kubernetes/kubeadm/issues/787 and probably will be on kubernetes 1.11, as a work around there is two easy fix:

1. Create a systemd dropping to override kubelet. Example:

```
[Service]
Environment='KUBELET_DNS_ARGS=--cluster-dns=172.31.0.10 --cluster-domain=cluster.local --resolv-conf=/run/systemd/resolve/resolv.conf'
```
Restart kubelet:

    systemctl daemon-reload
    systemctl restart kubelet

2. Create a symlink on /etc/resolv.conf pointing to /run/systemd/resolve/resolv.conf (backup first)

Do the same on all machines.

### Pharos Installer

I suggest that pharos-cluster check for the existence of the file in Ubuntu 18.04 and perform one of the fix above. We need to check after the kubeadm fix is released to make sure it do not conflict (possible resulting in two flags added). Check the pull request to see how it was fixed in kubeadm side.

SpComb:
Fixes #448 pod `dnsPolicy: Default` behavior on Ubuntu bionic with the systemd-resolved stub resolver enabled.

This covers both the general issue and the specific fix for `systemd-resolved`: it uses the upstream resolvers if available, and otherwise fails if the host is configured to use localhost nameservers.

### DIFF
--- a/lib/pharos/configuration/host.rb
+++ b/lib/pharos/configuration/host.rb
@@ -20,7 +20,7 @@ module Pharos
       attribute :http_proxy, Pharos::Types::Strict::String
 
       attr_accessor :os_release, :cpu_arch, :hostname, :api_endpoint, :private_interface_address, :checks
-      attr_accessor :resolv_localhost
+      attr_accessor :resolv_localhost, :systemd_resolved_stub
 
       def to_s
         address
--- a/lib/pharos/configuration/host.rb
+++ b/lib/pharos/configuration/host.rb
@@ -8,6 +8,11 @@ module Pharos
     class Host < Dry::Struct
       constructor_type :schema
 
+      class ResolvConf < Dry::Struct
+        attribute :nameserver_localhost, Pharos::Types::Strict::Bool
+        attribute :systemd_resolved_stub, Pharos::Types::Strict::Bool
+      end
+
       attribute :address, Pharos::Types::Strict::String
       attribute :private_address, Pharos::Types::Strict::String
       attribute :private_interface, Pharos::Types::Strict::String
@@ -19,8 +24,7 @@ module Pharos
       attribute :container_runtime, Pharos::Types::Strict::String.default('docker')
       attribute :http_proxy, Pharos::Types::Strict::String
 
-      attr_accessor :os_release, :cpu_arch, :hostname, :api_endpoint, :private_interface_address, :checks
-      attr_accessor :resolv_localhost, :systemd_resolved_stub
+      attr_accessor :os_release, :cpu_arch, :hostname, :api_endpoint, :private_interface_address, :checks, :resolvconf
 
       def to_s
         address
--- a/lib/pharos/configuration/host.rb
+++ b/lib/pharos/configuration/host.rb
@@ -20,6 +20,7 @@ module Pharos
       attribute :http_proxy, Pharos::Types::Strict::String
 
       attr_accessor :os_release, :cpu_arch, :hostname, :api_endpoint, :private_interface_address, :checks
+      attr_accessor :resolv_localhost
 
       def to_s
         address
--- a/lib/pharos/autoload.rb
+++ b/lib/pharos/autoload.rb
@@ -45,6 +45,7 @@ module Pharos
   module Configuration
     autoload :Host, 'pharos/configuration/host'
     autoload :Taint, 'pharos/configuration/taint'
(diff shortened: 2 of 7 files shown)


SpComb:
```
root@terom-bionic-test:~# cat /etc/systemd/system/kubelet.service.d/11-pharos.conf 
[Service]
Environment='KUBELET_EXTRA_ARGS=--read-only-port=0 --node-ip=206.189.0.173 --hostname-override=terom-bionic-test --pod-infra-container-image=quay.io/kontena/pause-amd64:3.1'
Environment='KUBELET_DNS_ARGS=--cluster-dns=10.250.0.10 --cluster-domain=cluster.local --resolv-conf=/run/systemd/resolve/resolv.conf'
ExecStartPre=-/sbin/swapoff -a
```

lib/pharos/configuration/host.rb
```diff
@@ -20,6 +20,7 @@ class Host < Dry::Struct
       attribute :http_proxy, Pharos::Types::Strict::String
 
       attr_accessor :os_release, :cpu_arch, :hostname, :api_endpoint, :private_interface_address, :checks
+      attr_accessor :resolv_localhost, :systemd_resolved_stub
```
jakolehm: probably should be under `checks`?

SpComb: Dunno, it's just as much metadata as `os_reselase` and `cpu_arch` are? It's not a healthcheck...

jakolehm: ok, maybe then an accessor that has both attributes?

SpComb: A new `class Pharos::Configuration::HostResolver < Dry::Struct`?

SpComb: Alternatively, we could do these checks directly in the `ConfigureKubelet` phase, and skip the host attributes... they are not necessary in this case.

jakolehm: A new class sounds good to me.

SpComb:
> @SpComb ok to merge?

Yes, tested with mixed xenial + bionic

PR #473, opened 2018-07-04
[WiP] Upgrade to kubernetes 1.11

### LINKED ISSUE #419: Kubernetes 1.11
https://github.com/kubernetes/sig-release/blob/master/releases/release-1.11/release-1.11.md

### LINKED ISSUE #472: etcd version upgrade does not work
The `configure-etcd.sh` script incorrectly uses `if [ ! func_call ]` as a conditional, which is always false (the `func_call` string literal is never an empty string): https://github.com/kontena/pharos-cluster/blob/0561caa3470593b1d53fa1ea21073da68c6523e1/lib/pharos/scripts/configure-etcd.sh#L11

This prevents changes to `ETCD_VERSION=...` from triggering a new `/etc/kubernetes/manifests/pharos-etcd.yaml` write.

SpComb:
Fixes #419 
Fixes #472

# TODO
- [ ] worker node upgrades
- [ ] Build new packages/images
  - [ ] `https://dl.bintray.com/kontena/pharos-debian` xenial/bionic 1.11.0
  - [ ] `https://dl.bintray.com/kontena/pharos-rpm` 1.11.0
  - [ ] `quay.io/kontena` `kube-apiserver/kube-controllermanager/kube-proxy/kube-scheduler` 1.11.0
  - [ ] `quay.io/kontena` `etcd ` 3.2.18
  - [ ] `quay.io/kontena` `coredns` 1.1.3
- [ ] Test
  - [ ] Optional configurations: external etcd, cri-o, ...


SpComb:
`kubeadm` upgrade now requires `cri-tools`:

```
    $ sudo VERSION=1.11.0 ARCH=amd64 sh -x < upgrade-kubeadm.sh
    +     set     -ex    
    +     kubeadm     version     -o     short    
    + [ v1.10.4 = v1.11.0 ]
    + cd /tmp
    + export DEBIAN_FRONTEND=noninteractive
    + apt-get download kubeadm=1.11.0-00
    Get:1 https://packages.cloud.google.com/apt kubernetes-xenial/main amd64 kubeadm amd64 1.11.0-00 [9,422 kB]
    Fetched 9,422 kB in 1s (6,268 kB/s)
    + dpkg -i --ignore-depends=kubelet kubeadm_1.11.0-00_amd64.deb
    (Reading database ... 83419 files and directories currently installed.)
    Preparing to unpack kubeadm_1.11.0-00_amd64.deb ...
    Unpacking kubeadm (1.11.0-00) over (1.10.4-00) ...
    dpkg: dependency problems prevent configuration of kubeadm:
     kubeadm depends on cri-tools (>= 1.11.0); however:
      Package cri-tools is not installed.
    
    dpkg: error processing package kubeadm (--install):
     dependency problems - leaving unconfigured
    Errors were encountered while processing:
     kubeadm
    ! 1
```

SpComb:
Upgrading the `kubeadm` package to run the `kubeadm upgrade` breaks the kubelet, because the new `kubeadm` package updates the `/etc/systemd/system/kubelet.service.d/10-kubeadm.conf` file to run the kubelet with `--config=/var/lib/kubelet/config.yaml`, but the `kubeadm upgrade` step does not write out this config, so it gets stuck with the `kubelet` in a restart loop on the missing config.

This is a documented limitation for the kubeadm upgrade path: https://kubernetes.io/docs/tasks/administer-cluster/kubeadm/kubeadm-upgrade-1-11/#upgrade-the-control-plane

> Note that upgrading the kubeadm package on your system prior to upgrading the control plane causes a failed upgrade. Even though kubeadm ships in the Kubernetes repositories, it’s important to install it manually. The kubeadm team is working on fixing this limitation.

SpComb:
Initial install and upgrade from pharos 1.2 should work now, tested with custom hacks to use upstream repos instead of the pharos ones.

SpComb:
Further testing/work pending on updated pharos images/packages.

SpComb:
`kubeadm reset` now prompts:

```
+ kubeadm reset
[reset] WARNING: changes made to this host by 'kubeadm init' or 'kubeadm join' will be reverted.
[reset] are you sure you want to proceed? [y/N]: Aborted reset operation
/app/lib/pharos/ssh/remote_command.rb:66:in `run!'
/app/lib/pharos/ssh/client.rb:60:in `exec!'
/app/lib/pharos/ssh/client.rb:72:in `exec_script!'
/app/lib/pharos/host/configurer.rb:74:in `exec_script'
/app/lib/pharos/host/el7/el7.rb:82:in `reset'
/app/lib/pharos/phases/reset_host.rb:10:in `call'
/app/lib/pharos/phase_manager.rb:70:in `block in apply'
/app/lib/pharos/phase_manager.rb:26:in `block (2 levels) in run_parallel'
```

lib/pharos/host/el7/scripts/upgrade-kubeadm.sh
```diff
@@ -8,7 +8,11 @@ if [ $(kubeadm version -o short) = "v${VERSION}" ]; then
...
-yum install -y "kubeadm-${VERSION}"
-lineinfile "^0:kubeadm-" "0:kubeadm-${VERSION}-0.*" $versionlock
\ No newline at end of file
+BIN_URL="https://dl.bintray.com/kontena/pharos-bin/kube/${VERSION}/kubeadm-${ARCH}.gz"
+
+curl -fsSL $BIN_URL -o /tmp/kubeadm.gz
+curl -fsSL "${BIN_URL}.asc" -o tmp/kubeadm.gz.asc
+gpg --verify /tmp/kubeadm.gz.asc /tmp/kubeadm.gz
+gunzip /tmp/kubeadm.gz
+install -o root -g root -m 0755 -t /usr/local/bin /tmp/kubeadm # XXX: overrides package version?
```
SpComb: This should be installed somewhere temporarily only for the duration of the upgrade, only until the kubeadm package can be upgraded... leaving behind a version of kubeadm in `/usr/local/bin` is bad.

SpComb: This also gets left behind on `pharos-cluster reset` and causes the next `pharos-cluster up` to run using the wrong version of kubeadm.

SpComb:
With `kubeadm` now depending on the separate `cri-tools` package, our own `cri-o` package should no longer be including `/usr/local/bin/crictl`?

SpComb:
Upgrading a CentOS 7 worker node from 1.10 -> 1.11 with `kubeadm upgrade node config` leaves the kubelet confiured with the default `cgroupDriver: cgroupfs`:

```
Jul 25 11:36:10 terom-centos-test kubelet[5939]: F0725 11:36:10.541309    5939 server.go:262] failed to run Kubelet: failed to create kubelet: misconfiguration: kubelet cgroup driver: "cgroupfs" is different from docker cgroup driver: "systemd"
```

Per https://kubernetes.io/docs/setup/independent/install-kubeadm/#configure-cgroup-driver-used-by-kubelet-on-master-node kubeadm should be detecting it and configuring it automatically, but this isn't happening:

```
[root@terom-centos-test ~]# cat /var/lib/kubelet/kubeadm-flags.env
cat: /var/lib/kubelet/kubeadm-flags.env: No such file or directory
[root@terom-centos-test ~]# grep cgroupDriver /var/lib/kubelet/config.yaml
cgroupDriver: cgroupfs
```

lib/pharos/host/ubuntu/scripts/install-kube-packages.sh
```diff
@@ -7,6 +7,9 @@ apt-mark unhold kubelet kubectl kubeadm
 apt-get install -y kubelet=${KUBE_VERSION}-00 kubectl=${KUBE_VERSION}-00 kubeadm=${KUBEADM_VERSION}-00
 apt-mark hold kubelet kubectl kubeadm
 
+# use KUBELET_EXTRA_ARGS from /etc/systemd/system/kubelet.service.d/11-pharos.conf instead
+sed -i 's/^KUBELET_EXTRA_ARGS=/#\0/' /etc/default/kubelet
```
SpComb: Maybe `ConfigureKubelet` should be writing the `/etc/default/kubelet` instead, but that's annoying since it varies by OS.

lib/pharos/host/ubuntu/scripts/upgrade-kubeadm.sh
```diff
@@ -2,12 +2,15 @@
...
 
-cd /tmp
-export DEBIAN_FRONTEND=noninteractive
-apt-get download kubeadm=${VERSION}-00
-dpkg -i --ignore-depends=kubelet kubeadm_${VERSION}*.deb
-rm -f kubeadm_${VERSION}*.deb
\ No newline at end of file
+BIN_URL="https://dl.bintray.com/kontena/pharos-bin/kube/${VERSION}/kubeadm-${ARCH}.gz"
+
+curl -fsSL $BIN_URL -o /tmp/kubeadm.gz
```
jakolehm: We need to curl binary because kubeadm decided to depend on `cri-tools`?

lib/pharos/kubeadm.rb
```diff
@@ -121,23 +125,27 @@ def configure_internal_etcd(config)
           "https://#{h.peer_address}:2379"
         }
         config['etcd'] = {
-          'endpoints' => endpoints
+          'external' => {
+            'endpoints' => endpoints
```
jakolehm: Why don't we set `*File` keys here directly?

lib/pharos/phases/configure_dns.rb
```diff
@@ -6,7 +6,8 @@ class ConfigureDNS < Pharos::Phase
       title "Configure DNS"
 
       def call
-        patch_kubedns(
+        patch_deployment(
+          'coredns',
```
jakolehm: Does this kube-dns -> coredns change cause downtime when upgrading from Pharos 1.2?

lib/pharos_cluster.rb
```diff
@@ -9,7 +9,7 @@
 
 module Pharos
   CRIO_VERSION = '1.10.6'
-  KUBE_VERSION = ENV.fetch('KUBE_VERSION') { '1.10.5' }
+  KUBE_VERSION = ENV.fetch('KUBE_VERSION') { '1.11.0' }
   KUBEADM_VERSION = ENV.fetch('KUBEADM_VERSION') { KUBE_VERSION }
   ETCD_VERSION = ENV.fetch('ETCD_VERSION') { '3.1.12' }
```
jakolehm: Etcd version is still the same?

lib/pharos/host/ubuntu/scripts/upgrade-kubeadm.sh
SpComb: No, it's because upgrading the `kubeadm` package will upgrade the `/etc/systemd/system/kubelet.service.d/10-kubeadm.conf` to reference the `/var/lib/kubelet/config.yaml` and place the kubelet into a crash-and-restart loop.

I don't seem to have copied the exact error, but IIRC this somehow prevented the `kubeadm upgrade host config` from working, deadlocking the upgrade.

lib/pharos/kubeadm.rb
SpComb: Dunno, kept the existing code structure as-is... I can refactor it at the same time, need to test the external etcd setup anyways.

SpComb: ...probably because the external etcd case has all of these as conditionals, and it's a pain to write any other way.

lib/pharos_cluster.rb
SpComb: Myes, seems to have gotten lost in all of the `git add -p` I had to do to test this with the external kube repos.

```
    	[WARNING ExternalEtcdVersion]: this version of kubeadm only supports external etcd version >= 3.2.17. Current version: 3.1.12
```

lib/pharos/phases/configure_dns.rb
SpComb: Maybe, I didn't test to see how it behaves exactly.

If coredns is the new default, then I don't think we'd want to stick with kube-dns long-term. Options are:

* stick with just kube-dns for 1.2 and deal with it later
* mandatory migration to coredns for 1.2 
  * avoids any need to implement support for switching between kube-dns/coredns
* provide a gradual migration path for kube-dns -> coredns:
  * use coredns for new installs
  * keep kube-dns for existing installs
  * provide a `cluster.yml` knob to switch from kube-dns to coredns (or revert to `kube-dns` if coredns has some regressions)
  * eventually deprecate and remove `kube-dns`

SpComb: > keep kube-dns for existing installs

This would be non-trivial to implement, because it would require kube API access before `UpgradeMaster` / `ConfigureMaster` runs and generates the kubeadm config.

jakolehm:
CoreDNS being a multiarch image is causing headaches. First I tried to build coredns on separate repos for each architecture (like every other k8s image) but kubeadm wants to use multiarch -> breaks upgrade because image pull does not work. 

Then I tried to actually create a multiarch repo (with https://github.com/estesp/manifest-tool) but it seems that quay.io does not support v2.2 manifests (they are currently implementing it, see: https://github.com/moby/buildkit/issues/409#issuecomment-394757219 .

jakolehm:
> kubeadm wants to use multiarch -> breaks upgrade because image pull does not work.

Not completely sure if this is actually true. Based on logs it seems that CoreDNS addon is applied but then upgrade hangs on when it tries to remove kube-dns deployment.

```
    [bootstraptoken] configured RBAC rules to allow Node Bootstrap tokens to post CSRs in order for nodes to get long term certificate credentials
    [bootstraptoken] configured RBAC rules to allow the csrapprover controller automatically approve CSRs from a Node Bootstrap Token
    [bootstraptoken] configured RBAC rules to allow certificate rotation for all node client certificates in the cluster
    [addons] Applied essential addon: CoreDNS
... HANGS ...
```

jakolehm:
Reason why it halts there is that kubeadm waits for coredns replicas... this cannot happen because if invalid image name: https://github.com/kubernetes/kubernetes/blob/master/cmd/kubeadm/app/phases/upgrade/postupgrade.go#L141-L149

lib/pharos/phases/configure_dns.rb
jakolehm: > Does this kube-dns -> coredns change cause downtime when upgrading from Pharos 1.2?

Nope, it should be smooth ride (kube-dns deployment is removed after coredns is running).

lib/pharos/phases/configure_dns.rb
```diff
@@ -45,52 +46,53 @@ def max_unavailable
...
-                                "kube-dns"
-                              ]
-                            }
-                          ]
-                        },
-                        topologyKey: "kubernetes.io/hostname"
-                      }
-                    ]
-                  }
+        Pharos::Kube.session(@master.api_address).resource_client('apps/v1').patch_deployment(
```
jakolehm: Patch should be ok here, right?

SpComb: Yes, the `PATCH` matches what `kubectl set image` does:

```
$ kubectl -v8 -n kube-system set image deployments/coredns coredns=quay.io/kontena/coredns-amd64:1.1.3
...
I0807 11:11:05.700941    8436 request.go:874] Request Body: {"spec":{"template":{"spec":{"$setElementOrder/containers":[{"name":"coredns"}],"containers":[{"image":"quay.io/kontena/coredns-amd64:1.1.3","name":"coredns"}]}}}}
I0807 11:11:05.700985    8436 round_trippers.go:383] PATCH https://167.99.39.233:6443/apis/extensions/v1beta1/namespaces/kube-system/deployments/coredns
...
```

Not entirely sure what the `$setElementOrder` is doing, but it seems unnecessary. Apart from that the container image part is identical.

jakolehm: Just a note that this also changes how dns replicas/maxSurge/etc is sent.

SpComb: Yes, those are simple object-level PATCHes, no array merge needed. Verified that the end result is as intended.

lib/pharos/phases/configure_dns.rb
```diff
@@ -45,52 +46,53 @@ def max_unavailable
...
+                  affinity: {
+                    podAntiAffinity: {
+                      requiredDuringSchedulingIgnoredDuringExecution: [
+                        {
+                          labelSelector: {
+                            matchExpressions: [
+                              {
+                                key: "k8s-app",
+                                operator: "In",
+                                values: [name]
```
SpComb: The `podAntiAffinity` is broken, the pods get scheduled on the same node.

The `coredns` deployment uses a `k8s-app: kube-dns` label, so this shouldn't be `coredns`.

SpComb: Fixed 

```
NAME                                          READY     STATUS    RESTARTS   AGE       IP               NODEt
coredns-dcb4c7ddd-pn96m                       1/1       Running   0          1m        10.32.1.8        terom-xenial-test
coredns-dcb4c7ddd-tk2dm                       1/1       Running   0          1m        10.32.2.29       terom-bionic-test
```

SpComb:
Quick smoketest for this is passing, so this should be good enough to merge and continue testing in master together with other changes.

CI needs #504 to pass.

The CoreDNS image hack is unfortunate, but I don't know how.

PR #474, opened 2018-07-13
Upgrade dry-struct, dry-validation

SpComb:
Required for using the `pharos-kube-client` gem which depends on the newer version of dry-struct.


lib/pharos/addon.rb
```diff
@@ -24,7 +24,8 @@ def self.addon(name, &block)
   class Addon
     include Pharos::Logging
 
-    class Schema < Dry::Validation::Schema
+    # return class as use for superclass in Dry::Validation.Params
+    Schema = Dry::Validation.Schema(build: false) do
```
SpComb: :unamused: 

https://github.com/dry-rb/dry-validation/blob/v0.12.1/lib/dry/validation.rb#L20-L24

lib/pharos/configuration.rb
```diff
@@ -0,0 +1,12 @@
+require 'dry-struct'
+
+module Pharos
+  module Configuration
+    class Struct < Dry::Struct
+      transform_types do |type|
+        # all attributes are optional and default to nil...
+        type.meta(omittable: true)
```
SpComb: I don't think we should be leaving all of these structs so weakly typed, but fixing all the optional attributes to be optional and/or fixing the specs to pass all required attributes would need work... plus duplicating more of the `ConfigSchema` `dry-validation` rules?

spec/pharos/phases/label_node_spec.rb
```diff
@@ -102,7 +102,7 @@
             labels: { },
           },
           spec: {
-            taints: [ { key: 'node-role.kubernetes.io/master', value: nil, effect: 'NoSchedule' } ],
```
SpComb: Needs merge for #471 changes, specs are failing after master merge... I wonder why this doens't show up in drone.

lib/pharos/addon.rb
```diff
@@ -84,7 +83,7 @@ def to_h
       end
 
       def config_schema(&block)
-        @schema = Dry::Validation.Form(Schema, &block)
+        @schema = Dry::Validation.Params(Schema, &block)
```
jakolehm: Why `Form` -> `Params` ?

SpComb: https://github.com/dry-rb/dry-validation/pull/407

PR #477, opened 2018-07-18
Bump xenial to docker version 17.03.2 from xenial-updates docker.io

### LINKED ISSUE #476: Ubuntu xenial no longer provides docker.io 1.13.1
Pharos is not working anymore as it is using docker package which is missing from the repository.

The error is: `Version '1.13.1-0ubuntu1~16.04.2' for 'docker.io' was not found`

Version of this package is hardcoded at the `ubuntu_xebial.rb`:

```
34       def configure_container_runtime¬
 35         if docker?¬
 36           exec_script(¬
 37             'configure-docker.sh',¬
 38             DOCKER_PACKAGE: 'docker.io',¬
 39             DOCKER_VERSION: "#{DOCKER_VERSION}-0ubuntu1~16.04.2"¬
 40           )
```

SpComb:
Fixes #476

Upgrade Ubuntu xenial hosts from Docker 1.13.1 -> 17.03.2 per the supported Ubuntu `xenial-updates` version.

### DIFF
--- a/lib/pharos/host/ubuntu/ubuntu_xenial.rb
+++ b/lib/pharos/host/ubuntu/ubuntu_xenial.rb
@@ -7,7 +7,7 @@ module Pharos
     class UbuntuXenial < Ubuntu
       register_config 'ubuntu', '16.04'
 
-      DOCKER_VERSION = '1.13.1'
+      DOCKER_VERSION = '17.03.2'
       CFSSL_VERSION = '1.2'
 
       register_component(
@@ -36,7 +36,7 @@ module Pharos
           exec_script(
             'configure-docker.sh',
             DOCKER_PACKAGE: 'docker.io',
-            DOCKER_VERSION: "#{DOCKER_VERSION}-0ubuntu1~16.04.2"
+            DOCKER_VERSION: "#{DOCKER_VERSION}-0ubuntu2~16.04.1"
           )
         elsif crio?
           exec_script(


SpComb:
Upgrading the `docker.io` package does not restart docker with the new version...

```
root@terom-pharos-master:~# docker info -f '{{.ServerVersion}}'
1.13.1
root@terom-pharos-master:~# systemctl restart docker
root@terom-pharos-master:~# docker info -f '{{.ServerVersion}}'
17.03.2-ce
```

wolfedale:
The fix is working on my environment :-)

SpComb:
> Upgrading the docker.io package does not restart docker with the new version...

This seems to be a general policy, it will prompt via debconf for restarting on package upgrades, which we skip per `DEBIAN_FRONTEND=noninteractive`.

```
# restart Docker if it's running (thus this is an upgrade) and the user (via debconf) tells us they're OK with that
db_version 2.0
db_fset docker.io/restart seen false
if [ -n "$2" ] && [ "$1" = 'configure' ] \
	&& [ -z "${DEBCONF_RECONFIGURE:-}" ] \
	&& invoke-rc.d --disclose-deny docker status > /dev/null 2>&1 \
; then
	db_input high docker.io/restart || :
	db_go || :
	db_get docker.io/restart
	if [ "$RET" = 'true' ]; then
		# Docker daemon appears to be running and should be restarted
		invoke-rc.d docker restart || :
	fi
fi
db_stop
```

I presume we way to to consider configuring this, particularly since we're running with `{"live-restore": true}`:

```sh
debconf-set-selections <<EOF
docker.io docker.io/restart boolean true
EOF
```

PR #480, opened 2018-07-18
Upgrade cri-o to v1.10.6

jakolehm:
Includes a security fix, see: 
- http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2018-10892
- https://github.com/kubernetes-incubator/cri-o/releases/tag/v1.10.6

### DIFF
--- a/lib/pharos_cluster.rb
+++ b/lib/pharos_cluster.rb
@@ -8,11 +8,9 @@ require_relative "pharos/error"
 require_relative "pharos/root_command"
 
 module Pharos
-  CRIO_VERSION = '1.10.4'
-  CRICTL_VERSION = '1.0.0-beta.0'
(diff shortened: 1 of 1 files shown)


lib/pharos_cluster.rb
```diff
@@ -8,11 +8,9 @@
 require_relative "pharos/root_command"
 
 module Pharos
-  CRIO_VERSION = '1.10.4'
-  CRICTL_VERSION = '1.0.0-beta.0'
```
jakolehm: Removed `CRICTL_VERSION` and `DOCKER_VERSION` because they are not used anymore.

jakolehm:
Binary bundle is already in place: https://bintray.com/kontena/pharos-bin/cri-o/view/files?sort=name&order=asc#files/cri-o

PR #491, opened 2018-07-30
Update weave net to 2.4.0

jakolehm:
https://github.com/weaveworks/weave/releases/tag/v2.4.0

Todo:

- [x] update permissions
- [x] update pharos images


SpComb:
Release notes should mention that this upgrade will cause a rolling ~20s (on my dev cluster) network outage across each node in turn.

PR #494, opened 2018-08-01
cri-o: add docker.io to default registries

jakolehm:
Currently it's kinda surprising that non-qualified images work with docker but not with cri-o. This PR unifies default registries so that both supported runtimes use docker.io as the default registry (if non-qualified image repo is used). This should fix conformance problems with cri-o.


lib/pharos/host/el7/scripts/configure-cri-o.sh
```diff
@@ -77,6 +77,7 @@ lineinfile "^stream_address =" "stream_address = \"${CRIO_STREAM_ADDRESS}\"" "/e
 lineinfile "^cgroup_manager =" "cgroup_manager = \"systemd\"" "/etc/crio/crio.conf"
 lineinfile "^log_size_max =" "log_size_max = 134217728" "/etc/crio/crio.conf"
 lineinfile "^pause_image =" "pause_image = \"${IMAGE_REPO}/pause-${CPU_ARCH}:3.1\"" "/etc/crio/crio.conf"
+lineinfile "^registries =" "registries = [ \"docker.io\"" "/etc/crio/crio.conf"
```
nevalla: should there be closing `]` somewhere?

jakolehm: original file is like this:
```
registries = [
]
```

PR #499, opened 2018-08-03
[wip] Specify cluster name, username and kubeconfig path in cluster.yml

### LINKED ISSUE #438: Possibility to set kubeconfig path + merge if existing config
```
kube_config: 
  path: ~/.kube/config
  context: foo
```

kke:
Fixes #438

- Set cluster, user and context names in cluster.yml
- Set path for kubectl config. If file exists, the new config will be merged on top of it

Example config:

```yaml
# cluster.yml
---
cluster:
  name: pharos-cluster # default comes from kube and is 'kubernetes'
  kube_config:
    path: ~/.kube/config # default is ~/.pharos/ip.address.of.apiserver
    user: admin # default comes from kube and is 'kubernetes-admin'
    context: root # default comes from kube and is 'kubernetes-admin@kubernetes'
```

### DIFF
--- a/cluster.example.yml
+++ b/cluster.example.yml
@@ -60,3 +60,9 @@ addons:
     interval: 7d
   kured:
     enabled: true
+cluster:
+  name: pharos-cluster
+  kube_config:
+    path: ~/.kube/config
+    user: admin
+    context: admin@pharos-cluster
--- a/lib/pharos/phases/configure_client.rb
+++ b/lib/pharos/phases/configure_client.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'pharos/kube/config'
+
 module Pharos
   module Phases
     class ConfigureClient < Pharos::Phase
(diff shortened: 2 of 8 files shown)


cluster.example.yml
```diff
@@ -60,3 +60,9 @@ addons:
     interval: 7d
   kured:
     enabled: true
+cluster:
```
jakolehm: I think this is unnecessary padding (the file itself is a cluster level information). Maybe we could just have `name` and `kubeconfig` as top level properties.

lib/pharos/phases/configure_client.rb
```diff
@@ -20,9 +22,18 @@ def save_config_locally
         Dir.mkdir(config_dir, 0o700) unless Dir.exist?(config_dir)
```
jakolehm: We don't need this anymore, right?

jakolehm: Or `config_dir` should come from `config_file`.

lib/pharos/up_command.rb
```diff
@@ -108,7 +108,11 @@ def configure(config)
       craft_time = Time.now - start_time
       puts pastel.green("==> Cluster has been crafted! (took #{humanize_duration(craft_time.to_i)})")
       puts "    You can connect to the cluster with kubectl using:"
-      puts "    export KUBECONFIG=~/.pharos/#{manager.sorted_master_hosts.first.api_address}"
+      if config.cluster&.kube_config&.path
+        puts "    export KUBECONFIG=#{config.cluster.kube_config.path}"
+      else
```
jakolehm: IMHO this does not make sense, we should just default to the old config path if user has not defined path in cluster.yml. Would make code easier to read (no if/else)

kke:
Replaced by #502

PR #502, opened 2018-08-07
Add pharos-cluster kubeconfig -command

### LINKED ISSUE #438: Possibility to set kubeconfig path + merge if existing config

kke:
Alternative to #499
Fixes #438 

```
$ pharos-cluster kubeconfig # dumps config from master host defined in cluster.yml to stdout
$ pharos-cluster kubeconfig -n foofoo # renames the server in the config to foofoo
$ pharos-cluster kubeconfig -C foofoo # renames the context in the config to foofoo
```


lib/pharos/kubeconfig_command.rb
```diff
@@ -0,0 +1,100 @@
...
+        File.realpath(config_path)
+      rescue Errno::ENOENT
+        signal_usage_error 'File does not exist: %<path>s' % { path: config_path }
+      end
+    end
+
+    option ['-n', '--name'], 'NAME', 'Overwrite cluster name', attribute_name: :new_name
+    option ['-C', '--context'], 'CONTEXT', 'Overwrite context name', attribute_name: :new_context
+
+    option ['-i', '--input'], '[FILE]', 'Use local file as input instead of remote'
```
jakolehm: `--input` is a bit confusing.. probably we don't need it because merging local files can be done with `kubectl`?

lib/pharos/kubeconfig_command.rb
```diff
@@ -0,0 +1,100 @@
...
+        merge_config = Pharos::Kube::Config.new(File.read(merge))
+        config << merge_config
+      end
+      puts config
+    end
+
+    private
+
+    # @return [Pharos::YamlFile]
+    def cluster_config
```
jakolehm: Config handling is duplicate code from `up` command.

lib/pharos/root_command.rb
```diff
@@ -3,14 +3,16 @@
...
+require_relative 'kubeconfig_command'
 
 module Pharos
   class RootCommand < Pharos::Command
     banner "pharos-cluster - Kontena Pharos cluster manager"
 
     subcommand ["build", "up"], "Initialize/upgrade cluster", UpCommand
-    subcommand ["reset"], "Reset cluster", ResetCommand
-    subcommand ["version"], "Show version information", VersionCommand
+    subcommand "kubeconfig", "Download kubernetes configuration files", KubeconfigCommand
```
jakolehm: "Fetch admin kubeconfig file" (outputs, not downloads + only one file)

jakolehm:
Not sure if this is really direct alternative to #499 ... because user should be able to opt-out from automatic kubeconfig in `~/.pharos/xyz`. We could stop saving that file but we still need a kubeconfig for accessing the k8s api (in-memory?).

kke:
It could always skip saving it and instruct to use `pharos-cluster kubeconfig`. It's already used from in-memory (`cluster_context['previous_config']`) after the `ConfigureClient` phase.

jakolehm:
> It could always skip saving it and instruct to use pharos-cluster kubeconfig. It's already used from in-memory (cluster_context['previous_config']) after the ConfigureClient phase.

Always skip saving makes sense to me. @jnummelin @nevalla @SpComb wdyt?

kke:
Also possible: some kind of 

```
$ pharos-cluster kubectl -c cluster.yml  -- run hello-kube --image=k8s.gcr.io/echoserver:1.10 --port=8080
```

that would use `KubeConfigCommand` to create a tempfile, set `KUBECONFIG`, run kubectl and remove the tempfile.

Or 

```
$ pharos-cluster kubeconfig --env
KUBECONFIG=/tmp/pharos123345423vdfvfdvfdfvd.admin.conf
$ eval $(pharos-cluster kubeconfig --env)
```

Or

```
$ pharos-cluster kubeconfig --temp-file
/tmp/pharos123345423vdfvfdvfdfvd.admin.conf
$ export KUBECONFIG=$(pharos-cluster kubeconfig --temp-file)
```

lib/pharos/kubeconfig_command.rb
```diff
@@ -0,0 +1,49 @@
...
+    def config_file_content
+      file = ssh.file(REMOTE_FILE)
+      signal_usage_error "Remote file #{REMOTE_FILE} not found" unless file.exist?
+      file.read
+    end
+
+    def ssh
+      return @ssh if @ssh
+      opts = {}
+      opts[:keys] = [master_host.ssh_key_path] if master_host.ssh_key_path
```
SpComb: Don't duplicate this code, use `Pharos::SSH::Manager.client_for(master_host)`

lib/pharos/up_command.rb
```diff
@@ -111,7 +111,8 @@ def configure(config)
       craft_time = Time.now - start_time
       puts pastel.green("==> Cluster has been crafted! (took #{humanize_duration(craft_time.to_i)})")
       puts "    You can connect to the cluster with kubectl using:"
-      puts "    export KUBECONFIG=~/.pharos/#{manager.sorted_master_hosts.first.api_address}"
+      puts "      pharos-cluster kubeconfig > ~/.pharos/admin.conf"
```
jakolehm: Maybe `pharos-cluster` -> `$0` ?

SpComb:
I don't understand this PR, it doesn't write the `~/.pharos/*` kubeconfigs anymore, but the `Pharos::Kube` still needs those?

```
==> Configure kube client @ terom-pharos-master
No such file or directory @ rb_sysopen - /root/.pharos/master.terom-pharos-dev.kontena.works
/usr/local/lib/ruby/2.4.0/psych.rb:472:in `initialize'
/usr/local/lib/ruby/2.4.0/psych.rb:472:in `open'
/usr/local/lib/ruby/2.4.0/psych.rb:472:in `load_file'
/usr/local/bundle/gems/k8s-client-0.3.1/lib/k8s/config.rb:81:in `load_file'
/app/lib/pharos/kube.rb:58:in `host_config'
/app/lib/pharos/kube.rb:45:in `client'
/app/lib/pharos/phase.rb:83:in `kube_client'
/app/lib/pharos/phases/configure_client.rb:19:in `client_prefetch'
/app/lib/pharos/phases/configure_client.rb:9:in `call'
/app/lib/pharos/phase_manager.rb:92:in `block in apply'
/app/lib/pharos/phase_manager.rb:53:in `yield_phase_with_retry'
/app/lib/pharos/phase_manager.rb:44:in `block in run_serial'
/app/lib/pharos/phase_manager.rb:43:in `map'
/app/lib/pharos/phase_manager.rb:43:in `run_serial'
/app/lib/pharos/phase_manager.rb:73:in `run'
/app/lib/pharos/phase_manager.rb:89:in `apply'
/app/lib/pharos/cluster_manager.rb:119:in `apply_phase'
/app/lib/pharos/cluster_manager.rb:93:in `apply_phases'
/app/lib/pharos/up_command.rb:104:in `configure'
/app/lib/pharos/up_command.rb:45:in `block in execute'
/app/lib/pharos/up_command.rb:44:in `chdir'
/app/lib/pharos/up_command.rb:44:in `execute'
/app/lib/pharos/root_command.rb:18:in `run'
```

lib/pharos/up_command.rb
jakolehm: And `~/.pharos/admin.conf` -> `admin.conf` .

jakolehm: And example probably should include same options that `up` had?

lib/pharos/phases/configure_client.rb
```diff
@@ -17,28 +14,6 @@ def call
         cluster_context['previous-config'] = build_config(config_map)
       end
 
-      def save_config_locally
```
jakolehm: How do we load config for kube client now?

lib/pharos/kube/config.rb
```diff
@@ -0,0 +1,124 @@
+# frozen_string_literal: true
+
+require 'pharos/yaml_file'
+
+module Pharos
+  module Kube
+    # Loads and manipulates kubeconfig files
+    class Config
```
SpComb: BTW there's also a `K8s::Config` now, but it's not necessarily relevant here... because it's based on `Dry::Struct`, that also means that it's read-only... it's not really designed to write or modify configs.

lib/pharos/phases/configure_client.rb
kke: I think I had one and it worked because of that.

lib/pharos/kube/config.rb
kke: K8s yardocs have `@param config [Pharos::Kube::Config]` in several places

jnummelin:
> Always skip saving makes sense to me.

Users have been depending that the configs are saved, so maybe not a good idea to just drop it "silently". Maybe put it behind a flag that would still default to saving for a while, or something.

lib/pharos/kubeconfig_command.rb
```diff
@@ -0,0 +1,47 @@
...
+
+    REMOTE_FILE = "/etc/kubernetes/admin.conf"
+
+    def execute
+      config = Pharos::Kube::Config.new(config_file_content)
+      config.rename_cluster(new_name) if new_name
+      config.rename_context(new_context) if new_context
+      config.update_server_address(master_host.api_address)
+      merge_list.each do |merge|
+        merge_config = Pharos::Kube::Config.new(File.read(merge))
```
jnummelin: IMO it would make sense to also somehow support `KUBECONFIG` env, or the default location `~/.kube/config`.

lib/pharos/up_command.rb
```diff
@@ -110,8 +110,8 @@ def configure(config)
 
       craft_time = Time.now - start_time
       puts pastel.green("==> Cluster has been crafted! (took #{humanize_duration(craft_time.to_i)})")
-      puts "    You can connect to the cluster with kubectl using:"
-      puts "    export KUBECONFIG=~/.pharos/#{manager.sorted_master_hosts.first.api_address}"
+      puts "    To configure kubectl for connecting to the cluster, use:"
+      puts "      #{$PROGRAM_NAME} kubeconfig > ~/.kube/config"
```
jakolehm: I think we still should show the `export KUBECONFIG` hint.

jakolehm:
> Users have been depending that the configs are saved, so maybe not a good idea to just drop it "silently".

We need to document this properly in the release notes. If someone is using locally saved config in automation it will most likely break things... but it should be easy to fix by adding an additional `pharos-cluster kubeconfig` command.

lib/pharos/up_command.rb
jakolehm: `~/.kube/config` is not a good idea as an example because it will overwrite local kubeconfig that most likely is already there. Or did I miss something?

lib/pharos/kubeconfig_command.rb
jnummelin: I mean support those in the merging

lib/pharos/up_command.rb
jnummelin: yeah, I agree that it's a bad example

Issue #501, opened 2018-08-07
 undefined method `[]' for nil:NilClass (NoMethodError)


antonyveyre:
Launching pharos from Mac os to ubuntu on vmware give this issue. 

Traceback :
```
Traceback (most recent call last):
	6: from /pharos-cluster/lib/pharos/phase_manager.rb:26:in `block (2 levels) in run_parallel'
	5: from /pharos-cluster/lib/pharos/phase_manager.rb:70:in `block in apply'
	4: from /pharos-cluster/lib/pharos/phases/gather_facts.rb:14:in `call'
	3: from /pharos-cluster/lib/pharos/phases/gather_facts.rb:25:in `gather_host_facts'
	2: from /pharos-cluster/lib/pharos/phases/gather_facts.rb:60:in `cpu_arch'
	1: from /pharos-cluster/lib/pharos/phases/gather_facts.rb:60:in `each'
/pharos-cluster/lib/pharos/phases/gather_facts.rb:62:in `block in cpu_arch': undefined method `[]' for nil:NilClass (NoMethodError)
```

Found solution: 
```
diff --git a/lib/pharos/phases/gather_facts.rb b/lib/pharos/phases/gather_facts.rb
index ccb7e67..8162a0e 100644
--- a/lib/pharos/phases/gather_facts.rb
+++ b/lib/pharos/phases/gather_facts.rb
@@ -56,13 +56,9 @@ module Pharos

       # @return [Pharos::Configuration::CpuArch]
       def cpu_arch
-        cpu = {}
-        @ssh.exec!('lscpu').split("\n").each do |line|
-          match = line.match(/^(.+):\s+(.+)$/)
-          cpu[match[1]] = match[2]
-        end
+        arch = @ssh.exec!('uname -m')
         Pharos::Configuration::CpuArch.new(
-          id: cpu['Architecture']
+          id: arch.tap{|x| x.strip!}
         )
       end
```

SpComb:
Can you show what `lscpu` on the Ubuntu host outputs?

Agreed that the loop should skip mismatching lines, but `uname -m` would be better anyways.

antonyveyre:
ubuntu@master0:~$ uname -m
x86_64

ubuntu@master0:~$ lscpu
Architecture : x86_64
Mode(s) opératoire(s) des processeurs :32-bit, 64-bit
Boutisme : Little Endian
Processeur(s) : 2
Liste de processeur(s) en ligne :0,1
[...]

PR #504, opened 2018-08-07
Fixup #455 for newer dry-struct

SpComb:
#455 was written against git master before the `Dry::Struct` upgrade in #474, and is broken with the newer dry-struct:

```
undefined method `constructor_type' for Pharos::Configuration::Host::Route:Class
Did you mean?  constructor
/app/lib/pharos/configuration/host.rb:28:in `<class:Route>'
/app/lib/pharos/configuration/host.rb:14:in `<class:Host>'
/app/lib/pharos/configuration/host.rb:8:in `<module:Configuration>'
/app/lib/pharos/configuration/host.rb:7:in `<module:Pharos>'
/app/lib/pharos/configuration/host.rb:6:in `<top (required)>'
/app/lib/pharos/config.rb:4:in `require_relative'
/app/lib/pharos/config.rb:4:in `<top (required)>'
/usr/local/lib/ruby/site_ruby/2.4.0/rubygems/core_ext/kernel_require.rb:59:in `require'
/usr/local/lib/ruby/site_ruby/2.4.0/rubygems/core_ext/kernel_require.rb:59:in `require'
/app/lib/pharos/up_command.rb:59:in `load_config'
/app/lib/pharos/up_command.rb:36:in `execute'
/usr/local/bundle/gems/clamp-1.2.1/lib/clamp/command.rb:63:in `run'
/usr/local/bundle/gems/clamp-1.2.1/lib/clamp/subcommand/execution.rb:11:in `execute'
/usr/local/bundle/gems/clamp-1.2.1/lib/clamp/command.rb:63:in `run'
/usr/local/bundle/gems/clamp-1.2.1/lib/clamp/command.rb:132:in `run'
/app/lib/pharos/root_command.rb:16:in `run'
bin/pharos-cluster:12:in `<main>'
```

PR #505, opened 2018-08-07
Change lscpu to uname -m

antonyveyre:
fix for https://github.com/kontena/pharos-cluster/issues/501


SpComb:
Drone is :x: because master is broken pending #504, otherwise LGTM

PR #509, opened 2018-08-08
Generate sha256sums on release and add to github release page

kke:
Generate `pharos-cluster-linux-amd64.sha256` and `pharos-cluster-darwin-amd64.sha256` and upload them to github release as attachments.

`xargs -P0` should make the uploads happen in parallel, usually meaning that the creation timestamp that is visible in the [release info JSON](https://api.github.com/repos/kontena/pharos-cluster/releases/latest)'s asset list should be very similar.


kke:
If someone switches the binary, they can just as well switch the `.sha256`. Gpg would be more trustworthy I guess.

SpComb:
Convention would be to have a single `SHA256SUMS` file with checksums for all of the release files.

> xargs -P0 should make the uploads happen in parallel, usually meaning that the creation timestamp that is visible in the release info JSON's asset list should be very similar.

Nothing wrong with the `xargs -P0`, but a `for ... in ...; do ...` loop would be fine as well, I wouldn't worry too much about the timestamps.

kke:
Outdated due to binaries no longer being uploaded  to github

PR #512, opened 2018-08-08
Retry phases (with exp backoff) on errors

jakolehm:


lib/pharos/phase_manager.rb
```diff
@@ -19,25 +19,37 @@ def initialize(ssh_manager:, **options)
...
+      end
+    end
+
+    # @param phase [Pharos::Phases::Base]
+    # @param retry_times [Integer]
+    def yield_phase_with_retry(phase, retry_times = 10)
+      retries = 0
+      begin
         yield phase
+      rescue StandardError => exc
```
SpComb: `StandardError` also includes things like ruby `NameError`... I don't think it's a good idea to retry all phases on any error like that, there should be an explicit per-phase list of errors to retry on... something like

```ruby
class ConfigureWeave < Pharos::Phase
    retry_error K8s::Error
    
    ...
end
```

jakolehm: Idk if explicit list per-phase is a good idea.. because then almost every phase needs to list OpenSSL etc errors?

jakolehm: Maybe fixed list of more specific errors?

jakolehm: Now it retries only on specific errors (kube, openssl, ssh).

lib/pharos/phase_manager.rb
```diff
@@ -4,6 +4,13 @@ module Pharos
   class PhaseManager
     include Pharos::Logging
 
+    RETRY_ERRORS = [
+      OpenSSL::SSL::SSLError,
+      Kubeclient::HttpError,
+      Net::SSH::Disconnect,
```
SpComb: I doubt that this retry loop by itself would be enough to reconnect on `Net::SSH` errors.

jakolehm: True, I'll remove `Net::SSH::*`.

lib/pharos/phase_manager.rb
```diff
@@ -4,6 +4,13 @@ module Pharos
...
+      Net::SSH::ConnectionTimeout
+    ].freeze
```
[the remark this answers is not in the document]
SpComb: While that would fix issues like random dpkg locking errors (`E: Unable to lock the administration directory (/var/lib/dpkg/) is another process using it?`), it seems a bit too broad to me... I would only retry errors that are explicitly known to be retryable.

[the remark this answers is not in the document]
jakolehm: > it seems a bit too broad to me.

I kinda disagree with this... if user has only option to re-run `up` then we should just retry these errors automatically.

lib/pharos/phase_manager.rb
```diff
@@ -19,25 +25,37 @@ def initialize(ssh_manager:, **options)
...
+    # @param phase [Pharos::Phases::Base]
+    # @param retry_times [Integer]
+    def yield_phase_with_retry(phase, retry_times = 10)
+      retries = 0
+      begin
         yield phase
+      rescue *RETRY_ERRORS => exc
+        raise if retries >= retry_times
+
+        logger.error { "[#{phase.host}] got error (#{exc.class.name}) #{exc.message}" }
```
SpComb: `logger.debug{ "#{exc.backtrace.join("\n")}" }` would be a good idea for debugging errors during development.

jakolehm: Debug backtrace is now there.

jakolehm:
Rebased & fixed for K8s::Client errors.

jakolehm:
Rebased again. Retry with the new k8s-client seems to work:

```
==> Configure kube client @ 188.166.49.177
    [188.166.49.177] Fetching kubectl config ...
    [188.166.49.177] Configuration saved to /Users/jari/.pharos/188.166.49.177
    [188.166.49.177] got error (K8s::Error::NotFound): GET /apis/metrics.k8s.io/v1beta1 => HTTP 404 Not Found: 404 page not found
    [188.166.49.177] retrying after 1 seconds ...
    [188.166.49.177] Fetching kubectl config ...
    [188.166.49.177] Configuration saved to /Users/jari/.pharos/188.166.49.177
    [188.166.49.177] got error (K8s::Error::NotFound): GET /apis/metrics.k8s.io/v1beta1 => HTTP 404 Not Found: 404 page not found
    [188.166.49.177] retrying after 2 seconds ...
    [188.166.49.177] Fetching kubectl config ...
    [188.166.49.177] Configuration saved to /Users/jari/.pharos/188.166.49.177
```

PR #475, opened 2018-07-13
New K8s::Client

SpComb:
Depends on #474
Fixes #227

Replace the `kubeclient` gem with the new [`k8s-client`](https://github.com/kontena/k8s-client) gem.

This has a better API (subjective), and uses excon HTTP request keepalive/pipelining for *significantly* (10x) better API discovery and  `Pharos::Kube::Stack#apply` performance.

PR #520, opened 2018-08-10
Fix Pharos::Kube::Stack label/annotation override

SpComb:
#475 didn't actually override the label used for `Pharos::Kube::Stack`s due to how the `LABEL` constant is used as a default value in the superclass initialize method.  This caused existing stack resources to gain an additional `k8s.kontena.io/stack` label, and new resources would be missing the `pharos.kontena.io/stack` label.

Changing the stack label is harmless, because existing resources are found by name during the apply, however, it would better to keep it the same in case any stack resources are meant to be purged during the ugprade.

### DIFF
--- a/lib/pharos/kube.rb
+++ b/lib/pharos/kube.rb
@@ -32,6 +32,10 @@ module Pharos
 
         new(name, resources)
       end
+
+      def initialize(name, resources = [])
+        super(name, resources, label: LABEL, checksum_annotation: CHECKSUM_ANNOTATION)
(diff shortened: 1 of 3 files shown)


SpComb:
Added specs, but unfortunately that requires mocking of the `K8s::Stack` => `K8s::Client` bits, because the `K8s::Stack` API doesn't actually expose the labeled resources. This would be more suitable as an end-to-end integration test against an actual kube blackbox...

PR #522, opened 2018-08-10
Load libraries in phase manager to fix version command

### LINKED ISSUE #521: pharos-cluster version throws "uninitialized constant"
```
pharos-cluster version
Kontena Pharos:
  - pharos-cluster version 1.3.0-beta.2
uninitialized constant Pharos::PhaseManager::OpenSSL
```

kke:
Fixes #521

Adds `require`s to phase manager so that it can define the list of retriable errors.

### DIFF
--- a/lib/pharos/phase_manager.rb
+++ b/lib/pharos/phase_manager.rb
@@ -1,9 +1,6 @@
 # frozen_string_literal: true
 
-require 'openssl'
-require 'excon'
-require 'k8s/error'
-require 'pharos/ssh/remote_command'
+require_relative 'kube'
 
 module Pharos
   class PhaseManager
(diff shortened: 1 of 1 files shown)


lib/pharos/phase_manager.rb
```diff
@@ -1,5 +1,10 @@
 # frozen_string_literal: true
 
+require 'openssl'
+require 'excon'
+require 'k8s/error'
+require 'pharos/ssh/remote_command'
```
jakolehm: or just `require_relative 'kube'`?

jakolehm: reason: openssl & excon are `k8s-client` deps

PR #530, opened 2018-08-14
Lockdown crio stream-address to localhost-only for kubelet proxying

### LINKED ISSUE #508: Cri-o should use --redirect-container-streaming=false
For added security (similar to dockershim).

See https://github.com/kubernetes/kubernetes/pull/64006 .

SpComb:
Fixes #508

Replace the host public/private IP (reachable by the kube-apiserver for redirects) with `127.0.0.1`, for kubelet CRI proxying.

Tested that `kubectl logs` and `kubectl exec` still works.

### DIFF
--- a/lib/pharos/host/el7/el7.rb
+++ b/lib/pharos/host/el7/el7.rb
@@ -52,7 +52,7 @@ module Pharos
           exec_script(
             'configure-cri-o.sh',
             CRIO_VERSION: Pharos::CRIO_VERSION,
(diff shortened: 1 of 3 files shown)


SpComb:
Upstream change (for what release version?) is to listen on a randomly chosen localhost-only port: https://github.com/kubernetes-incubator/cri-o/pull/1714

This PR keeps the fixed 10010 port.

PR #546, opened 2018-08-17
Improve kubeconfig hint

### LINKED ISSUE #541: kubeconfig requires too much stuff
Which makes whole cli slow-ish. It should rely on autoload.

jakolehm:
`$HOME/.pharos/config` is confusing if you have multiple clusters. Better to instruct user to save the config to the same path where the command is executed (usually the same folder that has cluster.yml).

Fixes #541

### DIFF
--- a/lib/pharos/up_command.rb
+++ b/lib/pharos/up_command.rb
@@ -112,8 +112,8 @@ module Pharos
       craft_time = Time.now - start_time
       puts pastel.green("==> Cluster has been crafted! (took #{humanize_duration(craft_time.to_i)})")
       puts "    To configure kubectl for connecting to the cluster, use:"
-      puts "      #{$PROGRAM_NAME} kubeconfig > $HOME/.pharos/config"
-      puts "      export KUBECONFIG=$KUBECONFIG:$HOME/.pharos/config"
+      puts "      #{$PROGRAM_NAME} kubeconfig > ./kubeconfig"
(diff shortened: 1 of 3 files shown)


lib/pharos/up_command.rb
```diff
@@ -112,8 +112,8 @@ def configure(config)
       craft_time = Time.now - start_time
       puts pastel.green("==> Cluster has been crafted! (took #{humanize_duration(craft_time.to_i)})")
       puts "    To configure kubectl for connecting to the cluster, use:"
-      puts "      #{$PROGRAM_NAME} kubeconfig > $HOME/.pharos/config"
-      puts "      export KUBECONFIG=$KUBECONFIG:$HOME/.pharos/config"
+      puts "      #{$PROGRAM_NAME} kubeconfig > kubeconfig"
+      puts "      export KUBECONFIG=./kubeconfig"
```
kke: That will only work as long as the user is in that directory and after that may cause connection to some other cluster. Maybe `$PWD/kubeconfig` to both?

jakolehm: That said, it's also missing `-c` and `--tf-json` options.. idk how to fetch them from clamp (only if they are defined).

kke: `ARGV[1..-1]` seems to work

jakolehm: Fixed.

PR #548, opened 2018-08-17
Fix kube client config

### LINKED ISSUE #534: 1.2 -> 1.3 ugprade assumes that ~/.pharos/* kubeconfigs already exist
The `UpgradeMaster` phase uses the `kube_client` for the `dns_patch_thread`: https://github.com/kontena/pharos-cluster/blob/master/lib/pharos/phases/upgrade_master.rb#L53

This assumes that the `~/.pharos/*` kubeconfig for the master already exists prior to running `pharos-cluster up`, because the `ConfigureClient` phase that writes that file only runs after the `UpgradeMaster` phase.

The upgrade will fail if the `pharos-cluster up` command is run on a different host than what was used to `pharos-cluster up` the cluster before the upgrade.

SpComb:
Fixes #534 by calling `Phases::ConfigureClient` twice, allowing an optional kube config used for upgrades

* Refactoring to fix the usage of `Pharos::Kube.client` for both phases and addons, removing the broken fallback to `~/.pharos/*`
* Split `LoadClusterConfiguration` out of `ConfigureClient`
* Run `ConfigureClient` twice, with an optional early call for upgrades

### DIFF
--- a/lib/pharos/phase.rb
+++ b/lib/pharos/phase.rb
@@ -79,6 +79,7 @@ module Pharos
     # @return [K8s::Client]
     def kube_client
       fail "Phase #{self.class.name} does not have kube @master" unless @master
+      fail "Phase #{self.class.name} does not have kubeconfig cluster_context" unless cluster_context['kubeconfig']
 
       @kube_client ||= Pharos::Kube.client(@master.api_address, cluster_context['kubeconfig'])
     end
--- a/lib/pharos/addon.rb
+++ b/lib/pharos/addon.rb
@@ -125,17 +125,17 @@ module Pharos
       end
     end
 
-    attr_reader :config, :cpu_arch, :cluster_config, :master
+    attr_reader :config, :cpu_arch, :cluster_config, :kube_client
 
     # @param config [Hash,Dry::Validation::Result]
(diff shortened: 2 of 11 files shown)


lib/pharos/phase.rb
```diff
@@ -79,6 +79,7 @@ def host_configurer
     # @return [K8s::Client]
     def kube_client
       fail "Phase #{self.class.name} does not have kube @master" unless @master
+      fail "Phase #{self.class.name} does not have kubeconfig cluster_context" unless cluster_context['kubeconfig']
 
       @kube_client ||= Pharos::Kube.client(@master.api_address, cluster_context['kubeconfig'])
```
SpComb: Note that the value for `@master.api_address` being passed here is completely irrelevant now, it could be anything at all.

SpComb: The refactoring here could be taken a step further to replace the `master` param in the phases with a `kube_client`, but I'm not sure I want to do that in the same PR.

For the `Pharos::Addon`, it happened here because the class would have needed a new `cluster_context` or `kube_config` param as well, and it made more sense to just go all the way and pass the `kube_client` instead.

PR #554, opened 2018-08-20
Add chpharos token to telemetry

jakolehm:


jakolehm:
@nevalla PTAL

lib/pharos/resources/telemetry/06-secret.yml.erb
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: telemetry-token
+  namespace: kube-system
+data:
+  token: "<%= token %>"
```
jnummelin: `token` here vs `customer_token` in `ConfigureTelemetry `??

PR #557, opened 2018-08-20
Initial Network LB addon implementation

jnummelin:
Bit of WIP still.

Initial go at setting up [MetalLB](https://metallb.universe.tf/) via addon.

Need to still figure out how to properly support possibly per node node selectors for peers. The current way is bit crude. :) Then again, not sure if it would make sense to model all the possibilities of node selectors as dry-schema. :/ 50-60 as the (in)famous ski-jumper said...

To set this up for `layer2` would be:
```
metal-lb:
    enabled: true
    address_pools:
      - name: default
        protocol: layer2
        addresses:
          - 147.75.84.224/27
```

For BGP respectively:
```
metal-lb:
    enabled: true
    address_pools:
      - name: default
        protocol: bgp
        addresses:
          - 147.75.84.224/27
    peers:
      - peer_address: "10.80.119.128"
        peer_asn: "65530"
        my_asn: "65000"
        node_selectors:
          - match-expressions:
            - key: kubernetes.io/hostname
              operator: Equals
              values: pharos-worker-0
      - peer_address: "10.80.119.130"
        peer_asn: "65530"
        my_asn: "65000"
        node_selectors:
          - match-expressions:
            - key: kubernetes.io/hostname
              operator: Equals
              values: pharos-worker-1
```
In this case the config is bit complex as the example is from Packet infra where all nodes need different BGP peering config. Hence there's a peer config for each host. In environment where there's single router available one could configure only single peer. (as in MetalLB basic config examples)


addons/metallb/addon.rb
```diff
@@ -0,0 +1,80 @@
...
+        config: {
+          # This part is using string keys to get the output yaml correct as it's used as plain string
+          'address-pools' => config.address_pools.map { |pool|
+            {
+              'name' => pool[:name],
+              'protocol' => pool[:protocol],
+              'addresses' => pool[:addresses]
+            }
+          },
+          'peers' => config.peers.map { |peer|
```
Timer: This should only be present if `config.address_pools.count { |pool| pool.dig(:protocol) == 'bgp' }.positive?`.

addons/metallb/resources/20-speaker-daemonset.yml
```diff
@@ -0,0 +1,53 @@
...
+      annotations:
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "7472"
+    spec:
+      serviceAccountName: speaker
+      terminationGracePeriodSeconds: 0
+      hostNetwork: true
+      containers:
+      - name: speaker
+        image: metallb/speaker:master
```
Timer: This should probably be pinned to the addon version. 😄

addons/metallb/resources/21-controller-deployment.yml
```diff
@@ -0,0 +1,49 @@
...
+        prometheus.io/port: "7472"
+    spec:
+      serviceAccountName: controller
+      terminationGracePeriodSeconds: 0
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534 # nobody
+      containers:
+      - name: controller
+        image: metallb/controller:master
```
Timer: Ditto.

addons/metallb/resources/20-speaker-daemonset.yml
```diff
@@ -0,0 +1,53 @@
+apiVersion: apps/v1beta2
+kind: DaemonSet
+metadata:
+  namespace: metallb-system
+  name: speaker
```
Timer: The ability to specify `tolerations` would be much appreciated.

addons/metallb/resources/20-speaker-daemonset.yml
jakolehm: Also image should point to the pharos registry.

jnummelin: done

addons/metallb/resources/20-speaker-daemonset.yml
jnummelin: @Timer So you'd want only specific nodes to be able to publish the routes? I _think_  you can do the same by targeting the peer configs using proper `node_selectors`.

jnummelin:
There's now support for adding both `nodeSelector` and `tolerations` for the speaker daemonset.

non-oss/pharos_pro/addons/kontena-network-lb/addon.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+Pharos.addon 'kontena-network-lb' do
+  version '0.7.3'
+  license 'Apache License 2.0'
```
jakolehm: Should be "Kontena License".

jnummelin: fixed

PR #558, opened 2018-08-21
Added optional secret name for private registry

wolfedale:
When someone is using private registry (eg, gitlab) we need to create secret so we can execute pull on the specific ingres-nginx-backend image.

### DIFF
--- a/addons/ingress-nginx/addon.rb
+++ b/addons/ingress-nginx/addon.rb
@@ -15,6 +15,7 @@ Pharos.addon 'ingress-nginx' do
     optional(:node_selector).filled(:hash?)
     optional(:default_backend).schema {
       optional(:image).filled(:str?)
+      optional(:pull_secret_name).filled(:str?)
     }
   }
 
@@ -48,4 +49,5 @@ Pharos.addon 'ingress-nginx' do
     return 2 if r < 2
     r
   end
+
 end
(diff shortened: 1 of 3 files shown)


addons/ingress-nginx/addon.rb
```diff
@@ -15,6 +15,7 @@
     optional(:node_selector).filled(:hash?)
     optional(:default_backend).schema {
       optional(:image).filled(:str?)
+      optional(:secret_name).filled(:str?)
```
jnummelin: IMO the name could be something bit more descriptive. In this case the secret is being used as `imagePullSecret` hence I'd prefer the name to also reflect it's usage.

addons/ingress-nginx/addon.rb
```diff
@@ -48,4 +50,8 @@ def default_backend_replicas
     return 2 if r < 2
     r
   end
+
+  def secret_name
```
jnummelin: I don't think you need a specific method for this. As the attribute is marked as optional in the schema, if it's not provided it will be `nil` anyway. Without this method, you could use the value as `config.default_backend&.dig(:secret_name)` in the `apply_resources` vars

kke:
Should this be closed?

jnummelin:
@wolfedale Please re-open if you still feel this is needed. At least CLA needs to be signed and the conflicts resolved before we can take it further.

PR #563, opened 2018-08-23
Drop the Pharos.addon DSL in favor of regular subclassing

### LINKED ISSUE #562: Add-on constants are not namespaced

```
Pharos.addon "foo" do
  PATH = "/etc/config"
end

Pharos.addon "bar" do
  PATH = "/etc/bar"
end

# PATH => "/etc/bar"
```

kke:
Fixes #562 

To avoid dynamic constant assignment and overlapping constants, the block format `Pharos.addon "name" do` definition syntax has now been removed.

Any add-ons must now be defined using a regular class syntax:

```diff
- Pharos.addon "addon-name" do
+ class AddonName < Pharos::Addon
```

Full example:

```ruby
# kured/addon.rb
class KuredAddon < Pharos::Addon
  addon_name 'kured' # by default generated by dash-casing the class name
  version 'master-b27aaa1'
  license 'Apache License 2.0'

  install do
    apply_resources
  end
end
```


kke:
Outdated

PR #564, opened 2018-08-23
Do not allow addons to define consts

### LINKED ISSUE #562: Add-on constants are not namespaced

SpComb:
Fixes #562
Alternative to #563

Only two of the core addons use consts in a way that is trivial to refactor.

Hack in validation to detect if any addon defines global consts when loading:

### DIFF
--- a/addons/ingress-nginx/addon.rb
+++ b/addons/ingress-nginx/addon.rb
@@ -27,16 +27,13 @@ Pharos.addon 'ingress-nginx' do
     )
   }
 
-  DEFAULT_BACKEND_ARM64_IMAGE = 'docker.io/kontena/pharos-default-backend-arm64:0.0.2'
-  DEFAULT_BACKEND_IMAGE = 'docker.io/kontena/pharos-default-backend:0.0.2'
-
   def image_name
     return config.default_backend[:image] if config.default_backend&.dig(:image)
 
-    if cpu_arch.name == 'arm64'
-      DEFAULT_BACKEND_ARM64_IMAGE
+    if cpu_arch.name == 'amd64'
+      'docker.io/kontena/pharos-default-backend:0.0.2'
     else
-      DEFAULT_BACKEND_IMAGE
+      "docker.io/kontena/pharos-default-backend-#{cpu_arch.name}:0.0.2"
     end
   end
 
--- a/addons/openebs/addon.rb
+++ b/addons/openebs/addon.rb
@@ -4,20 +4,14 @@ Pharos.addon 'openebs' do
   version '0.5.3'
   license 'Apache License 2.0'
 
-  DEFAULT_CLASS_OPTS = {
-    default_class: false,
-    capacity: '5G'
-  }.freeze
-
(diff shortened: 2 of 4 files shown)


SpComb:
Testing with `ingress-nginx`:

```patch
--- a/addons/ingress-nginx/addon.rb
+++ b/addons/ingress-nginx/addon.rb
@@ -4,6 +4,8 @@ Pharos.addon 'ingress-nginx' do
   version '0.17.1'
   license 'Apache License 2.0'
 
+  FOO = 'bar'
+
   config {
     attribute :configmap, Pharos::Types::Hash
     attribute :node_selector, Pharos::Types::Hash
```

```
$ pharos-cluster version
Kontena Pharos:
  - pharos-cluster version 1.3.0-rc.3
Addon ingress-nginx cannot define any constants: FOO
/app/lib/pharos/addon.rb:20:in `addon'
/app/addons/ingress-nginx/addon.rb:3:in `<top (required)>'
/usr/local/lib/ruby/site_ruby/2.4.0/rubygems/core_ext/kernel_require.rb:59:in `require'
/usr/local/lib/ruby/site_ruby/2.4.0/rubygems/core_ext/kernel_require.rb:59:in `require'
/app/lib/pharos/addon_manager.rb:22:in `block (2 levels) in load_addons'
/app/lib/pharos/addon_manager.rb:22:in `each'
/app/lib/pharos/addon_manager.rb:22:in `block in load_addons'
/app/lib/pharos/addon_manager.rb:21:in `each'
/app/lib/pharos/addon_manager.rb:21:in `load_addons'
/app/lib/pharos/cluster_manager.rb:43:in `load'
/app/lib/pharos/version_command.rb:8:in `execute'
/app/lib/pharos/root_command.rb:18:in `run'
```

kke:
> True, that's an unintended side-effect... and possibly also explains why consts in class_eval are defined on the top-level scope, because otherwise any required gems would end up in the wrong place?

```ruby
class Foo
end

Foo.class_eval 'FOO="bar"'
```

This works just fine because the class has a name.

> But wouldn't those kinds of inside classes also suffer from exactly the same kind of global const collisions as consts like FOO = ...? That sounds like a feature, not a bug here.

Yes, that's why the solution could be to simply drop the block format `Pharos.addon` and face none of this kind of problems.

SpComb:
> > This also removes the possibility to use external gems
> 
> True, that's an unintended side-effect... and possibly also explains why consts in class_eval are defined on the top-level scope, because otherwise any required gems would end up in the wrong place?

Actually, `require '...'` still works fine, as long as the `addon.rb` does it outside of the `Pharos.addon do ...` block:

```ruby
require 'foo'

Pharos.addon 'foo' do
  ...
end
```

The addon could also define some custom top-level classes/modules there as well, if it wanted to.

PR #566, opened 2018-08-23
Nfs-provisioner addon

jakolehm:
https://github.com/kubernetes-incubator/external-storage/tree/master/nfs

Can be used with a claim like:
```yaml
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: nfs-test
  annotations:
    volume.beta.kubernetes.io/storage-class: "nfs-provisioner"
spec:
  accessModes:
    - ReadWriteMany
  resources:
    requests:
      storage: 1Gi
```

### DIFF
--- a/addons/nfs-provisioner/resources/04-pvc.yml.erb
+++ b/addons/nfs-provisioner/resources/04-pvc.yml.erb
@@ -8,4 +8,4 @@ spec:
     - ReadWriteOnce
   resources:
     requests:
-      storage: <%= config.storage_size %>Gi
+      storage: <%= config.storage_size %>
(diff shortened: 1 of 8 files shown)


addons/nfs-provisioner/resources/04-pvc.yml.erb
```diff
@@ -0,0 +1,11 @@
+kind: PersistentVolumeClaim
```
jnummelin: So this makes the assumption that there's already an existing default class setup for the cluster. Just thinking that can we really make that assumption?

jakolehm:
Kontena Storage addon already has ReadWriteMany support (#694)

PR #553, opened 2018-08-20
Normalize internal loading of files to avoid loading twice

kke:
Fixes #533 

- Changes all `require_relative`s to `require`
- Adds a `Pharos::RootPath` and uses it to require phases and internal addons
- `uniq`'s the list of addon directories in `load_addons`
- Changes `bin/pharos-cluster` to require the `lib/pharos_cluster` from `lib_path`

Option B would be to change all `require`s to `require_relative` and change `lib/pharos/autoload.rb` to use full paths.

Option C would be to remove all internal requires and make everything autoload either manually or dynamically in `autoload.rb`.

PR #576, opened 2018-08-29
Deprecate user addons directory in favor of pharos-addons

kke:
(base is pr #553)

- Adds a deprecation warning if `addons/` directory is added to addon loadpaths.
- Adds `pharos-addons/` to addon loadpaths.

### DIFF
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -45,9 +45,10 @@ module Pharos
       )
       addon_dirs = [
         File.join(__dir__, '..', '..', 'addons'),
-        File.join(Dir.pwd, 'addons'),
+        File.join(Dir.pwd, 'pharos-addons'),
         File.join(__dir__, '..', '..', 'non-oss', 'addons')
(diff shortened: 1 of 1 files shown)


lib/pharos/cluster_manager.rb
```diff
@@ -39,9 +39,17 @@ def addon_manager
...
       Pharos::PhaseManager.load_phases(__dir__ + '/phases/')
       addon_dirs = [
-        File.join(__dir__, '..', '..', 'addons'),
-        File.join(Dir.pwd, 'addons')
-      ] + @config.addon_paths.map { |d| File.join(Dir.pwd, d) }
+        File.join(__dir__, '..', '..', 'addons')
+      ]
+      addon_dirs += @config.addon_paths.map { |d| File.join(Dir.pwd, d) }
+
+      if File.directory?(File.join(Dir.pwd, 'addons')) && !File.exist?(File.join(Dir.pwd, 'pharos-cluster.gemspec'))
```
jakolehm: I think we don't need this warning anymore?

Issue #587, opened 2018-09-03
Pharos cluster does not support no_proxy configuration


ciniajohanjo:
In our environment all outgoing addresses require proxy. There is our own private docker repository, which is not accessible with proxy.

Working solution is to modify on all cluster nodes: 
manually add /etc/system.d/system/docker.d/http-proxy.conf to have NO_PROXY-option 

> Environment="NO_PROXY=repository.server.local" 

Also modifying  /etc/environment to contain following line 

> NO_PROXY="repository.server.local" 

It would be nice to have cluster.yml a option to setup the no-proxy-settings, due we need to verify the settings every time we run 
> pharos up -c cluster.yml 

Affects the pharos 1.3.0 -version

Sakhas:
Pharos still does not add NO_PROXY/no_proxy option configured in cluster.yml to docker http-config (/etc/system.d/system/docker.d/http-proxy.conf) for every node on the cluster. Http_proxy is now correctly set to http-proxy.conf though.

No_proxy still has to be set manually to docker http-proxy.conf after running pharos-cluster up.

kke:
New issue #850

PR #590, opened 2018-09-04
Run commands using --norc --noprofile

kke:
Changes the way comands are executed on remote hosts.

Before:

```
$ echo "command" | sudo ENV_VARIABLE=value sh -x
```

After:

```
$ echo "command" | sudo env -i ENV_VARIABLE=value bash --norc --noprofile -x -s --
```

The `env -i` runs bash without inheriting any environment variables from the parent process. When pharos opens an SSH session, all the scripts such as `.bash_profile` are evaluated and thus may set unexpected variables or define unexpected functions that break the pharos command execution (such as aliases for `ls` or `rm` or whatever).

This should help in situations where the shell init scripts output something to the console, which could get confused by pharos to be part of the command output.


jakolehm:
Does this still read `/etc/environment` ?

kke:
No, it sets `PATH`, `http_proxy`, `HTTP_PROXY` and `HTTPS_PROXY` from the parent process when launching subshells, something like:

```
sudo env -i PATH=$PATH http_proxy=$http_proxy SOME_ENV="some value" bash --norc --noprofile -x -s < "script"
```

Quite lengthy.

kke:
Related: #587

TODO: add `NO_PROXY`

kke:
Includes #611

lib/pharos/config_schema.rb
```diff
@@ -56,6 +56,7 @@ def self.messages
               optional(:ssh_key_path).filled
               optional(:container_runtime).filled(included_in?: ['docker', 'cri-o'])
               optional(:http_proxy).filled(:str?)
+              optional(:no_proxy).filled(:str?)
```
jakolehm: Just wondering if we should break backwards compatibility and change proxy keys to something like this:
```yaml
proxy:
  http: ""
  no_proxy: ""
```

jnummelin: +1 for this change

jakolehm: @kke ping

kke: Does it make sense to have it per-host?

kke: Or should it maybe actually be `environment:`

kke: In which case the whole thing could be 1) supplied to the Net::SSH::Start as is to be used in commands and 2) written into `/etc/environment` in a phase for daemons.

jakolehm: Generic `environment:` makes sense since these all are actually "just" environment variables. I would still keep these per host, we can later add "global" parameters that can be overridden per host.

kke:
Removed the unrelated stuff about setting the proxy stuff in `/etc/environment`, those are covered in #651

PR #607, opened 2018-09-18
Improve addon config error message

jakolehm:
Before:
```
Pharos::AddonManager::InvalidConfig : {:schedule=>["is missing"]}
```

After:
```
Pharos::AddonManager::InvalidConfig : host-upgrades => schedule is missing
```

Fixes #606

PR #609, opened 2018-09-18
Release v1.3.2

jakolehm:
## Changes since v1.3.1

- kubernetes 1.11.3 (#598)
- cri-o 1.11.3 (#599)
- weave net 2.4.1 (#605)
- improve addon config error message (#607)
- normalize add-on paths to avoid loading twice (#553)
- don't abort when apt mark unhold reports nothing to unhold (#593)
- fix broken pipe handling (#602)
- fix el7 kubelet restart on upgrade (#608)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    pharos-cluster (1.3.0.rc.1)
+    pharos-cluster (1.3.2)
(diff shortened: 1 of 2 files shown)


kke:
Is #567 irrelevant?

jakolehm:
> Is #567 irrelevant?

It was not in the milestone. Not sure if we have issues without it?

PR #611, opened 2018-09-18
Add "no_proxy" for setting NO_PROXY in /etc/environment

### LINKED ISSUE #587: Pharos cluster does not support no_proxy configuration

kke:
Fixes #587 

Adds `no_proxy` setting to host config:

```
hosts:
  - address: 1.1.1.1
    private_interface: eth1
    user: root
    ssh_key_path: ~/.ssh/my_key
    role: master
    no_proxy: "*.example.com,127.0.0.1"
```

The setting will be written to `/etc/environment`.

### DIFF
--- a/lib/pharos/config_schema.rb
+++ b/lib/pharos/config_schema.rb
@@ -56,6 +56,7 @@ module Pharos
               optional(:ssh_key_path).filled
               optional(:container_runtime).filled(included_in?: ['docker', 'cri-o'])
               optional(:http_proxy).filled(:str?)
+              optional(:no_proxy).filled(:str?)
             end
           end
         end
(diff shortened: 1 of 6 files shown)


kke:
Removes the seemingly pointless `SET_HTTP_PROXY` toggle which, as far as I can see, does nothing, except allows setting blank proxy settings, which essentially is the same as removing them?

kke:
`FTP_PROXY` is still not being set, but maybe it's irrelevant.

kke:
Replaced by #651

PR #617, opened 2018-09-21
Kontena license

jakolehm:
- `non-oss` folder is licenced under "Kontena License"
- new phase: validate cluster version
  - oss version: allow patch upgrades
  - non-oss version: allow all upgrades
  - disallow downgrades
- new non-oss addons:
  - kontena-lens
  - license-manager
- two build artifacts: oss and non-oss binary


non-oss/addons/kontena-lens/resources/02-service-account.yml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: lens-operator
+  namespace: kontena-lens
+imagePullSecrets:
```
nevalla: No need to assign imagePullsecrets

non-oss/addons/kontena-lens/resources/03-cluster-role-binding.yml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: ClusterRoleBinding
+metadata:
+  name: lens-operator
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
```
nevalla: I guess this is fine for now, but we need to add proper ClusterRole definition later

non-oss/addons/kontena-lens/resources/07-dashboard-deployment.yml
```diff
@@ -0,0 +1,59 @@
...
+        - image: quay.io/kontena/lens:latest
+          name: dashboard
+          imagePullPolicy: Always
+          env:
+            - name: KUBE_KONTENA_URL
+              value: http://usermanagement:9999
+            - name: KUBE_TERMINAL_URL
+              value: http://localhost:9998
+            - name: REDIS_CLIENT_HOST
+              value: redis
```
nevalla: I think we should enable built-in user management by default or add it behind config variable:
```
- name: USER_MANAGEMENT_ENABLED
   value: "true"
```

non-oss/addons/kontena-lens/resources/02-service-account.yml
jakolehm: Fixed.

jakolehm:
> We could also open separate PR to polish those after this is merged. WDYT?

Yeah, we probably should create a followup PR for any enhancements.. let's fix only major issues here.

build/drone/ubuntu_oss.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# build binary
+apt-get update -y
+apt-get install -y -q squashfs-tools build-essential ruby bison ruby-dev git-core texinfo curl
+curl -sL https://dl.bintray.com/kontena/ruby-packer/0.5.0-dev/rubyc-linux-amd64.gz | gunzip > /usr/local/bin/rubyc
+chmod +x /usr/local/bin/rubyc
+gem install bundler
+version=${DRONE_TAG#"v"}
+package="pharos-cluster-oss-linux-amd64-${version}"
+rubyc -o $package pharos-cluster
```
kke: Should have the stuff from #621

build/drone/ubuntu_oss.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# build binary
+apt-get update -y
+apt-get install -y -q squashfs-tools build-essential ruby bison ruby-dev git-core texinfo curl
+curl -sL https://dl.bintray.com/kontena/ruby-packer/0.5.0-dev/rubyc-linux-amd64.gz | gunzip > /usr/local/bin/rubyc
+chmod +x /usr/local/bin/rubyc
+gem install bundler
+version=${DRONE_TAG#"v"}
+package="pharos-cluster-oss-linux-amd64-${version}"
+rubyc -o $package -d /__enclose_io_memfs__ pharos-cluster
```
kke: need to `sudo mkdir` that dir first

kke: and the build user needs access

jakolehm: > need to sudo mkdir that dir first

Good catch, fixed.

> and the build user needs access

Build user is root.

PR #623, opened 2018-09-25
Rename OSS build packages

nevalla:
This PR will rename built oss packages as `pharos-cluster-linux-amd64-${version}-oss`.

The motivation behind this is that we could then use `${version}-oss` as version for the binary.  It would make more straight-forward to install and resolve binaries, for example:
```sh
$ chpharos list-remote
2.0.0
2.0.0-oss
1.3.2
1.3.1
...
$ chpharos install 2.0.0-oss
```


kke:
I think `+` would be more semver (`2.0.0-rc1+oss`), but I'm fine with this too.

nevalla:
yes, that `+oss` might be better. I Will change it.

PR #634, opened 2018-09-28
Kontena backup addon implementation

### LINKED ISSUE #94: Backup add-on
Would be nice to get some backup setup automatically as on add-on.

Not sure if we'd need to backup anything else than etcd?

jnummelin:
Can be setup like:
```
  kontena-backup:
    enabled: true
    cloud_credentials: /Users/jussi/.aws/credentials
    aws:
      bucket: jussi-backup-test
      region: eu-central-1
```
`cloud_credentials` refers to a file where cloud credential can be loaded from and fed into the secret the addon is using.

Supports now AWS and GCP, I tried to model eveything so that it's pretty easy to build support for other clouds easily.

WIP items:

- [ ] Docs
- [x] Hash stringification needs to be generalized
~- [ ] Decide better place to load commands dynamically~

fixes #94


jnummelin:
Fixed all the autoloads.

Still need to figure out where to load the commands properly, the current way in autoload just feels bit wrong. Any good ideas are welcome. :)

Also need to figure out why the specs fail on Drone, works-on-my-machine and also when running on `ruby:2.4` container as Drone also does. :/

non-oss/commands/backup/client_helper.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+module ClientHelper
+  # FIXME These kube client related helpers should be somewhere more common
+  def client
+    @client ||= create_client
+  end
+
+  def create_client
+    if ENV['KUBE_TOKEN'] && ENV['KUBE_CA'] && ENV['KUBE_SERVER']
```
kke: (We have this snippet several times in several projects, it should probably be in k8s-client)

lib/pharos/autoload.rb.orig
```diff
@@ -0,0 +1,70 @@
+# frozen_string_literal: true
```
jakolehm: Merge conflict?

non-oss/addons/kontena-backup/addon.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+
+Pharos.addon 'kontena-backup' do
+  using Pharos::CoreExt::DeepTransformKeys
+
+  version '0.1.0'
```
jakolehm: Just thinking should we follow pharos version here? Anyway `0.1.0` sounds a bit too unstable 😄

non-oss/addons/kontena-backup/resources/40-toolbox-deployment.yml
```diff
@@ -0,0 +1,25 @@
...
+  template:
+    metadata:
+      labels:
+        component: ark
+    spec:
+      restartPolicy: Always
+      serviceAccountName: ark
+      containers:
+        - name: toolbox
+          image: quay.io/kontena/pharos-backup-toolbox:0.1.0
```
jakolehm: Should use pharos registry.

non-oss/addons/kontena-backup/addon.rb
jakolehm: Auto-bump:  `version: Pharos::VERSION` ?

jnummelin: that then creates a problem on docs side, as we have the version in addon docs

jakolehm: As discussed in Slack: `0.9.6+kontena.1`

non-oss/addons/kontena-backup/resources/40-toolbox-deployment.yml
jnummelin: done

PR #635, opened 2018-09-28
Nested transform helpers to hashes and case conversions to strings

kke:
### Adds:
- `Hash#deep_transform_keys`
- `Hash#deep_stringify_keys`
- `String#camelcase`
- `String#underscore`
- `String#camelback`

### Usage:

Either as a module or a refinement (no global monkey patch)

```ruby
class Foofoo
  using Pharos::CoreExt::DeepTransformKeys

  def test
    { foo: { bar: [ { baz: 1 } ] } }.deep_transform_keys { |k| k.to_s.reverse.to_sym })
  end
  # => { oof: { rab: [ { zab: 1} ] } }
end
```

```ruby
class Foofoo
  def test
    { foo: { bar: [ { baz: 1 } ] } }.extend(Pharos::CoreExt::DeepTransformKeys).deep_stringify_keys
  # => { "foo" => { "bar" => [ { "baz" => 1} ] } }
end
```

```ruby
class Foofoo
  using Pharos::CoreExt::StringCasing

  def test
    "foo Bar baz".underscore
  end
  # => "foo_bar_baz"
end
```

```ruby
class Foofoo
  def test
    str = "foo BarBaz"
    str.extend(Pharos::CoreExt::StringCasing)
    str.camelback
  end
  # => "fooBarBaz"
end
```

### DIFF
--- a/lib/pharos/autoload.rb
+++ b/lib/pharos/autoload.rb
@@ -31,6 +31,8 @@ module Pharos
 
   module CoreExt
     autoload :IPAddrLoopback, 'pharos/core-ext/ip_addr_loopback'
+    autoload :DeepTransformKeys, 'pharos/core-ext/deep_transform_keys'
+    autoload :StringCasing, 'pharos/core-ext/string_casing'
   end
 
   module SSH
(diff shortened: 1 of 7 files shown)


jnummelin:
added to 2.0 milestone as there's few other PRs that would benefit from these "helpers"

kke:
Of course all of this would be available through some 3rd party gems.

PR #642, opened 2018-10-02
Remove SSHManager, HostConfigManager and PhaseManager

### LINKED ISSUE #643: ConfigureTelemetry spec is flaky
`configure_telemetry_spec` is flaky when `rspec --order rand` used (like it should)

Getting either `invalid superclass for File` or `undefined method 'missing_extensions?' for nil:NilClass`.

Probably something to do with FakeFS.

kke:
### Removes:

- `Pharos::SSH::Manager` by adding `Pharos::Configuration::Host#ssh`
- `Pharos::HostConfigManager` by refactoring `Pharos::Host::Configurer` interface
- `Pharos::PhaseManager` by moving retry logic to `Pharos::Phase#run` and scheduling logic (`run_parallel`/`_serial`) to `Pharos::ClusterManager`

### Rationale:

- SSH manager only held a hash of `{ host => SSH::Client.new(host.address) }` and a method to `disconnect_all`. The config is pretty much always in memory and the hosts can therefore maintain their own ssh clients memoized. All clients can be disconnected via `config.hosts.map { |host| host.ssh.disconnect }`.
- `HostConfigManager` did some weird mass-assignment to inject the same current config to a singleton/class instance variable in each configurer class. This was the whole functionality of HostConfigManager: `Dir.glob` load all configurers and inject the config.
- `PhaseManager` was doing what a Phase should have been doing (run itself). There's still room for refactoring like in #630 (which needs reworking)

### ClusterManager

`ClusterManager` still does the work of `UpCommand` and includes some functionality that should actually be in `AddonManager`. Its most essential functionality is to define the order of phases. #630 or some variation of it should drop `ClusterManager` too.

Fixes #643 (configure telemetry spec FakeFS flakiness)


kke:
Includes a lot of whitespace autocorrects from upgraded rubocop(?)

kke:
Makes  `Pharos::Phases::ConfigureTelemetry#customer_token returns empty string if not found` spec flaky due to the `--order rand`.

lib/pharos/cluster_manager.rb
```diff
@@ -50,19 +41,18 @@ def load
       addon_dirs = addon_dirs.map { |dir| Pathname.new(dir).realpath.to_s }.uniq
 
       Pharos::AddonManager.load_addons(*addon_dirs)
-      Pharos::HostConfigManager.load_configs(@config)
     end
 
     def gather_facts
-      apply_phase(Phases::GatherFacts, config.hosts, ssh: true, parallel: true)
+      apply_phase(Phases::GatherFacts, config.hosts, parallel: true)
```
jakolehm: Just thinking if `Phase` should provide info about parallelism?

lib/pharos/host/configurer.rb
```diff
@@ -1,15 +1,21 @@
...
 module Pharos
   module Host
     class Configurer
-      attr_reader :host, :ssh
+      attr_reader :host, :config
 
       SCRIPT_LIBRARY = File.join(__dir__, '..', 'scripts', 'pharos.sh').freeze
 
-      def initialize(host, ssh)
+      def initialize(host, config = nil)
```
jakolehm: What is the `config` here?

lib/pharos/host/configurer.rb
```diff
@@ -91,6 +97,18 @@ def docker?
         @host.docker?
       end
 
+      def self.configurers
+        @configurers ||= Set.new
+      end
+
+      def self.for_os_release(os_release, os_version = nil)
```
jakolehm: Yardocs would help.

lib/pharos/phase.rb
```diff
@@ -1,11 +1,22 @@
 # frozen_string_literal: true
 
 require 'logger'
+require 'openssl'
```
jakolehm: Openssl is autoloaded but required here, why?

lib/pharos/phase.rb
```diff
@@ -24,16 +35,18 @@ def self.register_component(component)
 
     # @param host [Pharos::Configuration::Host]
     # @param config [Pharos::Config]
-    # @param ssh [Pharos::SSH::Client]
     # @param master [Pharos::Configuration::Host]
```
jakolehm: We are missing `cluster_context` yardoc.

lib/pharos/phase.rb
```diff
@@ -24,16 +35,18 @@ def self.register_component(component)
...
+    def initialize(host, config: nil, master: nil, cluster_context: nil)
       @host = host
       @config = config
-      @ssh = ssh
       @master = master
       @cluster_context = cluster_context
     end
 
+    def ssh
+      @ssh ||= host.ssh
```
jakolehm: Host already caches `ssh`, no need to do it here?

lib/pharos/cluster_manager.rb
kke: Done in #630 or other follow-up to this one

lib/pharos/host/configurer.rb
kke: Instead of the `configurers.each { |c| c.cluster_config = config }` the configurer gets the config via initialize.

lib/pharos/phase.rb
kke: No need really, but gets a direct reference instead of having to do a method lookup + invocation on 
`host` every time.

jakolehm: I would then remove it from here, I'm quite sure that we don't really see any difference in performance.

lib/pharos/host/configurer.rb
jakolehm: Ok, we should add yardoc for more pleasant reading experience.

kke:
Fixed the telemetry spec flakiness, so Fixes #643

kke:
Too far out of sync

PR #646, opened 2018-10-04
Addon: Helm

jakolehm:
Deploys [Helm](https://www.helm.sh) Tiller to `kube-system` namespace. Tiller is secured so that it listens only on localhost. In practice this means that only users with proper RBAC rights can access tiller via `helm` cli.

## Addon configuration

```yaml
addons:
  helm:
    enabled: true
```

## Helm CLI configuration

```
$ helm init -c
```


addons/helm/resources/deployment.yml.erb
```diff
@@ -0,0 +1,46 @@
...
+        app: helm
+        name: tiller
+    spec:
+      containers:
+      - env:
+        - name: TILLER_NAMESPACE
+          value: kube-system
+        - name: TILLER_HISTORY_MAX
+          value: "10"
+        image: gcr.io/kubernetes-helm/tiller:v<%= version %>
```
nevalla: where is this coming from?

jakolehm: Do you mean version or?

nevalla: yes, version.

jakolehm: It's the addon version.

PR #647, opened 2018-10-04
Add support for enabling/disabling admission plugins

### LINKED ISSUE #644: Add configurability for admission controllers
https://kubernetes.io/docs/reference/access-authn-authz/admission-controllers

Would be nice to have the possibility to define in cluster.yml which admission controllers are enabled. The enabled-by-default list is (as of 1.11):
```
NamespaceLifecycle
LimitRanger
ServiceAccount
PersistentVolumeLabel
DefaultStorageClass
DefaultTolerationSeconds
MutatingAdmissionWebhook
ValidatingAdmissionWebhook
ResourceQuota
Priority
```

I guess the biggest "problem" is how to handle the configuration of the admission controller as some of them require extra config files etc. in place.

All the ac's should be configurable via single file, which in turn point to other files:
```
kind: AdmissionConfiguration
apiVersion: apiserver.k8s.io/v1alpha1
plugins:
- name: EventRateLimit
  path: eventconfig.yaml
...
```

So it might make sense to have something like this in the cluster.yml:
```
admission_controllers:
  - name: EventRateLimit
    config: path/to/config
  - name: PodSecurityPolicy
  - name: SomeOther
```
So the config file path would be optional as not all ac's need those at all.

jnummelin:
fixes #644 

Introduces new top-level config key `admission_plugins` and can be used like:
```
admission_plugins:
  - name: AlwaysPullImages
    enabled: true
  - name: LimitRanger
    enabled: false
```

This now supports the cases where user can enable those plugins that do **NOT** require any additional config files in place. For those we can later add something like `config_path` key to upload the given config file in place.

I used `admission plugin` as the term, kube docs seems to use both `admission controller` and `admission plugin` pretty inconsistently.

### DIFF
--- a/lib/pharos/config.rb
+++ b/lib/pharos/config.rb
@@ -11,6 +11,7 @@ require_relative 'configuration/audit'
 require_relative 'configuration/kube_proxy'
 require_relative 'configuration/kubelet'
 require_relative 'configuration/telemetry'
+require_relative 'configuration/admission_plugin'
 
 module Pharos
   class Config < Pharos::Configuration::Struct
@@ -46,6 +47,7 @@ module Pharos
     attribute :image_repository, Pharos::Types::String.default('quay.io/kontena')
     attribute :addon_paths, Pharos::Types::Array.default([])
     attribute :addons, Pharos::Types::Hash.default({})
+    attribute :admission_plugins, Types::Coercible::Array.of(Pharos::Configuration::AdmissionPlugin)
 
     attr_accessor :data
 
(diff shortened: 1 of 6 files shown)


jnummelin:
Seems that by default kubeadm does some enabling and disabling too:
```
    - --disable-admission-plugins=PersistentVolumeLabel
    - --enable-admission-plugins=NodeRestriction
```
Those are what gets set by kubeadm if no config is specified for Pharos.

jnummelin:
Changed now to work with PSPs.  `PodSecurityPolicy` and `NodeRestriction` plugins are enabled by default. They can be disabled explicitly through cluster.yml

jnummelin:
I don't think we need to worry about `PersistentVolumeLabel` plugin:

> Starting from 1.11, this admission controller is disabled by default.

PR #649, opened 2018-10-04
Improve Kontena Lens RBAC rules

nevalla:
Refactored previous `cluster-admin` role binding into separated roles and bindings. Maybe need to think about naming a bit still.

Also changed `name`, `host` and `email` as optional.


non-oss/addons/kontena-lens/resources/07-dashboard-deployment.yml
```diff
@@ -27,6 +27,8 @@ spec:
               value: http://localhost:9998
             - name: REDIS_CLIENT_HOST
               value: redis
+            - name: USER_MANAGEMENT_ENABLED
```
jakolehm: Should we add option for this?

nevalla: Yes, don't see any reason why not

PR #651, opened 2018-10-05
Add "environment:" host config to update /etc/environment

### LINKED ISSUE #587: Pharos cluster does not support no_proxy configuration

kke:
Fixes #587 

Removes the host `proxy:` configuration and replaces it with a generic "environment:" configuration for updating `/etc/environment`.

When setting the `http_proxy`, you should usually also set the uppercase `HTTP_PROXY` since some programs use one and some the other.

This also enables the use of `NO_PROXY` as requested in #587 and separately resolved in #611 

To remove an entry from `/etc/environment` set a null value:

```yaml
hosts:
  - address: 127.0.0.1
     environment:
       PATH: null # or leave blank
```

### DIFF
--- a/lib/pharos/host/configurer.rb
+++ b/lib/pharos/host/configurer.rb
@@ -96,6 +96,36 @@ module Pharos
         self.class.cluster_config
       end
 
+      # @return [Pharos::SSH::File]
+      def env_file
+        @ssh.file('/etc/environment')
+      end
+
+      def update_env_file
+        return if @host.environment.nil? || @host.environment.empty?
+
+        host_env_file = env_file
+        if host_env_file.exist?
+          original_data = host_env_file.read.lines.map do |line|
+            line.strip!
+            next if line.start_with?('#')
+            key, val = line.split('=', 2)
+            val = nil if val.to_s.empty?
+            [key, val]
+          end.to_h
+        else
+          original_data = {}
+        end
+
+        new_content = original_data.merge(@host.environment).compact.map do |key, val|
+          next if val.nil?;
+          "#{key}=#{val}"
+        end.join("\n") + "\n"
+
+        host_env_file.write(new_content)
+        @ssh.disconnect; @ssh.connect # reconnect to reread env
+      end
+
       class << self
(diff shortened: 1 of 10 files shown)


lib/pharos/host/configurer.rb
```diff
@@ -96,6 +96,35 @@ def cluster_config
         self.class.cluster_config
       end
 
+      # @return [Pharos::SSH::File]
+      def env_file
+        @ssh.file('/etc/environment')
+      end
+
+      def update_env_file
+        return if @host.environment.nil? || @host.environment.empty?
```
jakolehm: Is this one-way thing? One can add envs but never remove them?

kke: Yes. If the `original_data.merge(@host.environment).compact.map` below would be changed to something that would not overwrite `nil`'s during merge, you could delete by setting `environment: FOOFOO: null`.

kke: Now changed to do so.

```ruby
{ a: nil, b: 'foo' }.merge(a: 'foo', c: 'bar') { |a,b| b }
=> {:a=>nil, :b=>"foo", :c=>"bar"}
{ a: nil, b: 'foo' }.merge(a: 'foo', c: 'bar') { |a,b| b }.compact
=> {:b=>"foo", :c=>"bar"}
```

jakolehm: Specs would be nice.

kke: Whoomp, there it is.

PR #653, opened 2018-10-05
Pod security policies

jakolehm:
## Usage

```yaml
pod_security_policy:
  default_policy: "99-pharos-restricted"
```


jakolehm:
I think we need to make "privileged" policy the default (plus an option to change the default). This way all the helm charts etc will work and users can still harden the cluster via cluster.yml. 

@jnummelin WDYT?

jakolehm:
There is now 2 policies: `00-pharos-privileged` and `99-pharos-restricted`. Privileged is the default policy.

PR #655, opened 2018-10-09
Switch to multi-arch compatible image registry

jakolehm:
- moves all core & addon images to a central registry
- uses multi-arch images wherever possible -> makes possible to mix amd64 & arm64 machines in a cluster
- improves air-gapped setups

### DIFF
--- a/addons/ingress-nginx/addon.rb
+++ b/addons/ingress-nginx/addon.rb
@@ -24,6 +24,7 @@ Pharos.addon 'ingress-nginx' do
     apply_resources(
       configmap: config.configmap || {},
       node_selector: config.node_selector || {},
+      default_backend_image: config.default_backend['image'],
       default_backend_replicas: default_backend_replicas
     )
   }
--- a/addons/ingress-nginx/addon.rb
+++ b/addons/ingress-nginx/addon.rb
@@ -22,23 +22,9 @@ Pharos.addon 'ingress-nginx' do
     apply_resources(
       configmap: config.configmap || {},
       node_selector: config.node_selector || {},
-      image: image_name,
       default_backend_replicas: default_backend_replicas
     )
   }
-
-  DEFAULT_BACKEND_ARM64_IMAGE = 'docker.io/kontena/pharos-default-backend-arm64:0.0.2'
-  DEFAULT_BACKEND_IMAGE = 'docker.io/kontena/pharos-default-backend:0.0.2'
-
-  def image_name
-    return config.default_backend[:image] if config.default_backend&.dig(:image)
-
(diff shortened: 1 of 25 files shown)


jakolehm:
@jnummelin @kke PTAL

addons/ingress-nginx/addon.rb
```diff
@@ -22,24 +24,10 @@
     apply_resources(
       configmap: config.configmap || {},
       node_selector: config.node_selector || {},
-      image: image_name,
```
jnummelin: This essentially makes the default backend image non-configurable. Why not just pass the `config.default_backend[:image]`, that should default correctly to `registry.pharos.sh/...`

lib/pharos/phases/upgrade_master.rb
```diff
@@ -36,51 +36,12 @@ def upgrade
...
 
-        dns_patch_thread = create_dns_patch_thread if dns_patch_needed?
         @ssh.tempfile(content: cfg.to_yaml, prefix: "kubeadm.cfg") do |tmp_file|
           @ssh.exec!("sudo /usr/local/bin/pharos-kubeadm-#{Pharos::KUBEADM_VERSION} upgrade apply #{Pharos::KUBE_VERSION} -y --ignore-preflight-errors=all --allow-experimental-upgrades --config #{tmp_file}")
         end
-        dns_patch_thread.join if dns_patch_needed?
 
         logger.info { "Control plane upgrade succeeded!" }
       end
-
```
jnummelin: Yay, we can finally get rid of this ugly hack :D

lib/pharos_cluster.rb
```diff
@@ -14,4 +14,5 @@ module Pharos
   ETCD_VERSION = ENV.fetch('ETCD_VERSION') { '3.2.18' }
   KUBELET_PROXY_VERSION = '0.3.7'
   COREDNS_VERSION = '1.1.3'
+  TELEMETRY_VERSION = '0.1.0'
```
jnummelin:
```
$ docker pull registry.pharos.sh/kontenapharos/pharos-telemetry:0.1.0
Error response from daemon: manifest for registry.pharos.sh/kontenapharos/pharos-telemetry:0.1.0 not found
```

addons/ingress-nginx/addon.rb
jakolehm: Good catch, fixed.

lib/pharos_cluster.rb
jakolehm: Multi-arch repo should be there now.

PR #666, opened 2018-10-11
Upgrade kubeadm via configured repository

jakolehm:
We should try to avoid hardcoding any urls in scripts... this PR improves air-gapped setups because it does not depend on a fixed repository.


lib/pharos/host/ubuntu/ubuntu.rb
```diff
@@ -7,12 +7,6 @@ def install_essentials
         exec_script('configure-essentials.sh')
       end
 
-      def configure_repos
```
jakolehm: Cleanup some leftovers...

PR #669, opened 2018-10-11
 Add support for post install messages

nevalla:
This PR adds support for post install messages. This PR focus only on addons, but it can be easily leverage to cover also phases too.

PR adds `#post_install_message` method to `Addon` class and it can be used like this:
```ruby
# addon.rb
install {
    host = config.host || "lens.#{worker_node_ip}.nip.io"
    apply_resources(
      host: host
    )
    post_install_message("Kontena Lens is running at https://#{host}.")
  }
```

These post install messages are displayed after "Cluster has been crafted" messages
```
==> Cluster has been crafted! (took 1 minute 14 seconds)
    To configure kubectl for connecting to the cluster, use:
      pharos kubeconfig -c cluster.yml  > kubeconfig
      export KUBECONFIG=./kubeconfig
    Post-install message from kontena-lens:
      Kontena Lens is running at https://<url>.
```

### DIFF
--- a/lib/pharos/addon.rb
+++ b/lib/pharos/addon.rb
@@ -133,7 +133,7 @@ module Pharos
       end
     end
 
-    attr_reader :config, :cpu_arch, :cluster_config, :kube_client
+    attr_reader :config, :cpu_arch, :cluster_config, :kube_client, :post_install_message
 
     # @param config [Hash,Dry::Validation::Result]
     # @param enabled [Boolean]
@@ -186,6 +186,10 @@ module Pharos
       else
         delete_resources
       end
+    end
+
+    def notify(msg)
+      @post_install_message = msg
     end
 
     # @param vars [Hash]
--- a/lib/pharos/addon.rb
+++ b/lib/pharos/addon.rb
@@ -133,7 +133,7 @@ module Pharos
       end
     end
 
-    attr_reader :config, :cpu_arch, :cluster_config, :kube_client, :post_install_message
+    attr_reader :config, :cpu_arch, :cluster_config, :kube_client
 
     # @param config [Hash,Dry::Validation::Result]
     # @param enabled [Boolean]
@@ -188,8 +188,12 @@ module Pharos
       end
     end
 
-    def notify(msg)
-      @post_install_message = msg
+    def post_install_message(msg = nil)
+      if msg
+        @post_install_message = msg
+      else
+        @post_install_message
+      end
     end
 
     # @param vars [Hash]
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -13,7 +13,9 @@ module Pharos
     def initialize(config, pastel: Pastel.new)
       @config = config
       @pastel = pastel
-      @context = {}
+      @context = {
+        'post_install_messages' => {}
+      }
     end
 
     # @return [Pharos::SSH::Manager]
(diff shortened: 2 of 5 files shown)


lib/pharos/addon.rb
```diff
@@ -188,6 +188,10 @@ def apply_uninstall
       end
     end
 
+    def notify(msg)
```
nevalla: Not sure if `notify_user` would be better name

jakolehm: Why `notify` and not `post_install_message`?

lib/pharos/up_command.rb
```diff
@@ -116,7 +116,12 @@ def configure(config)
       puts "    To configure kubectl for connecting to the cluster, use:"
       puts "      #{File.basename($PROGRAM_NAME)} kubeconfig #{defined_opts}> kubeconfig"
       puts "      export KUBECONFIG=./kubeconfig"
-
+      manager.post_install_messages.each do |component, message|
+        puts "    Post-install message from #{component}:"
+        message.split("\n").each do |line|
```
nevalla: This is mainly for formatting the multi-line message with proper identation

nevalla:
Renamed `Addon#notify` to `Addon#post_install_message`

lib/pharos/addon.rb
kke:
```ruby
def post_install_message(msg)
  @cluster_context['post_install_messages'][self.class.addon_name] ||= []
  @cluster_context['post_install_messages'][self.class.addon_name].concat(msg.lines)
end
```

with this single-string approach you need to do something like:
```
notify "foo\nfoofoo\nfoo\n"
```

and if you want to add to it:

```
notify "foofoo\nfoofoo"
@post_install_message << "\nfoofoo"
```

lib/pharos/cluster_manager.rb
```diff
@@ -13,7 +13,9 @@ class ClusterManager
     def initialize(config, pastel: Pastel.new)
       @config = config
       @pastel = pastel
-      @context = {}
+      @context = {
+        'post_install_messages' => {}
```
kke: and this could then be `'post_install_messages' => Hash.new { |hash,key| hash[key] = [] }` so you don't need the `||= []` in the `post_install_message` method

lib/pharos/addon.rb
kke: oh, addons don't have direct access to cluster context. maybe they should.

jakolehm: We were discussing about cluster_context in addons and decided not to pass it there.. is it really a good practice to "leak" cluster_context to addons?

kke: It already leaks the stuff that is updated directly to `config.hosts` by `gather_facts` et al, like  `host.checks` and whatever.

Maybe `cluster_context` could actually be an instance variable of `config` anyways. then addons, phases and host_configurers could all reach it through `config.cluster_context` since they all get the config.

kke: btw i don't know how threadsafe the cluster_context thing is anyways.. maybe it should be an instance of `Concurrent::Hash` instead of a regular hash (dry already brings in `concurrent-ruby`).

nevalla:
I would like to see this PR as simple as possible and not going to refactor `cluster_context` part.

@jakolehm @jussi do you have opinion where should we outout these post install messages? after or before:
```
To configure kubectl for connecting to the cluster, use:
      pharos kubeconfig -c cluster.yml  > kubeconfig
      export KUBECONFIG=./kubeconfig
```

jnummelin:
I'd put those generic messages BEFORE the config hint

PR #670, opened 2018-10-12
[PoC] Plugin as subcommand

kke:
As discussed.

With this you can do something like:

```
$ cat <<EOF > /usr/local/bin/pharos-foofoo
#!/bin/bash

echo "Hello!"
EOF
$ chmod +x  /usr/local/bin/pharos-foofoo
$ pharos foofoo
Hello
```

Also it could be possible to scan the PATH for `pharos-*`and inject a wrapper-subcommand. This way there would also be a mention of any installed plugins in the main command help.

I don't know if there's any benefit.

### DIFF
--- a/lib/pharos/command.rb
+++ b/lib/pharos/command.rb
@@ -32,5 +32,19 @@ module Pharos
     def stdin_eof?
       $stdin.eof?
     end
+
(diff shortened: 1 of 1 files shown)


kke:
PoC pocced

PR #672, opened 2018-10-15
Improve system pod qos/priorities

jakolehm:


jakolehm:
We need to add new priorityclass for system-level addons.

jnummelin:
> We need to add new priorityclass for system-level addons.

I think that can be handled as separate PR

lib/pharos/resources/calico/daemonset.yml.erb
```diff
@@ -38,6 +38,7 @@ spec:
         - effect: NoExecute
           operator: Exists
       serviceAccountName: calico-node
+      priorityClassName: system-cluster-critical
```
jnummelin: Maybe even `system-node-critical`

lib/pharos/scripts/configure-etcd.sh
```diff
@@ -21,6 +21,7 @@ metadata:
   name: etcd
   namespace: kube-system
 spec:
+  priorityClassName: system-cluster-critical
```
jnummelin: I think this needs to be `system-node-critical` that has the highest priority. I mean if etcd would get preempted, everything breaks. :) Although this is a static pod, so I don't think these will actually ever get preempted.

lib/pharos/resources/weave/daemon-set.yml.erb
```diff
@@ -88,6 +88,7 @@ spec:
       securityContext:
         seLinuxOptions: {}
       serviceAccountName: weave-net
+      priorityClassName: system-cluster-critical
```
jnummelin: Same as for Calico, I'd put these as `system-node-critical`

jakolehm:
> I'd bump some of the priorities

Bumped.

PR #677, opened 2018-10-16
Allow metrics-server to be placed on a master node

### LINKED ISSUE #676: Metrics-server requires at least one worker node
Without any worker nodes installation will halt because metrics-server is never ready.

jakolehm:
Fixes #676

### DIFF
--- a/lib/pharos/resources/metrics-server/metrics-server-deployment.yml.erb
+++ b/lib/pharos/resources/metrics-server/metrics-server-deployment.yml.erb
@@ -16,6 +16,9 @@ spec:
         k8s-app: metrics-server
     spec:
       serviceAccountName: metrics-server
+      tolerations:
+        - effect: NoSchedule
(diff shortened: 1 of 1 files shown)


jnummelin:
What if we add this toleration AND add preference affinity to worker node? With something like this:
```
 preferredDuringSchedulingIgnoredDuringExecution:
      - weight: 1
        preference:
          matchExpressions:
          - key: node-role.kubernetes.io/master
            operator: DoesNotExist
```

That would usually run it on workers but in case masters-only cluster it would allow it on on them.

jakolehm:
> Just thinking if this could/should be a cluster.yml option. Although I don't really see why it would cause any harm to run it on masters.

I think we can automatically detect if we actually need this toleration... it's only needed if cluster.yml does not have any worker nodes.

jakolehm:
@jnummelin PTAL

PR #688, opened 2018-10-17
Improve kontena-lens addon

nevalla:
This PR improves Kontena Lens addon by enabling token authentication webhook and creating lens config during the installation.

It also changes addon configuration a bit:
```
kontena-lens:
  enabled: true
  name: 'name of the cluster' # default 'pharos-cluster'
  host: 'https://lens-host' # default 'https://<worker-ip->.nip.io'
  user_management:
    enabled: true|false # default true
  tls:
    email: john@example.org # email used for LE certificate fetching, not required if host is not defined
```
This PR adds new `modify_cluster_config` hook to `Pharos::Addon`. Hook allows addon to set cluster configuration option if it is not set yet. `kontena-lens` addon adds token authentication webhook to point `lens-authenticator` pod running on master hosts if user management is enabled. 

## Todo
- [x] test without host option
- [x] test with custom domain

### DIFF
--- a/lib/pharos/addon_manager.rb
+++ b/lib/pharos/addon_manager.rb
@@ -78,7 +78,10 @@ module Pharos
 
     # @return [K8s::Client]
     def kube_client
-      @kube_client ||= Pharos::Kube.client(@config.master_host.api_address, @cluster_context['kubeconfig'])
+      if !@kubeclient && @cluster_context['kubeconfig']
+        @kube_client = Pharos::Kube.client(@config.master_host.api_address, @cluster_context['kubeconfig'])
+      end
+      @kube_client
     end
 
     def options
@@ -95,8 +98,6 @@ module Pharos
         addon = addon_class.new(config, enabled: true, **options)
         addon.validate
         yield_addon_with_retry(addon, &block)
-        post_install_message = addon.post_install_message
-        @cluster_context['post_install_messages'][addon.name] = post_install_message if post_install_message
       end
 
       with_disabled_addons do |addon_class|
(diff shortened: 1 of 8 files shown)


lib/pharos/addon_manager.rb
```diff
@@ -95,8 +98,6 @@ def each(&block)
         addon = addon_class.new(config, enabled: true, **options)
         addon.validate
         yield_addon_with_retry(addon, &block)
-        post_install_message = addon.post_install_message
-        @cluster_context['post_install_messages'][addon.name] = post_install_message if post_install_message
```
nevalla: moved post install message handling to `ClusterManager` because `each` method is not only for install

non-oss/addons/kontena-lens/addon.rb
```diff
@@ -5,20 +5,111 @@
...
+        },
+        user: {
+          name: 'kube-apiserver'
+        }
+      }
+    )
+  end
+
+  def wait_for_dashboard(host)
+    puts "    Waiting for Kontena Lens to get up and running"
```
jakolehm:
```suggestion
    puts "    Waiting for Kontena Lens to get up and running ..."
```

non-oss/addons/kontena-lens/addon.rb
```diff
@@ -5,20 +5,111 @@
...
+    nil
+  end
+
+  def init_lens_config(name, host, admin_password)
+    cluster_config = {
+      clusterName: name,
+      clusterUrl: "https://#{master_host_ip}:6443",
+      adminPassword: admin_password
+    }
+    command = "sudo curl -k -L -X POST -d '#{cluster_config.to_json}' -s -H \"Host: #{host}\" -H \"Content-Type: application/json\" http://localhost/api/cluster"
```
jakolehm: Just a note that our "cluster" CRD might be a bit too generic (meaning: name might clash)

nevalla: Actually, we don't need it at all. It's legacy. I will open an issue to get rid of it.

Issue #667, opened 2018-10-11
License assign sub-command


jakolehm:
Should exchange a valid license key to a subscription token and store it to `pharos-license` secret in `kube-system` namespace. 

This replaces `license-manager` addon functionality.

kke:
How to update the secret?

- [X] A: use `-c cluster.yml` + ssh to retrieve kubeconfig and use it to configure `k8s-client` 
- [ ] B: use `-c cluster.yml` + ssh + remote `kubectl`
- [ ] C: parse `ENV['KUBECONFIG']` using `Pharos::Kube::Config` and use the resulting config and the current context in it to configure `k8s-client`

I have a working implementation of option A currently but as it's already using ssh and #668 will also rely on running the remote kubectl, maybe that should be used here too.

kke:
Switched implementation to option `B`.

PR #690, opened 2018-10-17
(PoC) Add non-oss directory to LOAD_PATH

kke:
This would ease the subcommand registration in something like #667 or #634  and provide means to monkey-patch or replace the oss-implementations of classes.

When the `non-oss/` directory is in `$LOAD_PATH` before the `lib/` path, something like:

```ruby
require 'pharos/phases/upgrade_host'
```

 or

```ruby
module Pharos
  module Phases
    autoload :UpgradeHost, 'pharos/phases/upgrade_host'
  end
end
``` 

will load the file from `non-oss/pharos/phases/upgrade_host.rb` instead of `lib/pharos/phases/upgrade_host.rb` if it exists there.


jakolehm:
I don't like this idea, it's too "magical".

PR #693, opened 2018-10-17
Improved PATH manipulation in el7/centos configure-essentials.sh

### LINKED ISSUE #691: El7/Centos7 configure-essentials.sh removes /bin from PATH
https://github.com/kontena/pharos-cluster/blob/e31a0b534a6c5790757f3d53e66caec39690ea5f/lib/pharos/host/el7/scripts/configure-essentials.sh

1. It writes a new `PATH` into `/etc/environment` that does not include the `/bin`
2. It tries to use binaries from an `sbin` directory after that without reloading/sourcing the `/etc/environment` modified above (`getenforce`/`setenforce`)
3. It uses a quite uncertain way to detect if the path in `/etc/environment` should be set. (By current logic when the file does not contain the string `/usr/local/bin`)

kke:
Fixes #691 

- Inspects the current `PATH` to find out if setting a new default path is desirable instead of grepping the environment file for `/usr/local/bin` 
- Uses `lineinfile` instead of `>>` to update the PATH in env file
- Does not run `getenforce` inside a subshell


kke:
#692 replaced most of this.

kke:
Still partly relevant

kke:
#692 did not fix the removal of `/bin` from PATH.

kke:
Conflict unconflicted

PR #694, opened 2018-10-17
Kontena storage addon fixes

jakolehm:
- fix addon config errors
- implement `kontena-storage-fs` (ReadWriteMany)


jakolehm:
@nevalla PTAL

PR #698, opened 2018-10-18
Add developer and devops roles

nevalla:
This PR adds two roles `developer` and `devops`. 
With `developer` role user has read access to resources required by Kontena Lens. `devops` role is basically the same as `cluster-admin` but with it it's easy to give admin access to a certain namespace.

### DIFF
--- a/non-oss/pharos_pro/addons/kontena-lens/resources/developer-role.yml
+++ b/non-oss/pharos_pro/addons/kontena-lens/resources/developer-role.yml
@@ -1,0 +1,29 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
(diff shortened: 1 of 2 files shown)


non-oss/pharos_pro/addons/kontena-lens/resources/developer-role.yml
```diff
@@ -0,0 +1,29 @@
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: developer
```
jakolehm: I would add "delete pods" right.

nevalla: Yes, forgot that. Will add.

PR #703, opened 2018-10-22
add support for configuring insecure registries through cluster.yml

### LINKED ISSUE #588: Add support for insecure docker registry
Sometimes private docker registries don't use SSL-certificates. 

Could cluster.yml have  option to add list of insecure registries, which would be added to 

/etc/docker/daemon.json 

> {
>   ... 
>  "insecure-registries": ["insecure.server.local"]
>  ... 
> }

jnummelin:
fixes #588 

Insecure registries can be configured with:
```
container_runtime:
  insecure_registries:
    - "localhost:5000"
    - "foobar:5000"
```

Covers both Docker & CRI-O on all supported distros.

The `inspect` "hacks" are bit leaking now to all configurers, maybe I need to figure out single common place to generate the appropriate string format.

### DIFF
--- a/lib/pharos/host/configurer.rb
+++ b/lib/pharos/host/configurer.rb
@@ -95,6 +95,18 @@ module Pharos
         @host.custom_docker?
       end
 
+      # Return stringified json array for insecure registries
+      #
+      # @return [String]
+      def insecure_registries
+        if crio?
+          cluster_config.container_runtime.insecure_registries.map(&:inspect).join(",").inspect
+        else
+          # docker & custom docker
+          cluster_config.container_runtime.insecure_registries.to_json.inspect
+        end
+      end
+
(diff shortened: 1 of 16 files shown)


lib/pharos/host/configurer.rb
```diff
@@ -95,6 +95,18 @@ def custom_docker?
         @host.custom_docker?
       end
 
+      # Return stringified json array for insecure registries
+      #
+      # @return [String]
+      def insecure_registries
```
jakolehm: This needs specs... `.inspect` is so weird here 😄

jnummelin: Now there's some specs. There might be better ways to do this than with `.inspect`, essentially we want to properly escape the quotes.

jnummelin:
specs fixed

PR #705, opened 2018-10-23
Selectively reset a subset of hosts

### LINKED ISSUE #668: Possibility to reset only subset of hosts

### LINKED ISSUE #702: Reset & Up commands do not require --yes in non-interactive mode
The `UpCommand` and `ResetCommand` only check for `--yes` or prompts for confirmation when `$stdin.tty?` is true, otherwise they will proceed to perform the operation.

```
$ pharos reset
==> Do you really want to reset (reset will wipe configuration & data from all hosts)? (y/N)

$ pharos reset --yes
==> Sharpening tools ...
==> Starting to reset cluster ...

$ cat cluster.yml | pharos reset
==> Sharpening tools ...
==> Starting to reset cluster ...
```

Expected:

```
$ cat cluster.yml | pharos reset
ERROR: --yes option required when running in a non-interactive mode
```

kke:
Fixes #668 
Fixes #702 

Adds options to `pharos reset` to only reset a subset of hosts.

## TODO/OPEN
- The drain does not seem to finish on a three host setup
- ...


lib/pharos/configuration/host.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+# frozen_stritng_literal: true
```
jakolehm: What is this?

lib/pharos/cluster_manager.rb
```diff
@@ -132,7 +132,16 @@ def apply_phases
       apply_phase(Phases::ConfigureTelemetry, [master_hosts.first], master: master_hosts.first)
     end
 
-    def apply_reset
+    def apply_reset_hosts(hosts)
+      master_hosts = sorted_master_hosts
+      apply_phase(Phases::GatherFacts, hosts, ssh: true, parallel: true)
+      apply_phase(Phases::ConfigureClient, [master_hosts.first], ssh: true, master: master_hosts.first, parallel: false, optional: true)
+      apply_phase(Phases::Drain, hosts, parallel: false)
```
jakolehm: Drain & DeleteHost should be optional.. probably also ConfigureClient because master might be down.

lib/pharos/cluster_manager.rb
```diff
@@ -132,7 +132,20 @@ def apply_phases
...
+    def apply_reset_hosts(hosts, drain: true, delete: true)
+      master_hosts = sorted_master_hosts
+      apply_phase(Phases::GatherFacts, hosts, ssh: true, parallel: true)
+
+      if drain || delete
+        apply_phase(Phases::ConfigureClient, [master_hosts.first], ssh: true, master: master_hosts.first, parallel: false, optional: true)
+      end
+
+      apply_phase(Phases::Drain, hosts, parallel: false) if drain
+      apply_phase(Phases::DeleteHost, hosts, parallel: false, master: master_hosts.first) if delete
```
jakolehm: Drain and delete can only happen if `hosts` does not include any masters, right? Just thinking if we really need both options.

jakolehm:
@kke conflicts...

kke:
I dropped the host address parameter as it can be defined by using the `-a` from the `filtered_hosts` option set.

PR #717, opened 2018-10-25
More generic terraform json parser

### LINKED ISSUE #715: Cannot pass all host config options from terraform output
For example `ssh_proxy_command`.

kke:
Fixes #715 

Not sure if actually any better than just adding the `:ssh_proxy_command`


kke:
Outdated and bad, should redo / rethink according to discussion in #715

PR #719, opened 2018-10-25
Bastion host

jakolehm:
Proxies ssh & api calls through configured bastion host.

```yaml
bastion:
  address: "<connectable_ip>"
  user: "bastion_user"
  ssh_key_path: "/path/to/ssh.key"
```

also possible to set via terraform output:

```hcl
output "pharos_bastion" {
  value = {
    address = "${module.bastion.public_ip}"
    user    = "bastion_user"
    ssh_key_path = "./ssh_keys/${var.cluster_name}.pem"
  }
}
```


jakolehm:
@kke @jnummelin PTAL

jakolehm:
Added option to specify bastion per host.

lib/pharos/config.rb
```diff
@@ -32,11 +33,14 @@ def self.load(raw_data)
 
       # inject api_endpoint to each host object
       config.hosts.each { |h| h.api_endpoint = config.api&.endpoint }
+      # inject bastion to each host object
+      config.hosts.each { |h| h.configure_bastion(config.bastion) } if config.bastion
```
kke: could be just `h.bastion = config.bastion` inside the hosts enumeration above?

lib/pharos/config_schema.rb
```diff
@@ -59,9 +59,19 @@ def self.messages
...
               optional(:environment).filled
+              optional(:bastion).schema do
+                required(:address).filled(:str?)
+                optional(:user).filled(:str?)
+                optional(:ssh_key_path).filled(:str?)
+              end
             end
           end
         end
+        optional(:bastion).schema do
```
kke: We don't have this sort of "`hosts`-global" settings for any other host feature.

We could make a generic one at some point.

lib/pharos/configuration/bastion.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module Pharos
+  module Configuration
+    class Bastion < Pharos::Configuration::Struct
```
kke: could this actually just be an instance of `Host`?

lib/pharos/terraform/json_parser.rb
```diff
@@ -57,9 +61,12 @@ def parse_host(bundle, host, index)
           role: bundle['role'] || 'worker'
         }
         host[:private_address] = bundle['private_address'][index] if bundle['private_address']
+        host[:private_interface] = bundle['private_interface'][index] if bundle['private_interface']
         host[:labels] = bundle['label'][0] if bundle['label']
+        host[:taints] = bundle['taints'] if bundle['taints']
```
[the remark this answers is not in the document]
jakolehm: yeah, naming should be consistent.

lib/pharos/configuration/bastion.rb
jnummelin: IMO no, at least without heavy refactoring. `Host` has lot of other stuff in it too which we do not need/want at `Bastion`. And having both as `Host` would also make it as never ending recursion :)

lib/pharos/config.rb
jakolehm: Now with one loop.

lib/pharos/config_schema.rb
jakolehm: @jnummelin WDYT?

jnummelin: Would be nice to have global host settings. So maybe move `bastion` into something like:
```
host_config:
  bastion: ...
```

No need to support all the options in this PR, we can later add other PRs to support all host options on global level

jakolehm: Ok, I'm going to remove "global" bastion then from this pr.. let's open a separate issue about `host_config:`.

jakolehm:
@kke @jnummelin PTAL

PR #721, opened 2018-10-25
Allow configuring of tolerations for ingress-nginx daemonset

### LINKED ISSUE #720: Add tolerations to ingress-nginx
Essentially we need a way to add tolerations via cluter.yml to ingress-nginx daemonset to allow it deployed on tainted hosts too

jnummelin:
fixes #720 

Tolerations can be configured using something like:
```
addons:
  ingress-nginx:
    enabled: true
    tolerations:
      - key: "key"
        operator: "Equal"
        value: "value"
        effect: "NoSchedule"
```

### DIFF
--- a/addons/ingress-nginx/resources/daemonset.yml.erb
+++ b/addons/ingress-nginx/resources/daemonset.yml.erb
@@ -25,12 +25,14 @@ spec:
       <%- end -%>
       serviceAccountName: nginx-ingress-serviceaccount
       hostNetwork: true
+      <%- unless config.tolerations.empty? -%>
       tolerations:
       <%- config.tolerations.each do |t| -%>
         -
           <%- t.each do |k,v| -%>
           <%= k %>: <%= v %>
           <%- end -%>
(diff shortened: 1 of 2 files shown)


addons/ingress-nginx/resources/daemonset.yml.erb
```diff
@@ -25,6 +25,13 @@ spec:
       <%- end -%>
       serviceAccountName: nginx-ingress-serviceaccount
       hostNetwork: true
+      tolerations:
```
jakolehm: Probably should not render `tolerations:` at all if not set?

jnummelin: The end result is the same, there's no tolerations. :) But I can guard that too if needed

jakolehm: I think it looks better if guarded.

jnummelin: done

PR #727, opened 2018-10-26
Run the interactive ssh through Net::SSH

kke:
This PR replaces shelling out to system `ssh` with a pure-ruby Net::SSH solution. This way the bastion host stuff and proxy_command etc are used automatically.


jakolehm:
Rubocop says no.

kke:
Now handles screen resizing

PR #734, opened 2018-10-29
add options for file audit configuration

### LINKED ISSUE #733: Enable audit logging by default
Using log backend: https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#log-backend

jnummelin:
fixes #733 

File audit logging can be configured with:
```
audit:
  file:
    path: /var/log/kube_audit/audit.json
    max_size: 100
    max_age: 30
    max_backups: 20
```

This also moves the webhook audit under `audit.webhook` which is a breaking change.


lib/pharos/kubeadm.rb
```diff
@@ -193,6 +194,24 @@ def volume_mounts_for_authentication_token_webhook
         volume_mounts
       end
 
+      def configure_audit_file(config)
+        config['apiServerExtraArgs'].merge!(
+          "audit-log-path" => @config.audit.file.path,
+          "audit-log-maxage" => @config.audit.file.max_age.to_s,
+          "audit-log-maxbackup" => @config.audit.file.max_backups.to_s,
+          "audit-log-maxsize" => @config.audit.file.max_size.to_s,
+          "audit-policy-file" => AUDIT_CFG_DIR + '/policy.yml'
```
jakolehm: `policy.yml` is missing?

jnummelin: nope, it's put into place at https://github.com/kontena/pharos-cluster/blob/feature/audit-logging/lib/pharos/phases/setup_master.rb#L14

jnummelin:
> why not enable this by default?

Audit generates quite a bit of data. E.g. in my cold-running test cluster it generates ~80M per hour. So I thought one should not really have it unless you really want the have audit

> can user configure both file & webhook audit

Yes.

jakolehm:
> Audit generates quite a bit of data. E.g. in my cold-running test cluster it generates ~80M per hour. So I thought one should not really have it unless you really want the have audit

Does it really matter if we have sane defaults for size & max_age? Writing audit log to a file should not be too resource intensive?

jnummelin:
> Does it really matter if we have sane defaults for size & max_age? Writing audit log to a file should not be too resource intensive?

We can make it on by default. How about with something like:
```
          'path' => '/var/log/kubernetes/audit.json',
          'max_size' => 100, # Max 100M files
          'max_age' => 30, # Max 30 days old audits
          'max_backups' => 20 # Max 20 rolled files, each 100M --> 2G used at max
```

jnummelin:
@jakolehm there's now defaulting in place PTAL

lib/pharos/config_schema.rb
```diff
@@ -15,7 +15,15 @@ class ConfigSchema
...
       'pod_security_policy' => {},
       'addon_paths' => [],
-      'container_runtime' => {}
+      'container_runtime' => {},
+      'audit' => {
+        'file' => {
+          'path' => '/var/log/kubernetes/audit.json',
+          'max_size' => 100, # Max 100M files
+          'max_age' => 30, # Max 30 days old audits
+          'max_backups' => 20 # Max 20 rolled files, each 100M
```
jakolehm: Maybe add a comment how much in total audit logs can consume.

PR #735, opened 2018-10-29
Add persistence config option to Lens addon

### LINKED ISSUE #730: kontena-lens: redis should use persistent volume if available

nevalla:
This PR add `persistence` option to Kontena Lens addon. If it is enabled, addon will create persistent volume claim for redis deployment using default storage class:

```
addons:
  kontena-lens:
    enabled: true
    persistence:
      enabled: true
```

This requires that default storage class is created before lens is deployed. Otherwise lens dashboard won't get up and running. To ensure correct order of add-ons deployments, this PR adds `priority` class method to `Pharos::Addon` class, and that is used to sort add-ons before deploying.

In future we can add `storage_class` option to config options so user could define other storage class than default.

Fixes #730


non-oss/pharos_pro/addons/kontena-lens/addon.rb
```diff
@@ -3,6 +3,7 @@
 Pharos.addon 'kontena-lens' do
   version '1.1.0'
   license 'Kontena License'
+  priority 10
```
jakolehm: What is the default priority?

nevalla: priority method returns `@priority.to_i`, by default it is 0 (`nil.to_i`). Then the order is the order where addons are introduced in `cluster.yml`

PR #737, opened 2018-10-30
Allow to pass default value to confirm_yes! method

nevalla:


lib/pharos/command_options/yes.rb
```diff
@@ -9,14 +9,14 @@ def self.included(base)
       end
 
       module InstanceMethods
-        def confirm_yes!(message)
+        def confirm_yes!(message, default = true)
```
kke: maybe make it a keyword argument?

nevalla: changed

PR #739, opened 2018-10-31
Validate localhost dns resolve

### LINKED ISSUE #738: Validate localhost dns resolving
We should validate that localhost actually resolves to the same machine.

jakolehm:
Fixes #738

### DIFF
--- a/lib/pharos/phases/validate_host.rb
+++ b/lib/pharos/phases/validate_host.rb
@@ -16,6 +16,8 @@ module Pharos
         validate_unique_hostnames
         logger.info { "Validating host routes ..." }
         validate_routes
+        logger.info { "Validating localhost dns resolve ..." }
+        validate_localhost_resolve
       end
 
       def check_distro_version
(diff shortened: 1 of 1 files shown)


lib/pharos/phases/validate_host.rb
```diff
@@ -46,6 +48,11 @@ def validate_unique_hostnames
         raise Pharos::InvalidHostError, "Duplicate hostname #{@host.hostname} for hosts #{duplicates.map(&:address).join(',')}"
       end
 
+      def validate_localhost_resolve
+        localhost_ip = @ssh.exec!("ping -c 1 localhost | grep -Eo -m 1 '[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}';")
```
kke: Could be `getent ahostsv4 localhost | grep STREAM | head -n 1 | cut -d ' ' -f1` in case there's an old version of grep

lib/pharos/phases/validate_host.rb
```diff
@@ -46,6 +48,11 @@ def validate_unique_hostnames
...
+        raise Pharos::InvalidHostError, "Localhost ip does not look right: #{localhost_ip}" unless localhost_ip.start_with?('127.')
```
kke: Could be `IPAddr.new(localhost_ip).loopback?` (which needs `using Pharos::CoreExt::IPAddrLoopback if RUBY_VERSION < '2.5.0'`)

lib/pharos/phases/validate_host.rb
jakolehm: Tested with CentOS7 which probably is the oldest.

PR #741, opened 2018-10-31
[RFC] Add global: keyword for global host settings

kke:
Reduce config repetition by adding a `global:` keyword that applies to all hosts.

If a host redefines a key, it will override the global.

TODO:
- should maybe do a deep merge to allow for example to override something inside a nested hash instead of replacing the whole hash.


kke:
No interest

PR #755, opened 2018-11-02
Don't validate cluster version if api down

jakolehm:


lib/pharos/phases/validate_version.rb
```diff
@@ -9,6 +9,10 @@ class ValidateVersion < Pharos::Phase
 
       def call
         return unless kubeconfig?
+        if @host.master_sort_score.positive?
+          logger.warn { 'Master seems unhealthy, cannot detect cluster version.' }
```
kke:
```suggestion
          logger.warn { "Master seems unhealthy, can't detect cluster version." }
```

PR #765, opened 2018-11-04
Fix upgrade from previous versions

jakolehm:
Upgrade from kubernetes 1.11.3 to 1.11.4 hangs when done via `kubeadm upgrade`. This PR changes the upgrade behaviour in following ways:

- no need to do full upgrade if major.minor is same
- no need to run reconfigure if previous phase has already configured control-plane

### DIFF
--- a/lib/pharos/phases/upgrade_master.rb
+++ b/lib/pharos/phases/upgrade_master.rb
@@ -12,9 +12,19 @@ module Pharos
       def upgrade?
         file = @ssh.file('/etc/kubernetes/manifests/kube-apiserver.yaml')
         return false unless file.exist?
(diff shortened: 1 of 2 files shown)


lib/pharos/phases/upgrade_master.rb
```diff
@@ -12,11 +12,21 @@ def kubeadm
...
+        new_major_minor = parse_major_minor(Pharos::KUBE_VERSION)
+        return false if current_major_minor == new_major_minor
 
         true
       end
 
+      # @param version [String]
+      # @return [String]
+      def parse_major_minor(version)
+        version.split('.')[0...2].join('.')
```
kke:
```suggestion
        version[/(\d+\.\d+)/, 1]
```

PR #768, opened 2018-11-05
Region aware peer addresses

jakolehm:
- Resolves used peer address (etcd, master, weave) from `failure-domain.beta.kubernetes.io/region` label.
- adds [Flying Shuttle](https://github.com/kontena/weave-flying-shuttle) sidecar to Weave Net, allowing dynamic peer configuration based on node region/external-ip labels

### DIFF
--- a/lib/pharos/config.rb
+++ b/lib/pharos/config.rb
@@ -99,12 +99,12 @@ module Pharos
 
     # @return [Array<String>]
     def etcd_regions
-      @etcd_regions ||= etcd_hosts.map(:&region).compact.uniq
+      @etcd_regions ||= etcd_hosts.map(&:region).compact.uniq
     end
 
     # @return [Array<String>]
     def regions
-      @regions ||= hosts.map(:&region).compact.uniq
+      @regions ||= hosts.map(&:region).compact.uniq
     end
 
     # @param kubeconfig [Hash]
--- a/lib/pharos/config.rb
+++ b/lib/pharos/config.rb
@@ -66,7 +66,7 @@ module Pharos
 
     # @return [Array<Pharos::Configuration::Node>]
     def master_hosts
-      @master_hosts ||= hosts.select { |h| h.role == 'master' }
+      @master_hosts ||= hosts.select { |h| h.role == 'master' }.sort_by(&:master_sort_score)
     end
 
     # @return [Pharos::Configuration::Node]
@@ -85,10 +85,21 @@ module Pharos
 
       etcd_hosts = hosts.select { |h| h.role == 'etcd' }
       if etcd_hosts.empty?
-        master_hosts
+        master_hosts.sort_by(&:etcd_sort_score)
       else
-        etcd_hosts
+        etcd_hosts..sort_by(&:etcd_sort_score)
       end
+    end
+
+    # @param peer [Pharos::Configuration::Host]
+    # @return [String]
+    def etcd_peer_address(peer)
+      etcd_regions.size > 1 ? peer.address : peer.peer_address
+    end
+
+    # @return [Array<String>]
+    def etcd_regions
+      @regions ||= etcd_hosts.map { |host| host.region }.compact.uniq
     end
 
     # @param kubeconfig [Hash]
--- a/lib/pharos/config.rb
+++ b/lib/pharos/config.rb
@@ -99,7 +99,12 @@ module Pharos
 
     # @return [Array<String>]
     def etcd_regions
-      @regions ||= etcd_hosts.map { |host| host.region }.compact.uniq
+      @etcd_regions ||= etcd_hosts.map { |host| host.region }.compact.uniq
+    end
+
+    # @return [Array<String>]
+    def regions
+      @regions ||= hosts.map { |host| host.region }.compact.uniq
     end
 
     # @param kubeconfig [Hash]
--- a/lib/pharos/config.rb
+++ b/lib/pharos/config.rb
@@ -87,7 +87,7 @@ module Pharos
       if etcd_hosts.empty?
         master_hosts.sort_by(&:etcd_sort_score)
       else
-        etcd_hosts..sort_by(&:etcd_sort_score)
+        etcd_hosts.sort_by(&:etcd_sort_score)
       end
     end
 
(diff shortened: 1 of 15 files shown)


jakolehm:
@jnummelin PTAL .. I'll write specs after we agree that this can work.

jnummelin:
In general the logic looks pretty good. We just need to test different scenarios.

I was thinking the other day that maybe this could be as an "experimental" feature behind some flag/env, but looks like there's so many places effected that it might be hard to do.

jakolehm:
@jnummelin PTAL (again)

lib/pharos/phases/label_node.rb
```diff
@@ -15,30 +15,26 @@ def call
...
       # @param node [K8s::Resource]
       def patch_labels(node)
-        kube_nodes.update_resource(
-          node.merge(
-            metadata: {
-              labels: @host.labels
-            }
-          )
-        )
+        kube_nodes.merge_patch(node.metadata.name, {
```
jakolehm: Not sure why I did this... maybe I should revert these back to `update_resource`?

jnummelin: 👍

lib/pharos/config.rb
```diff
@@ -66,7 +66,7 @@ def dns_replicas
 
     # @return [Array<Pharos::Configuration::Node>]
     def master_hosts
-      @master_hosts ||= hosts.select { |h| h.role == 'master' }
+      @master_hosts ||= hosts.select { |h| h.role == 'master' }.sort_by(&:master_sort_score)
```
kke: This will only sort once and won't change even if sort scores change. I think it wouldn't matter if it sorted every time without memoization (or the `checks` could invalidate the cache on change, but that's a bit premature optimization)

kke: I made almost the same changes in #855 but kept the memoizing `master_hosts` and added `sorted_master_hosts` with sorting that does not memoize.

jnummelin: I don't see why the sort order would change within `pharos up` run? AFAIK it's safe to memoize

kke: There's no score until gather_facts is run

lib/pharos/resources/weave/daemon-set.yml.erb
```diff
@@ -82,6 +82,17 @@ spec:
           volumeMounts:
             - name: xtables-lock
               mountPath: /run/xtables.lock
+        - name: weave-flying-shuttle
+          env:
+            - name: HOSTNAME
```
jnummelin: uh, sounds bad idea to use `HOSTNAME`. wouldn't that "override" the hostname seen in the container?

lib/pharos/config.rb
jakolehm: Reason for this refactoring is that we picked the wrong master in some cases (sorting was done in ClusterManager). At least now the logic is in one place and can be easily tweaked (if needed).

lib/pharos/configuration/host.rb
```diff
@@ -93,16 +93,44 @@ def api_address
...
+        labels['node-address.kontena.io/external-ip'] = address
+        labels['node-role.kubernetes.io/worker'] = '' if worker?
+
+        labels
+      end
 
-        @attributes[:labels] || { 'node-role.kubernetes.io/worker': "" }
+      # @return [Hash]
+      def checks
+        @checks ||= {}
```
kke: I think you'll need to change `GatherFacts` `@host.checks = host_checks` to  `@host.checks.replace(host_checks)`

jakolehm: I don't think so... because of `attr_accessor :checks` .

lib/pharos/phases/configure_weave.rb
```diff
@@ -6,12 +6,18 @@ class ConfigureWeave < Pharos::Phase
...
       WEAVE_VERSION = '2.5.0'
+      WEAVE_FLYING_SHUTTLE_VERSION = '0.1.1'
 
       register_component(
         name: 'weave-net', version: WEAVE_VERSION, license: 'Apache License 2.0',
         enabled: proc { |c| c.network.provider == 'weave' }
       )
 
+      register_component(
+        name: 'weave-flying-shuttle', version: WEAVE_VERSION, license: 'Apache License 2.0',
```
jnummelin:
```suggestion
        name: 'weave-flying-shuttle', version: WEAVE_FLYING_SHUTTLE_VERSION, license: 'Apache License 2.0',
```

lib/pharos/config.rb
kke: This will never select the healthiest master since the sorted list is memoized before checks are made, so the sorting has no effect.

lib/pharos/resources/weave/daemon-set.yml.erb
jakolehm: See the other containers in this pod...

lib/pharos/configuration/host.rb
kke: Right. I removed the accessor since I added the reader manually.

lib/pharos/resources/weave/daemon-set.yml.erb
jnummelin: right, and it's in host network anyway.

lib/pharos/config.rb
jakolehm: That's a bit unfortunate (checked and yes it is memoized before gather facts) 😄

jnummelin: Oh, I though the checks would've made earlier. In that case I revert my statement, we shouldn't memoize :D

jakolehm: Fixed.

lib/pharos/config.rb
```diff
@@ -66,12 +66,12 @@ def dns_replicas
 
     # @return [Array<Pharos::Configuration::Node>]
     def master_hosts
-      @master_hosts ||= hosts.select { |h| h.role == 'master' }
+      hosts.select { |h| h.role == 'master' }.sort_by(&:master_sort_score)
```
kke:
```suggestion
      (@master_hosts ||= hosts.select { |h| h.role == 'master' }).sort_by(&:master_sort_score)
```

this would memoize the role selection part :)

jakolehm: Too magical, hard to read 😄

PR #784, opened 2018-11-08
Allow to use Kontena Lens with custom host without tls option

nevalla:


non-oss/pharos_pro/addons/kontena-lens/addon.rb
```diff
@@ -36,14 +36,15 @@
       email: config.tls&.email,
       user_management: user_management_enabled?
     )
-    wait_for_dashboard(host)
+    protocol = config.host && config.tls&.email ? 'http' : 'https' # with cert-manager we have to use http since tls secret is not in place immediately
```
jakolehm: Wat? Really hard to read this line... does this mean that `protocol` is `nil` if host is not set?
```suggestion
    protocol = config.tls&.email ? 'http' : 'https' if config.host # with cert-manager we have to use http since tls secret is not in place immediately
```

nevalla: if `config.host` and `config.tls.email` is defined then protocol is`http` , otherwise `https`. 
With your suggestion protocol is set to nil unless `config.host` is defined, which is wrong.

Now thinking that actually we need only check if `config.tls.email` is set, then use `http`, otherwise `https`.

jakolehm:
```suggestion
    protocol = (config.host && config.tls&.email) ? 'http' : 'https' # with cert-manager we have to use http since tls secret is not in place immediately
```

nevalla: rubocop is not happy with that

jakolehm: Ok, rubocop is the law but it looks so wrong to my eyes... maybe too little ☕️ .

nevalla: I will change that to check only `config.tls&.email`. I will also tweak ingress and certificate resources a bit. With these tweaks user is able to fetch valid cert for `lens.<worker-node-ip>.nip.io` too, which is nice.

nevalla:
@jakolehm Added some changes, PTAL

non-oss/pharos_pro/addons/kontena-lens/resources/12-ingress.yml.erb
```diff
@@ -10,7 +10,7 @@ spec:
   tls:
     - hosts:
       - <%= host %>
-      <% if config.host %>
+      <% if email %>
```
jakolehm: Should we skip whole `tls:` if no email?

nevalla: I assume `https` is not possible with ingress default certificate then

PR #786, opened 2018-11-08
Fix kontena-storage placement settings

jakolehm:


lib/pharos/core-ext/deep_transform_keys.rb
```diff
@@ -10,7 +10,7 @@ def self.deep_transform_keys(value = nil, &block)
         when Array
           value.map { |v| deep_transform_keys(v, &block) }
         when Hash
-          Hash[value.map { |k, v| [yield(k), deep_transform_keys(v, &block)] }]
+          Hash[value.map { |k, v| [yield(k.dup), deep_transform_keys(v, &block)] }]
```
jakolehm: Key seems to be frozen if `:hash?` is used with dry-struct.

PR #787, opened 2018-11-08
Reduce unnecessary extending in StringCasing refinement

kke:
Refinements are strange as is. Trying to make a refinement that can also work as an extendable module is not a working solution. The methods inside the module do not seem to be in the scope of the `refine`.

This PR changes the `Pharos::CoreExt::StringCasing` and `DeepTransformKeys` to be regular refinements, not to be used with `extend` at all.

### DIFF
--- a/lib/pharos/core-ext/deep_transform_keys.rb
+++ b/lib/pharos/core-ext/deep_transform_keys.rb
@@ -10,30 +10,36 @@ module Pharos
         when Array
           value.map { |v| deep_transform_keys(v, &block) }
         when Hash
-          Hash[value.map { |k, v| [yield(k.dup), deep_transform_keys(v, &block)] }]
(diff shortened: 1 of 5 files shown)


lib/pharos/core-ext/deep_transform_keys.rb
```diff
@@ -10,30 +10,36 @@ def self.deep_transform_keys(value = nil, &block)
         when Array
           value.map { |v| deep_transform_keys(v, &block) }
         when Hash
-          Hash[value.map { |k, v| [yield(k.dup), deep_transform_keys(v, &block)] }]
+          Hash[value.map { |k, v| [yield(k.frozen? ? k.dup : k), deep_transform_keys(v, &block)] }]
```
kke: actually not sure if this is needed anymore as there's no extending going on.

jakolehm:
@kke please rebase.

kke:
Rebased

kke:
Fixed in #1091

PR #796, opened 2018-11-11
Add Lens tls.enabled option

Timer:
Hi! I'm not a Ruby developer, but I gave this a whirl.

Motivation: Ingress into our cluster passes through a load balancer that handles https off-loading / termination (before reaching Traefik).

I'm not familiar with how to test this, but it should be most of the way there. :-)


nevalla:
Good job, considering you are not a Ruby developer! Thanks

Couple of changes I would propose:

How about if introducing method `tls_enabled?` instead of `set_defaults`
```ruby
def tls_enabled?
  config.tls&.enabled != false
end
```

It can be passed to manifest templates on `apply_resources`:
```ruby
apply_resources(
  host: host,
  email: config.tls&.email,
  tls_enabled: tls_enabled?,
  user_management: user_management_enabled?
)
```
and resolving protocol would be more straight-forward too:
```ruby
protocol =  !tls_enabled? || config.tls&.email ? 'http' : 'https' # with cert-manager we have to use http since tls secret is not in place immediately
```

And in manifest files it would be more readable to, for example (`12-ingress.yml.erb`):
```erb
<%- if tls_enabled -%>
  tls:
    - hosts:
      - <%= host %>
      <% if email %>
      secretName: kontena-lens-tls
      <% end %>
<%- end -%>
```
**Note:** `<%- #code -%>` ignores new lines after the ruby block, which is handy when manipulating YAML files.

To test modifications and running `pharos up` command from the source code, you can use:

```sh
PHAROS_NON_OSS=true bundle exec ./bin/pharos-cluster up
```
(you must execute `bundle install` before first run)

OR just ping me and I can test this out.

Timer:
Awesome, thanks! Tested and working. 😄

Timer:
Please let me know what I can do to help get this in faster.

I can't see the build failure reason, is there something I need to address? Thanks!

jakolehm:
@Timer needs conflict resolving.

nevalla:
Now since we are not polling dashboard readiness anymore I think we can just use protocol in post-install message:
```
    protocol = !tls_enabled? 'http' : 'https'
    message = "Kontena Lens is running at: " + pastel.cyan("#{protocol}://#{host}")
```

non-oss/pharos_pro/addons/kontena-lens/addon.rb
```diff
@@ -36,9 +37,11 @@
     apply_resources(
       host: host,
       email: config.tls&.email,
+      tls_enabled: tls_enabled?,
       user_management: user_management_enabled?
     )
-    message = "Kontena Lens is configured to respond at: " + pastel.cyan("https://#{host}")
+    protocol = !tls_enabled? 'http' : 'https'
```
Timer: Should we flip this?

```suggestion
    protocol = tls_enabled? 'https' : 'http'
```

nevalla: yes...better :+1:

Timer: Actually, does this need double questionmarks? I resolved this in GH UI.

```suggestion
    protocol = tls_enabled? ? 'https' : 'http'
```

Trying in terminal now...

Timer: Yeah, it needs the latter suggestion.

Timer:
AFAIK this is ready to go, correct?

nevalla:
Yes, looking good!

PR #749, opened 2018-11-01
Downgrade cri-o to 1.11.6

jakolehm:
Downgrades cri-o to v1.11.6.

Downgrade revealed interesting side-effects.. that's why this PR also improves container runtime upgrade/downgrade by making it more safe:
- drain/uncordon node if master is healthy
- allow only new nodes to be configured parallel

Fixes #746

PR #798, opened 2018-11-13
Remove Pharos::SSH::Manager in favor of simpler host.ssh accessor

kke:
A smaller slice of #642
In continuation with the singleton-modification in #749

Adds `host.ssh` which eliminates the need for the `Pharos::SSH::Manager`.


kke:
Most of the changes are just `@ssh` -> `ssh` or `Pharos::SSH::Manager.instance.client_for(host)` -> `host.ssh` and a whole bunch of spec adaptations.

kke:
There is a chance that some @ssh / SSH::Manager usages in `master` sneak in from merge, I'll check that first.

kke:
Seems ok and `pharos up` went through.

PR #802, opened 2018-11-13
Refactor output colorization

kke:
Moves Pastel from being initialized in several places and passed around as method arguments into a central string refinement.

`puts pastel.green("hello")` becomes:

```ruby
using Pharos::CoreExt::Colorize

puts "hello".green
```

Similar to the old school `colorize` gem.

To disable colorization globally, just do `Pharos::CoreExt::Colorize.disable!`

This helps in refactoring as there's one less thing to pass around, like `ClusterManager.new(pastel: @pastel)`


jakolehm:
@kke conflicts...

kke:
Up to date now

kke:
Up to date again

PR #805, opened 2018-11-14
Run all phases in parallel

kke:
Removes the `parallel: [true|false]` option from phases. 

It's the responsibility of the phase to know when to run exclusively, not cluster manager's.

Also, most of the `parallel: false` phases were run on a single host, so `parallel: false` did essentially nothing apart from not creating a thread for the runner.

The parts of phases that should not be run parallelly can utilize the phase mutex brought in by #749

### DIFF
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -58,7 +58,7 @@ module Pharos
 
     def gather_facts
       apply_phase(Phases::GatherFacts, config.hosts)
-      apply_phase(Phases::ConfigureClient, [config.master_host], master: sorted_master_hosts.first, optional: true)
+      apply_phase(Phases::ConfigureClient, [config.master_host], master: config.master_host, optional: true)
     end
 
     def validate
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -57,41 +57,43 @@ module Pharos
     end
 
     def gather_facts
-      apply_phase(Phases::GatherFacts, config.hosts, parallel: true)
-      apply_phase(Phases::ConfigureClient, [config.master_host], master: config.master_host, parallel: false, optional: true)
+      apply_phase(Phases::GatherFacts, config.hosts)
+      apply_phase(Phases::ConfigureClient, [config.master_host], master: sorted_master_hosts.first, optional: true)
     end
 
     def validate
       apply_phase(Phases::UpgradeCheck, %w(localhost))
       addon_manager.validate
       gather_facts
-      apply_phase(Phases::ValidateHost, config.hosts, parallel: true)
-      apply_phase(Phases::ValidateVersion, [config.master_host], master: config.master_host, parallel: false)
+      apply_phase(Phases::ValidateHost, config.hosts)
+      apply_phase(Phases::ValidateVersion, [config.master_host], master: config.master_host)
     end
 
     def apply_phases
       master_hosts = config.master_hosts
-      apply_phase(Phases::MigrateMaster, master_hosts, parallel: true)
-      apply_phase(Phases::ConfigureHost, config.hosts, master: master_hosts.first, parallel: true)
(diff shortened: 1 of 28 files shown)


kke:
The label nodes phase is now run on master only and it goes through `@config.hosts.each`, it could be somehow optimized further by getting all nodes at once or labeling them all in a single/batch request.

jakolehm:
Please rebase so that we can see travis e2e results.

kke:
Seems to work now

lib/pharos/cluster_manager.rb
```diff
@@ -85,29 +85,31 @@ def apply_phases
...
+        apply_phase(Phases::ConfigureCfssl, etcd_hosts)
+        apply_phase(Phases::ConfigureEtcdCa, [etcd_hosts.first])
+        apply_phase(Phases::ConfigureEtcdChanges, [etcd_hosts.first])
+        apply_phase(Phases::ConfigureEtcd, etcd_hosts)
       end
 
-      apply_phase(Phases::ConfigureSecretsEncryption, master_hosts, parallel: false)
-      apply_phase(Phases::SetupMaster, master_hosts, parallel: true)
-      apply_phase(Phases::UpgradeMaster, master_hosts, master: master_hosts.first, parallel: false) # requires optional early ConfigureClient
+      apply_phase(Phases::ValidateSecretsEncryption, master_hosts)
```
kke: The split is here. Validate finds and collects the first encryption config and writes the plain content to cluster_context

lib/pharos/cluster_manager.rb
```diff
@@ -85,29 +85,31 @@ def apply_phases
...
+      apply_phase(Phases::GenerateSecretsEncryptionKeys, ['localhost']) unless context['secrets_encryption']
```
kke: This runs only if `cluster_context` does not have a config. Generates the new config locally into cluster_context.

lib/pharos/cluster_manager.rb
```diff
@@ -85,29 +85,31 @@ def apply_phases
...
+      apply_phase(Phases::ConfigureSecretsEncryption, master_hosts)
```
kke: And this writes the config from cluster_context to all hosts

lib/pharos/cluster_manager.rb
```diff
@@ -119,8 +121,8 @@ def apply_phases
 
       apply_phase(Phases::ConfigureBootstrap, [master_hosts.first]) # using `kubeadm token`, not the kube API
 
-      apply_phase(Phases::JoinNode, config.worker_hosts, parallel: true)
-      apply_phase(Phases::LabelNode, config.hosts, master: master_hosts.first, parallel: false) # NOTE: uses the @master kube API for each node, not threadsafe
+      apply_phase(Phases::JoinNode, config.worker_hosts)
+      apply_phase(Phases::LabelNode, [master_hosts.first], master: master_hosts.first)
```
kke: Labelnode now runs only on master instead of every host separately and goes through `@config.hosts` to get labels / taints

lib/pharos/phases/configure_kube_proxy.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+module Pharos
+  module Phases
+    class ConfigureKubeProxy < Pharos::Phase
```
jakolehm: Not needed anymore.. probably leftover from rebase?

kke: Apparently. Removed.

kke:
Up to date again

lib/pharos/phases/label_node.rb
```diff
@@ -59,7 +66,7 @@ def find_node
       end
 
       def kube_nodes
-        kube_client.api('v1').resource('nodes')
+        @kube_nodes ||= kube_client.api('v1').resource('nodes')
```
kke: I assume there aren't going to be new nodes during the phase, so memoizing here.

kke:
Too far out of sync.

PR #806, opened 2018-11-14
Correct replica location

Timer:
Currently, `.spec.spec.replicas` has no effect (it should be `.spec.replicas`):
image

### DIFF
--- a/non-oss/pharos_pro/addons/kontena-lens/resources/04-user-management-deployment.yml.erb
+++ b/non-oss/pharos_pro/addons/kontena-lens/resources/04-user-management-deployment.yml.erb
@@ -6,6 +6,7 @@ metadata:
   labels:
     app: user-management
 spec:
+  replicas: 2
   selector:
     matchLabels:
       app: user-management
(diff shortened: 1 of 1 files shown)


jakolehm:
@nevalla PTAL

nevalla:
Yes, this is a good catch. This will fix the issue for fresh clusters. However, I tried to update existing cluster and Kubernetes API returned error:
```
got error (K8s::Error::InternalError): PATCH /apis/apps/v1/namespaces/kontena-lens/deployments/user-management => HTTP 500 Internal Server Error: Unable to remove nonexistent key: replicas
```
I think we have to give some extra love to handle updates as well.

jakolehm:
@jnummelin wdyt, does new k8s-client handle that patch error?

jnummelin:
> wdyt, does new k8s-client handle that patch error?

Nope. The problem is that it's already in the `last-used-config` annotation so the calculated patch will contain a rm operation for it. It needs to be taken out of that first.

Timer:
I'm not entirely sure how to move this forward. Would some derived changes from https://github.com/kontena/k8s-client/pull/58 make this work?

If field is present in last applied but absent from current resource, don't bother to patch with the removal.

jakolehm:
@jnummelin any ideas how to handle this error?

jnummelin:
One option is to first patch the annotation itself to remove the problematic paths from it and the re-shoot the stack.

jnummelin:
@Timer Did you try the approach I outlined in my previous comment?

Timer:
Replaced by #865

PR #808, opened 2018-11-16
Fix kontena-lens create-config success check

jakolehm:
Old behaviour fails if curl is http/2 capable.

### DIFF
--- a/non-oss/pharos_pro/addons/kontena-lens/addon.rb
+++ b/non-oss/pharos_pro/addons/kontena-lens/addon.rb
@@ -54,14 +54,17 @@ Pharos.addon 'kontena-lens' do
     @pastel ||= Pastel.new
   end
 
+  # @return [Pharos::Configuration::Host]
   def gateway_node
     cluster_config.worker_hosts.first || cluster_config.master_hosts.first
   end
 
+  # @return [String, NilClass]
   def gateway_node_ip
     gateway_node&.address
   end
 
+  # @return [String, NilClass]
   def master_host_ip
     cluster_config.master_host&.address
   end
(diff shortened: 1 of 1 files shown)


nevalla:
Initially this was implemented this way, but if curl request returned redirect (http -> https) response this failed. However, the implementation has changed after that and now we use correct protocol on curl request so I think this will work now. I will test with different config option combinations to be sure.

non-oss/pharos_pro/addons/kontena-lens/addon.rb
```diff
@@ -105,21 +112,27 @@ def lens_configured?
...
   def create_lens_config(name, protocol, host, admin_password)
     cluster_config = {
       clusterName: name,
       clusterUrl: "https://#{master_host_ip}:6443",
       adminPassword: admin_password
     }
-    command = "sudo curl -iksL -X POST -d '#{cluster_config.to_json}' -H \"Host: #{host}\" -H \"Content-Type: application/json\" #{protocol}://localhost/api/cluster"
-    result = ssh.exec(command)
-    raise Pharos::InvalidAddonError, "Could not create Kontena Lens configuration" unless result.output.lines.include?("HTTP/1.1 200 OK\r\n")
+    command = "sudo curl -iksL -o /dev/null -w \"%{http_code}\" -X POST -d '#{cluster_config.to_json}' -H \"Host: #{host}\" -H \"Content-Type: application/json\" #{protocol}://localhost/api/cluster" # rubocop:disable Style/FormatStringToken
```
kke: I guess the sudo is unnecessary

jakolehm: Yeah, removed.

PR #812, opened 2018-11-16
Use full hostname on vsphere

### LINKED ISSUE #809: vSphere cloud provider behavior
Per a bug in the documentation, vSphere requires the node name to match the OS's hostname.

See PR: https://github.com/kubernetes/website/pull/11017.

x-ref: https://github.com/kubernetes/kubernetes/issues/54482

Existing logic exists to handle this for `aws`, and the easiest mitigation path would be to support `vsphere` in this way as well:

```
      def hostname
        cloud_provider = @config.cloud&.provider
        if cloud_provider == 'aws'
          @ssh.exec!('hostname -f').strip
        else
          @ssh.exec!('hostname -s').strip
        end.downcase
      end
```

=>

```
      def hostname
        cloud_provider = @config.cloud&.provider
        if cloud_provider == 'aws' || cloud_provider == 'vsphere'
          @ssh.exec!('hostname -f').strip
        else
          @ssh.exec!('hostname -s').strip
        end.downcase
      end
```

jakolehm:
Fixes #809

### DIFF
--- a/lib/pharos/phases/gather_facts.rb
+++ b/lib/pharos/phases/gather_facts.rb
@@ -9,6 +9,8 @@ module Pharos
 
       title "Gather host facts"
 
+      FULL_HOSTNAME_CLOUD_PROVIDERS = %w(aws vsphere).freeze
+
       def call
         logger.info { "Checking sudo access ..." }
         check_sudo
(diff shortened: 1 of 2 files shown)


kke:
Is there some reason we're not using `-f` for everyone?

jakolehm:
> Is there some reason we're not using -f for everyone?

Other providers want short version... minority needs `-f`.

PR #819, opened 2018-11-19
[WIP/POC/RFC] Hide sensitive data in debug logs

kke:
Adds `hide:` option to ssh commands to hide sensitive data from debug logs

TODO: use where needed, currently only in `ValidateVersion`, which now looks like:

image


kke:
Didn't catch any interest

PR #822, opened 2018-11-20
Store kontena-lens config directly to the k8s api

jakolehm:
No more "Waiting for Kontena Lens to get up and running ...".

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,8 @@
 PATH
   remote: .
   specs:
-    pharos-cluster (2.0.0.rc.4)
+    pharos-cluster (2.0.2)
+      bcrypt
       bcrypt_pbkdf (>= 1.0, < 2.0)
       clamp (= 1.2.1)
       dry-struct (= 0.5.0)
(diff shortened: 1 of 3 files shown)


jakolehm:
@nevalla PTAL

nevalla:
The implementation is now 💯 times better. Only downside is that now pharos spit out lens address and credentials but dasboard is not there yet. Is that good or should we wait, for example, that dashboard pod is running or similar?

jakolehm:
I think we could just hint that deployment of the dashboard might take a while?

jnummelin:
Maybe have a hint like "Starting up Lens the first time might take couple of minutes, until that you'll see 503 with the address given above."

503 normally should be transient, but the user does not have any idea how long to wait for it to pass.

jakolehm:
Improved post-install messages.

jnummelin:
LGTM

kke:
```
/__enclose_io_memfs__/lib/ruby/2.5.0/rubygems/core_ext/kernel_require.rb:59:in `require': cannot load such file -- bcrypt_ext (LoadError)
```

PR #826, opened 2018-11-22
Improve reset logic

### LINKED ISSUE #825: Reset hangs if mounted persistent volumes
**What happened**:

`pharos reset` hangs if cluster has mounted persistent volumes (for example via kontena-storage).

`pharos reset` should not hang.

Create a cluster with kontena-lens (persistence enabled) + kontena-storage and try to reset.

jakolehm:
### Changes 

- drain/delete nodes always if kubernetes api is healthy
- remove kontena-storage data_dir if enabled
- kontena-storage & openebs addons now cleanup datapath on reset

Fixes #825

### DIFF
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -127,21 +127,14 @@ module Pharos
       apply_phase(Phases::ConfigureTelemetry, [master_hosts.first], master: master_hosts.first)
     end
 
-    def apply_reset_hosts(hosts, drain: true, delete: true)
+    # @param hosts [Array<Pharos::Configuration::Host>]
+    def apply_reset_hosts(hosts)
       master_hosts = sorted_master_hosts
-      apply_phase(Phases::GatherFacts, hosts, parallel: true)
-
-      if drain || delete
-        apply_phase(Phases::ConfigureClient, [master_hosts.first], master: master_hosts.first, parallel: false, optional: true)
+      if master_hosts.first.master_sort_score == 0
+        apply_phase(Phases::Drain, hosts, parallel: false)
+        apply_phase(Phases::DeleteHost, hosts, parallel: false, master: master_hosts.first)
       end
-
-      apply_phase(Phases::Drain, hosts, parallel: false) if drain
-      apply_phase(Phases::DeleteHost, hosts, parallel: false, master: master_hosts.first) if delete
       apply_phase(Phases::ResetHost, hosts, parallel: true)
-    end
-
-    def apply_reset_all
-      apply_phase(Phases::ResetHost, config.hosts, parallel: true)
     end
 
     def apply_addons_cluster_config_modifications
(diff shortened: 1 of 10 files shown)


lib/pharos/phases/reset_host.rb
```diff
@@ -8,6 +8,10 @@ class ResetHost < Pharos::Phase
       def call
         logger.info { "Removing all traces of Kontena Pharos ..." }
         host_configurer.reset
+        if @config.addons.dig('kontena-lens', 'enabled')
```
jakolehm: We probably should have `reset` in addon DSL. Does not make sense to hard-code addon reset logic here.

lib/pharos/cluster_manager.rb
```diff
@@ -133,23 +133,16 @@ def apply_phases
...
-    def apply_reset_hosts(hosts, drain: true, delete: true)
+    # @param hosts [Array<Pharos::Configuration::Host>]
+    def apply_reset_hosts(hosts)
       master_hosts = sorted_master_hosts
-      apply_phase(Phases::GatherFacts, hosts, ssh: true, parallel: true)
-
-      if drain || delete
-        apply_phase(Phases::ConfigureClient, [master_hosts.first], ssh: true, master: master_hosts.first, parallel: false, optional: true)
+      if master_hosts.first.master_sort_score == 0
+        apply_phase(Phases::Drain, hosts, parallel: false)
```
jakolehm: We need to do this by default because cluster hosts will be left in a weird state without draining (if storage is in use).

kke: But why when master sort score is 0? Drain and delete use the kube api, 0 implies api is not working.

lib/pharos/phases/reset_host.rb
```diff
@@ -8,6 +8,10 @@ class ResetHost < Pharos::Phase
...
+          data_dir = @config.addons.dig('kontena-lens', 'data_dir').strip
```
kke: NoMethodError strip for NilClass? I don't see a data_dir attribute in the schema. Should this be 'kontena-storage'?

jakolehm: Yeah, this is wrong.. going to refactor this to addons.

lib/pharos/cluster_manager.rb
```diff
@@ -133,23 +133,16 @@ def apply_phases
...
 
-    def apply_reset_hosts(hosts, drain: true, delete: true)
+    # @param hosts [Array<Pharos::Configuration::Host>]
+    def apply_reset_hosts(hosts)
       master_hosts = sorted_master_hosts
-      apply_phase(Phases::GatherFacts, hosts, ssh: true, parallel: true)
-
-      if drain || delete
-        apply_phase(Phases::ConfigureClient, [master_hosts.first], ssh: true, master: master_hosts.first, parallel: false, optional: true)
+      if master_hosts.first.master_sort_score == 0
```
kke: `master_hosts.first.checks['api_healthy']` (or a helper method) could be clearer.

jakolehm:
@kke PTAL

PR #828, opened 2018-11-22
Add config option to pass skip_refresh option to shell image

nevalla:
Airgapped installations do not have public internet and by default `helm init -c` will cache stable repo from internet. This PR allows to define: 
```
shell: 
  skip_refresh=true
```

This value is passed to terminal-shell-gateway and it passes it to running pod. Shell container will then run helm init command with `--skip-refresh` option.


kke:
Specs fail because they're running on ruby 2.4, needs rebase / `git merge master`

jakolehm:
@nevalla waiting for docs...

nevalla:
Rebased with master and docs PR is here: https://github.com/kontena/pharos-docs/pull/134

PR #835, opened 2018-11-23
Kubernetes v1.12.2

### LINKED ISSUE #616: Kubernetes 1.12.0
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.12.md

jakolehm:
- https://kubernetes.io/blog/2018/09/27/kubernetes-1.12-kubelet-tls-bootstrap-and-azure-virtual-machine-scale-sets-vmss-move-to-general-availability/
- https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.12.md

This PR does not include new config format. Config refactoring should happen in a separate PR.

Fixes #616


jakolehm:
Tested with ubuntu, debian and centos.. seems to work and initial install seems to be faster vs 1.11.x.

@jnummelin PTAL

PR #852, opened 2018-11-28
Transfer host environment variables from Terraform output

madddi:
Environment variables (like `HTTP_PROXY`) can be configured for hosts but are currently not transferred from Terraform JSON outputs.


madddi:
Done!

PR #854, opened 2018-11-28
Remove exponential backoff from retries

kke:
The exponential backoff system makes very little sense.

```
[host-00] retrying after 512 seconds ...
```

It seems unproductive to wait for almost ten minutes, the process could just as well complete successfully in the next couple of seconds. At least, that was the expectation when the first nine retries only slept for 0..256 seconds. User already waited for 10 minutes and the backoff goes and adds extra 10 minutes of doing nothing for no good reason.

This PR changes the `retry_count` into `retry_duration`. Things will be retried until `retry_duration` seconds have passed.


kke:
It could

PR #855, opened 2018-11-28
Move kube_client setup into config

### LINKED ISSUE #853: Pharos::Config#kube_client memoization breaks reconfiguration
`Pharos::Config.kube_client` memoization (assumedly) makes kubeclient reconfiguration crash.

kke:
Fixes #853 

Removes the `ConfigureClient` phase and replaces it with setting up the kubeclient in `Pharos::Config#kube_client`. A phase that makes the `prefetch` to warm up client cache is added.

The `cluster_context['master-ssh']` is also replaced with `master.ssh`.


kke:
This or something similar has been done in #1090 or #1044

PR #860, opened 2018-11-29
update ingress-nginx to 0.20.0

### LINKED ISSUE #723: Ingress-nginx 0.20.0
- https://github.com/kubernetes/ingress-nginx/releases/tag/nginx-0.20.0
- https://github.com/kubernetes/ingress-nginx/releases/tag/nginx-0.19.0
- https://github.com/kubernetes/ingress-nginx/releases/tag/nginx-0.18.0

jnummelin:
Looking at the release notes, wasn't able to spot any other new configs needed than `enable_dynamic_certificates`.

fixes #723

### DIFF
--- a/addons/ingress-nginx/addon.rb
+++ b/addons/ingress-nginx/addon.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 Pharos.addon 'ingress-nginx' do
-  version '0.17.1'
+  version '0.20.0'
   license 'Apache License 2.0'
 
   config {
@@ -13,6 +13,7 @@ Pharos.addon 'ingress-nginx' do
       'image' => 'registry.pharos.sh/kontenapharos/pharos-default-backend:0.0.3'
     )
     attribute :tolerations, Pharos::Types::Array.default([])
+    attribute :enable_dynamic_certificates, Pharos::Types::Bool.default(false)
   }
 
   config_schema {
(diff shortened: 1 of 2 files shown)


addons/ingress-nginx/addon.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 Pharos.addon 'ingress-nginx' do
-  version '0.17.1'
+  version '0.20.0'
```
jakolehm: I think we should go directly to `0.21.0` (it has security fixes).

jnummelin: yeah. need to modify some of the flags for it, I'll update the PR soon for 0.21.0

addons/ingress-nginx/addon.rb
```diff
@@ -13,6 +13,7 @@
       'image' => 'registry.pharos.sh/kontenapharos/pharos-default-backend:0.0.3'
     )
     attribute :tolerations, Pharos::Types::Array.default([])
+    attribute :enable_dynamic_certificates, Pharos::Types::Bool.default(false)
```
jakolehm: `enable_dynamic_certificates` vs `dynamic_certificates` 🤔

jnummelin: Tried to "model" it 1-to-1 with the upstream options

jakolehm: What if we just allow to set `extra_args` (array of arguments)? It would be much more generic...

jnummelin: Sure.

Issue #856, opened 2018-11-28
HTTP_PROXY environment variables are unset for kubeadm


madddi:
**What happened**:
`HTTP_PROXY` (and similar) environment variables are not configured for components configured in kubelet manifests.

**What you expected to happen**:
When configured, the proxy server should also be set for the Kubernetes control plane components.

**How to reproduce it (as minimally and precisely as possible)**:
1. Configure `HTTP_PROXY`, `HTTPS_PROXY`, `NO_PROXY` etc. for a host
2. Deploy
3. Check a kubelet manifest - the variables are missing

**Anything else we need to know?**:
The variables are explicitly `unset` in [kubeadmin-init.sh](https://github.com/kontena/pharos-cluster/blob/master/lib/pharos/scripts/kubeadm-init.sh) and [kubeadm-reconfigure.sh](https://github.com/kontena/pharos-cluster/blob/master/lib/pharos/scripts/kubeadm-reconfigure.sh). This was changed in #395. Not all variants of http proxy environment variables are unset there, which leads to inconsistent behavior depending on the spelling. When these lines are removed, `kubeadm` configures proxy servers correctly.

In very restricted environments with no internet access it's necessary to communicate with cloud-provider APIs via a proxy server. This is why we explicitly want to configure the server for _all_ components. A correct `no_proxy` environment variable excluding all internal resources is of course necessary then.

**Environment**:
- Pharos version (use `pharos --version`): `2.0.3+oss`
- Cloud provider or hardware configuration: `aws`

kke:
Probably relevant: 
https://github.com/kubernetes/kubeadm/issues/324
https://github.com/kubernetes/kubernetes/pull/53895 (allow CIDR notation in NO_PROXY)

kke:
@jakolehm @jnummelin This was deliberately and explicitly done in #395 - should it be reverted?

kke:
(can confirm that currently for example `NO_PROXY` ends up in the manifest.)

jakolehm:
@madddi @kke #395 included this change because in most (all?) cases we want to have direct communication between different kubernetes components. It has a side-effect that kubernetes components won't get configured proxy for external service either... which is a bug. Not sure if we can generate `NO_PROXY` automatically if user has not configured it?

jakolehm:
Or maybe we shouldn't `unset` if `NO_PROXY` is set?

kke:
[...]

PR #862, opened 2018-11-30
Optionally configure proxies for control plane

kke:
Related #856 

Unsets all environment variables ending with `_PROXY` (case insensitive) before (re)configuring the control plane. #395 unsets only `http_proxy`, `HTTP_PROXY` and `HTTPS_PROXY`, leaving `NO_PROXY` and others set.

The unsetting can be disabled by setting:

```yaml
control_plane:
  use_proxy: true
```

in cluster.yml (TODO: doc update)

Note: enabling proxy for the control plane will make kubernetes components communicate over the proxy unless you pay attention to the `NO_PROXY` setting.


lib/pharos/config_schema.rb
```diff
@@ -153,6 +153,9 @@ def self.messages
         optional(:kubelet).schema do
           optional(:read_only_port).filled(:bool?)
         end
+        optional(:kubeadm).schema do
```
jakolehm: IMHO this is not the right keyword for this... let's come up with something better:
- `api.use_proxy`
- `control_plane.use_proxy`
- `something_else.use_proxy`

@kke @jnummelin wdyt?

jnummelin: I agree, we should not "leak" kubeadm to the config, there might be time where do not use it at all.

From the above suggestions, the best is `control_plane.use_proxy`.

jakolehm: Alright, let's use `control_plane.use_proxy`.

kke: True that.  Control plane sounds fitting.

kke:
Should JoinNode reconfigure existing nodes?

jakolehm:
> Should JoinNode reconfigure existing nodes?

Now that you asked this... I think `JoinNode` is not the right place for this. Maybe it should be part of the `ConfigureKubelet` phase (kubelet systemd unit is the key here).

kke:
The JoinNode command will currently always get the proxy envs, I assume it will try to use them.

kke:
Like anything that uses `kubeadm`/`kubectl` through `ssh.exec`

jakolehm:
E2e seems to be broken.

kke:
I think it's passing now, but just wondering should the setting be used for all the `kubectl` / `kubeadm` usages?

lib/pharos/phases/configure_kubelet.rb
```diff
@@ -93,6 +93,15 @@ def existing_dropin
       def build_systemd_dropin
         options = []
         options << "Environment='KUBELET_EXTRA_ARGS=#{kubelet_extra_args.join(' ')}'"
+
```
jakolehm: I think we should have specs for `build_systemd_dropin` (because it starts to be a pretty complex).

kke: Added

jakolehm:
> I think it's passing now, but just wondering should the setting be used for all the kubectl / kubeadm usages?

Not sure about that.. anyway if those need to be tweaked it can be a separate issue/pr (this is about the control-plane).

PR #865, opened 2018-11-30
Correct replica location

Timer:
This corrects the replica location for the lens deployment.

I tested this and it appeared to work, but I'd appreciate someone else giving it a test as well. :-)


Timer:
We should probably note to remove `patch_old_resource` in v3.

PR #878, opened 2018-12-04
Fix misbehaving cri-o upgrades

jakolehm:


lib/pharos/host/debian/scripts/configure-cri-o.sh
```diff
@@ -70,4 +71,9 @@ else
     if [ "$orig_config" != "$(cat /etc/crio/crio.conf)" ]; then
         reload_systemd_daemon "crio"
     fi
+
+    if [ "$orig_version" != "$(/usr/local/bin/crio -v)" ]; then
```
kke: Should this be combined with the condition above? Now it will reload it twice if both are true.

lib/pharos/host/el7/el7.rb
```diff
@@ -56,6 +56,7 @@ def configure_container_runtime
             INSECURE_REGISTRIES: insecure_registries
           )
         elsif crio?
+          can_pull = ssh.exec("sudo crictl pull #{config.image_repository}/pause:3.1").success?
```
kke: (Also possible to use `ssh.exec?`)

lib/pharos/host/debian/scripts/configure-cri-o.sh
jakolehm: Right, other OS's exit after reload... will fix.

lib/pharos/host/el7/scripts/configure-cri-o.sh
```diff
@@ -75,4 +75,9 @@ else
         reload_systemd_daemon "crio"
         exit 0
     fi
+
+    if [ "$orig_version" != "$(/usr/local/bin/crio -v)" ]; then
```
kke: same here

lib/pharos/host/ubuntu/scripts/configure-cri-o.sh
```diff
@@ -75,4 +76,9 @@ else
         reload_systemd_daemon "crio"
         exit 0
     fi
+
+    if [ "$orig_version" != "$(/usr/local/bin/crio -v)" ]; then
```
kke: and here

PR #879, opened 2018-12-04
Don't trap SIGPIPE

kke:
Trapping SIGPIPE will make ssh/http connections sometimes skip their own error handling because an exception is not raised and instead pharos exits silently without printing any reason for doing so, except `echo $?` telling it was because of `141` which comes from`SIGPIPE`.


kke:
Wondering if it is possible to figure out if the broken pipe came from something like `pharos-cluster version | head -1` or a socket error. The later would benefit from non-silent exit.

kke:
Well at least `if $stdout.tty?` can be used to see if there was some sort of piping going on.

PR #880, opened 2018-12-04
Apply psp policies on upgrade if they were not previously enabled

jakolehm:
Allows smoother upgrade from installations which don't have psp enabled.


lib/pharos/phases/upgrade_master.rb
```diff
@@ -27,9 +30,18 @@ def parse_major_minor(version)
         version.split('.')[0...2].join('.')
       end
 
+      # @return [Boolean]
+      def pod_security_disabled?
+        api_manifest = ssh.file('/etc/kubernetes/manifests/kube-apiserver.yaml')
+        return false unless api_manifest.exist?
+
+        !api_manifest.read.match(/--enable-admission-plugins=.*(PodSecurityPolicy).*/)
```
kke:
```suggestion
        !api_manifest.read.match?(/--enable-admission-plugins=.*PodSecurityPolicy/)
```

jakolehm: Why did you remove `.*` from the end?

PR #881, opened 2018-12-04
Improve SSH client thread safety

kke:
Adds a mutex/monitor to `SSH::Client` and wraps the `exec` methods et al inside `synchronize {}`.

Simultaneous connection usage should currently happen almost nowhere, except maybe to the master host via `cluster_context['master-ssh']` or `master_host.ssh`, so this could be worth nothing.

### DIFF
--- a/lib/pharos/ssh/client.rb
+++ b/lib/pharos/ssh/client.rb
@@ -19,7 +19,6 @@ module Pharos
     }.freeze
 
     class Client
-
       include MonitorMixin
 
       attr_reader :session
--- a/lib/pharos/ssh/client.rb
+++ b/lib/pharos/ssh/client.rb
@@ -3,6 +3,7 @@
 require 'net/ssh'
 require 'net/ssh/gateway'
 require 'shellwords'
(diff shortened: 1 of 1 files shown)


lib/pharos/ssh/client.rb
```diff
@@ -98,35 +103,35 @@ def exec_script!(name, env: {}, path: nil, **options)
         cmd.concat(EXPORT_ENVS.merge(env).map { |key, value| "#{key}=\"#{value}\"" })
         cmd.concat(%w(bash --norc --noprofile -x -s))
         logger.debug { "exec: #{cmd}" }
-        exec!(cmd, stdin: script, source: name, **options)
+        synchronize { exec!(cmd, stdin: script, source: name, **options) }
```
jakolehm: These are wrapped twice with `synchronize`? Probably that's a bit too much?

kke: Yeah. Monitor implements a reentrant mutex though, but still unnecessary.

PR #883, opened 2018-12-04
E2e: smoke test on DO cluster

jakolehm:


.drone.yml
```diff
@@ -12,6 +12,24 @@ pipeline:
...
+  setup-e2e:
+    image: docker.io/hashicorp/terraform:0.11.10
+    secrets: [ digitalocean_token ]
+    commands:
+      - ./e2e/drone_setup.sh
+  e2e:
+    image: ruby:2.5
+    commands:
+      - gem install bundler -Nf
+      - bundle install --path bundler
```
kke: These should probably use `--without test development`.

It could also do `gem build pharos-cluster.gemspec && gem install pharos-cluster-*.gem` and skip bundling altogether.

jakolehm: Now uses the gem.

PR #885, opened 2018-12-05
Fix uninitialized constant Pharos::Host::Ubuntu::DOCKER_VERSION

jakolehm:


.drone.yml
```diff
@@ -17,18 +17,23 @@ pipeline:
     secrets: [ digitalocean_token ]
     commands:
       - ./e2e/drone_setup.sh
+    when:
+      event: [ push ]
```
jakolehm: Had to add these because DO token is available only for push.

PR #887, opened 2018-12-05
(style) remove dot from URL

matti:


jakolehm:
For future readers: this was a test to check build pipeline security. Thanks @matti !

PR #894, opened 2018-12-07
Run e2e tests without bundler

kke:
#883 still left in the `bundle install` before `gem build`.

Changes all e2e suites to `gem build` + `gem install` and skip bundler.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,9 @@ jobs:
   include:
     - stage: e2e
       name: "e2e: docker with weave"
+      install:
+        - gem build pharos-cluster.gemspec
+        - gem install pharos-cluster*.gem
       script: ./e2e/travis.sh
       rvm: 2.5
       dist: xenial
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -10,6 +10,9 @@ jobs:
   include:
     - stage: e2e
       name: "e2e: docker with weave"
+      install:
```
jakolehm: This will make travis slower unless installed gems are cached between builds.

jakolehm: Also: what is the reason for repeating this in `.travis.yml` vs putting it into `travis.sh`?

kke: Reason is to replace the travis default `install: bundle install`, I suppose it could be set to null. 

I'm not certain `gem install` would use the cached `bundle install`ed gems anyway unless GEM_PATH is set to the bundle-dir.

jakolehm: > 'm not certain gem install would use the cached bundle installed gems anyway unless GEM_PATH is set to the bundle-dir.

 Yeah, I meant that probably we should cache gems folder for faster gem install.

kke:
Already done elsewhere

PR #898, opened 2018-12-08
Enable flying-shuttle only when multi-region cluster

jakolehm:


lib/pharos/phases/configure_weave.rb
```diff
@@ -52,6 +52,7 @@ def ensure_resources
           ipalloc_range: @config.network.pod_network_cidr,
           arch: @host.cpu_arch,
           version: WEAVE_VERSION,
+          flying_shuttle_enabled: @config.regions.size > 1,
```
jnummelin: I think `flying_shuttle_enabled == false` should trigger removing of `--no-discovery` in the `EXTRA_ARGS` env for `weave` container. Without FS and having `--no-discovery`, there's nothing that configures peering for weave. Or am I missing something? :)

jakolehm: You are right, missed that completely.

jakolehm: Actually weave fetches peers from the k8s api.. so it should work quite ok (I tested this). But I still agree that we should remove `--no-discovery` if fs is disabled.

PR #907, opened 2018-12-10
Shell auto-completions

### LINKED ISSUE #679: Add shell auto completions
Now as there are multiple subcommands, bash completions become relevant.

This needs to be somehow coordinated with chpharos so it can switch completions when switching pharoses.

kke:
wip

Fixes #679 

Automatically generated shell completions, loaded like kube completions: `source <(pharos-cluster complete)`

Chpharos needs to be modified to load the completions upon `chpharos use` if `pharos complete` does not exit with error.

### DIFF
--- a/lib/pharos/complete_command.rb
+++ b/lib/pharos/complete_command.rb
@@ -1,0 +1,82 @@
+# frozen_string_literal: true
+
+module Pharos
+  class CompleteCommand < Pharos::Command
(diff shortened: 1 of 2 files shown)


jakolehm:
@kke ping

kke:
Yeah this can be killed, going to replace with a PR using the new gem

PR #920, opened 2018-12-14
[wip/poc] Persist addon files into secrets

kke:
Stores addon files into kube-system namespace secrets as `pharos-addon-files` with the key `addons.tar.gz` as a `tar.gz` package.

This should make it possible to run the `up` with the same config on another host in the future.

The cluster config yaml (including the hosts from terraform) is already stored into ConfigMap.


kke:
image

kke:
I think this idea was abandoned

PR #927, opened 2018-12-19
Improve ubuntu docker.io version pinning

jakolehm:
If specified version is not found in repos fallback to already installed version.

### DIFF
--- a/lib/pharos/host/ubuntu/ubuntu_bionic.rb
+++ b/lib/pharos/host/ubuntu/ubuntu_bionic.rb
@@ -43,7 +43,7 @@ module Pharos
           exec_script(
             'configure-docker.sh',
             DOCKER_PACKAGE: 'docker.io',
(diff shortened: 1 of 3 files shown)


lib/pharos/host/ubuntu/ubuntu_bionic.rb
```diff
@@ -43,7 +43,7 @@ def configure_container_runtime
           exec_script(
             'configure-docker.sh',
             DOCKER_PACKAGE: 'docker.io',
-            DOCKER_VERSION: "#{DOCKER_VERSION}-0ubuntu1~18.04.1",
+            DOCKER_VERSION: "#{DOCKER_VERSION}*",
```
kke: This asterisk thing looks like a supported way to install something like `docker.io-1.2.3-*`

Maybe it could be in the `.sh` instead of manipulated in the env.

```
apt-get install -y "$[DOCKER_PACKAGE}=${DOCKER_VERSION}*"
```

PR #928, opened 2018-12-20
Prefix debug output with ssh hostname

jakolehm:
It's very difficult to follow debug output because most phases are executed in parallel and output does not have any hints which host produced the line. This PR improves output by prefixing each debug output with ssh hostname.

### DIFF
--- a/lib/pharos/ssh/remote_command.rb
+++ b/lib/pharos/ssh/remote_command.rb
@@ -134,7 +134,7 @@ module Pharos
       # @param source [String, NilClass]
       # @return [Integer]
       def debug_cmd(cmd, source: nil)
-        $stdout.write("#{INDENT} #{pastel.cyan("#{@client.host}:")} #{pastel.cyan("$ #{cmd}" + (source ? " < #{source}" : ""))}\n")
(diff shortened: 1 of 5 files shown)


lib/pharos/ssh/remote_command.rb
```diff
@@ -122,29 +125,39 @@ def initialize_debug
...
       def debug?
         @debug
       end
 
+      # @param cmd [String]
+      # @param source [String, NilClass]
+      # @return [Integer]
       def debug_cmd(cmd, source: nil)
-        $stdout.write(INDENT + pastel.cyan("$ #{cmd}" + (source ? " < #{source}" : "")) + "\n")
+        $stdout.write("#{INDENT} #{pastel.cyan("#{@client.host}:")} #{pastel.cyan("$ #{cmd}" + (source ? " < #{source}" : ""))}\n")
```
kke:
```suggestion
         $stdout.write("#{INDENT} #{pastel.cyan("#{@client.host}:")} #{pastel.cyan("$ #{cmd}" + (" < #{source}" if source))}\n")
```

should silence rubocop and skips empty string creation

PR #931, opened 2018-12-20
create cluster issuer with kube CA

### LINKED ISSUE #406: cert-manager: create ca-issuer automatically
With option to turn it off.

jnummelin:
fixes #406

This PR creates automatically a cluster internal issuer using Kube CA. This allows an easy way to cluster services use TLS between services as the trusted CA is automatically distributed using the service accounts. Basically every single pod gets the trusted CA from that.

### DIFF
--- a/addons/cert-manager/addon.rb
+++ b/addons/cert-manager/addon.rb
@@ -10,8 +10,13 @@ Pharos.addon 'cert-manager' do
     attribute :email, Pharos::Types::String
   }
 
+  ca_issuer = custom_type {
+    attribute :enabled, Pharos::Types::Bool.optional.default(true)
+  }
+
   config {
     attribute :issuer, issuer
+    attribute :ca_issuer, ca_issuer
   }
 
   config_schema {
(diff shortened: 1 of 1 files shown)


addons/cert-manager/addon.rb
```diff
@@ -73,4 +89,39 @@ def migrate_le_acme_cluster_issuers
       rc.merge_patch(issuer.metadata.name, { spec: patch_spec }, strategic_merge: false)
     end
   end
+
+  def create_ca_issuer
```
jakolehm: Maybe `build_ca_issuer` ?

addons/cert-manager/addon.rb
```diff
@@ -73,4 +89,39 @@ def migrate_le_acme_cluster_issuers
...
+    K8s::Resource.new(
+      apiVersion: "certmanager.k8s.io/v1alpha1",
+      kind: "ClusterIssuer",
+      metadata: {
+        name: 'ca-issuer'
```
jakolehm: Not sure about the naming 🤔

PR #941, opened 2019-01-02
[WIP] Kubernetes v1.13.1

jakolehm:


lib/pharos/kubeadm/cluster_config.rb
```diff
@@ -0,0 +1,208 @@
+# frozen_string_literal: true
+
+module Pharos
+  module Kubeadm
+    class ClusterConfig
+      PHAROS_DIR = Pharos::Kubeadm::PHAROS_DIR
+      AUTHENTICATION_TOKEN_WEBHOOK_CONFIG_DIR = '/etc/kubernetes/authentication'
+      AUDIT_CFG_DIR = (PHAROS_DIR + '/audit').freeze
```
kke: I think rubocop will complain about .freezing immutables for these

jakolehm: I didn't change anything here.. just refactored code to a different class.

lib/pharos/scripts/kubeadm-renew-certs.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -eu
+
+if [ "$SKIP_UNSET_PROXY" = "true" ]; then
```
jakolehm: Not sure if we need this here?

lib/pharos/kubeadm/config_generator.rb
```diff
@@ -0,0 +1,84 @@
...
+      end
+
+      # @return [Array<Hash>]
+      def generate_config
+        [init_config.generate, cluster_config.generate, kubeproxy_config.generate]
+      end
+
+      # @return [String]
+      def generate_yaml_config
+        generate_config.map(&:to_yaml).join("\n")
```
kke:
```suggestion
         YAML.dump_stream(*generate_config)
```

lib/pharos/scripts/kubeadm-renew-certs.sh
kke: 🤔  I realize this comes from copypaste but I wonder if it's right, because it looks wrong.. `if skip_unset = true { unset }`

lib/pharos/kubeadm/cluster_config.rb
```diff
@@ -0,0 +1,208 @@
...
+          'hostPath' => SECRETS_CFG_DIR,
+          'mountPath' => SECRETS_CFG_DIR
+        }
+
+        config
+      end
+
+      # @return [Array<String>]
+      def build_extra_sans
+        extra_sans = Set.new(['localhost'])
```
kke: Rather unnecessary use of Set here and also `require 'set'` is needed when not running via bundler.

jakolehm: I think only reason for `Set` is uniqueness of addresses.

jakolehm:
Specs should now pass. PTAL (again) @jnummelin @kke.

PR #947, opened 2019-01-04
Fix erb trim mode warning

### LINKED ISSUE #946: Invalid ERB trim mode used
The mode has been invalid all the time, but Ruby 2.6.0 started giving warnings about it:

```
pharos-cluster-2.2.0.alpha.0/lib/pharos/yaml_file.rb:53:
warning: Invalid ERB trim mode: "%<>-" (trim_mode: nil, 0, 1, 2, or String composed of '%' and/or '-', '>', '<>')
```

[Here](https://github.com/ruby/ruby/blob/ab5a3db48e446a33f790fa5e8594a6dc24062987/lib/erb.rb#L675-L684)'s how the trim mode is handled:

```ruby
        perc = mode.include?('%')
        if mode.include?('-')
          return [perc, '-']
        elsif mode.include?('<>')
          return [perc, '<>']
        elsif mode.include?('>')
          return [perc, '>']
        else
          [perc, nil]
        end
```

Pharos (and classic and mortar?) is using `%-<>` which will be interpreted the same as just `%-`.

```
%  enables Ruby code processing for lines beginning with %
<> omit newline for lines starting with <% and ending in %>
>  omit newline for lines ending in %>
-  omit blank lines ending in -%>
```

I think the `%` is not something that is being used in our yamls, it would be something like:

```yaml
% foo = FooGetter.get(:foo)
% if foo
there is a mystical fooness in the air
% else
no foo detected
% end
```

Maybe the `%` should be dropped and just keep the `-`. I find it easier to control the newline creation explicitly like it has been done so far.

jakolehm:
Fixes #946

### DIFF
--- a/lib/pharos/yaml_file.rb
+++ b/lib/pharos/yaml_file.rb
@@ -50,7 +50,7 @@ module Pharos
 
     def erb_result(variables = {})
       Namespace.new(variables).with_binding do |ns_binding|
-        ERB.new(@content, nil, '%<>-').tap { |e| e.location = [@filename, nil] }.result(ns_binding)
+        ERB.new(@content, nil, '<>-').tap { |e| e.location = [@filename, nil] }.result(ns_binding)
       end
     rescue StandardError, ScriptError => ex
       raise ParseError, "#{ex.class.name} : #{ex.message} (#{ex.backtrace.first.gsub(/:in `with_binding'/, '')})"
--- a/lib/pharos/yaml_file.rb
+++ b/lib/pharos/yaml_file.rb
@@ -50,7 +50,7 @@ module Pharos
 
     def erb_result(variables = {})
       Namespace.new(variables).with_binding do |ns_binding|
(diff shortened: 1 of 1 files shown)


lib/pharos/yaml_file.rb
```diff
@@ -50,7 +50,7 @@ def read(variables = {})
 
     def erb_result(variables = {})
       Namespace.new(variables).with_binding do |ns_binding|
-        ERB.new(@content, nil, '%<>-').tap { |e| e.location = [@filename, nil] }.result(ns_binding)
+        ERB.new(@content, nil, '<>-').tap { |e| e.location = [@filename, nil] }.result(ns_binding)
```
kke:
```suggestion
        ERB.new(@content, nil, '-').tap { |e| e.location = [@filename, nil] }.result(ns_binding)
```

`<>-` is equal to `-` as the code does: 

```ruby
if mode.include?('-')
  return [perc, '-']
end
```

kke: I think it can be one of:
```
%
%<
%<>
%-
```

PR #951, opened 2019-01-07
Fix k8s api connection via bastion host

### LINKED ISSUE #950: Bastion host is broken
If bastion is defined:

```
ERROR: NameError : undefined local variable or method `host' for #<Pharos::Config:0x0000000003ae3488>
```

Error probably comes from: https://github.com/kontena/pharos-cluster/blob/v2.1.3/lib/pharos/config.rb#L125

kke:
Fixes #950


kke:
Now with specs and actually using the bastion host as gateway.

PR #953, opened 2019-01-07
pharos-cloud-controller

jnummelin:
Still bit "unstable", takes few minutes to everything get ready properly. I'm still bit unsure what causes the crash loops in the start, everything gets up-and-running in few minutes after cluster initialization

Uses still my privately build image, need to change that to the common registry once proven stable.


addons/pharos-cloud-controller/resources/03-cluster-role-pvl-controller.yml
```diff
@@ -0,0 +1,22 @@
+---
```
jakolehm: Do we really need this?

jnummelin: Yes. Unfortunately PV stuff is baked into the cloud-controller lower layers:

If I remove those:
```
ERROR: logging before flag.Parse: E0322 10:56:49.597378       1 reflector.go:205] github.com/kontena/pharos-cloud-controller/vendor/k8s.io/kubernetes/pkg/controller/cloud/pvlcontroller.go:110: Failed to list *v1.PersistentVolume: persistentvolumes is forbidden: User "system:serviceaccount:kube-system:pvl-controller" cannot list resource "persistentvolumes" in API group "" at the cluster scope

```

jnummelin:
Let's wait for #1186 to land and then I'll refactor this on top of that as "managed cloud provider".

jakolehm:
@jnummelin #1186 is now merged.

jnummelin:
Refactored on top of #1186

lib/pharos/configuration/host.rb
```diff
@@ -82,6 +82,8 @@ def labels
         labels = @attributes[:labels] || {}
 
         labels['node-address.kontena.io/external-ip'] = address
+        labels['node-address.kontena.io/internal-ip'] = peer_address
```
jakolehm: We probably should not set this if `address == peer_address`.

lib/pharos/configuration/host.rb
```diff
@@ -82,6 +82,9 @@ def labels
         labels = @attributes[:labels] || {}
 
         labels['node-address.kontena.io/external-ip'] = address
+        pa = private_address || private_interface_address
+        labels['node-address.kontena.io/internal-ip'] = pa if pa
```
jnummelin: this could be also done as:
```
        labels['node-address.kontena.io/external-ip'] = address
        labels['node-address.kontena.io/internal-ip'] = private_address || private_interface_address

        labels['node-role.kubernetes.io/worker'] = '' if worker? && !labels.find{ |k, _| k.to_s.start_with?("node-role.kubernetes.io") }

        labels.compact
```

dunno if it would be any cleaner that way?

lib/pharos/configuration/host.rb
```diff
@@ -82,6 +82,9 @@ def labels
         labels = @attributes[:labels] || {}
 
         labels['node-address.kontena.io/external-ip'] = address
+        pa = private_address || private_interface_address
```
jakolehm: IMHO `pa` ->  `internal_ip` would make this more readable.

jnummelin: done

PR #957, opened 2019-01-08
PoC Clamp-completion gem

kke:
Automatic completion generation gem for clamp.

Basically just by adding `subcommand "complete", "generate shell auto-completions", Clamp::Completer.new(self)`, it will make something like this work:

```console
$ source <(pharos complete bash)
# mac bash version requires trickery:
$ source /dev/stdin <<<"$(pharos complete bash)"
```

Subcommand completions work out of the box with no changes and so do flag-type option completions. For other option completions, it's possible to do something like:

```ruby
option %w(-c --config), 'PATH', "Path to config"

def complete_path # method name derived from the PATH opt value description
  { glob: "*.yml" }
end
```

or

```ruby
option %w(-r --role), 'ROLE', "Role name", attribute_name: :role_name

def complete_role_name # method name derived from the attribute_name
  "master worker" # simple word list
end
```

Generates separate completion configs for bash and zsh, including fancy description things for zsh's fancier completion system.

(specs fail due to rubocopping the "packaged" completer_gem)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,4 @@
 source "https://rubygems.org"
 
+gem 'clamp-completer', path: 'completer_gem'
 gemspec
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    pharos-cluster (2.2.0.alpha.0)
+    pharos-cluster (2.2.0.alpha.1)
       bcrypt
       bcrypt_pbkdf (>= 1.0, < 2.0)
       clamp (= 1.2.1)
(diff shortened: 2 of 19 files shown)


jnummelin:
👏  IMO looks very cool. I especially like the fact that argument etc completions can be defined hand-in-hand with the options. Makes it easy to actually keep everything in sync.

jakolehm:
Yes, looks very nice. But ofc the gem should be in a separate repo and not here.

kke:
Poc pocced, gem now available at [kontena/clamp-completer](https://github.com/kontena/clamp-completer), going to make a PR for pharos that will utilize it.

PR #963, opened 2019-01-10
BYO networking

### LINKED ISSUE #962: Allow to use custom network provider
Pharos currently supports calico & weave. Some cases require more exotic configurations , for these it would be nice to have a custom option (bring-your-own-network-provider).

Configuration could be something like this:

```yaml
network:
  provider: custom
  custom:
    path: ./path/to/custom/cni
    options: {}
```

jnummelin:
fixes #962

### DIFF
--- a/lib/pharos/phases/configure_custom_network.rb
+++ b/lib/pharos/phases/configure_custom_network.rb
@@ -7,7 +7,6 @@ module Pharos
 
       def call
         logger.info { "Configuring custom network ..." }
-        # TODO Set options more granularly?
         stack = Pharos::Kube.stack('custom-network', @config.network.custom.manifest_path, name: 'custom_network', cluster_config: @config)
         stack.apply(kube_client)
       end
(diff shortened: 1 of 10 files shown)


lib/pharos/phases/configure_custom_network.rb
```diff
@@ -0,0 +1,16 @@
...
+
+module Pharos
+  module Phases
+    class ConfigureCustomNetwork < Pharos::Phase
+      title "Configure Custom network"
+
+      def call
+        logger.info { "Configuring custom network ..." }
+        # TODO Set options more granularly?
+        stack = Pharos::Kube.stack('custom-network', @config.network.custom.manifest_path, name: 'custom_network', cluster_config: @config)
```
jnummelin: @jakolehm @kke WDYT, should the options be set "directly" or is it enough that they can be accessed in the templates through the cluster level config struct?

jakolehm: IMO `cluster_config` is enough for now.

jnummelin:
docs at https://github.com/kontena/pharos-docs/pull/131

PR #965, opened 2019-01-10
Add option to disable outgoing NAT with calico

### LINKED ISSUE #964: Option to disable Calico NAT

In calico, you can disable NAT on the default IP Pool by setting `CALICO_IPV4POOL_NAT_OUTGOING` to `false` on the calico node pods. ([from docs](https://docs.projectcalico.org/v3.4/reference/node/configuration)) This would be a useful option for pharos to be able to set on cluster creation.

**Why is this needed**:

We want our pod ips to be routable and identifiable by network components outside the cluster. I can give a valid podcidr range and use bgp peering to establish inbound routes, but pharos only configures a NAT calico setup. With this option we can disable NAT at the node level and use pod ips network wide.

**Example cluster.yml**:

```
network:
  provider: calico
  calico:
    ipip_mode: Never
    nat_outgoing: false
  service_cidr: 10.16.0.0/17
  pod_network_cidr: 10.16.128.0/17
```

captncraig:
Simple option to disable outgoing NAT in calico. [calico configuration docs](https://docs.projectcalico.org/v3.4/reference/node/configuration)

Fixes #964 

Will document seperately.

### DIFF
--- a/lib/pharos/phases/configure_calico.rb
+++ b/lib/pharos/phases/configure_calico.rb
@@ -52,7 +52,8 @@ module Pharos
           ipip_mode: @config.network.calico&.ipip_mode || 'Always',
           ipip_enabled: @config.network.calico&.ipip_mode != 'Never',
           master_ip: @config.master_host.peer_address,
-          version: CALICO_VERSION
+          version: CALICO_VERSION,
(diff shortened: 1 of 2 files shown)


jakolehm:
@jnummelin PTAL

PR #980, opened 2019-01-15
Kubernetes v1.13.2

jakolehm:
https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.13.md#changelog-since-v1131


jakolehm:
@nevalla PTAL

PR #984, opened 2019-01-16
Enable felix prometheus metrics

captncraig:
This enables the prometheus endpoint on the calico node pods. It should be safe to enable without much impact. In order to scrape, users will need to add a service / service monitor, but this is a requirement for any of that.


captncraig:
Forgot to add ports. Modifying and will reopen.

PR #985, opened 2019-01-16
Fix Ubuntu Bionic cri-o cgroup driver fallback logic

jakolehm:
Current logic is completely busted because it will fallback to `cgroupfs` always after first install.


jakolehm:
@jnummelin PTAL

PR #989, opened 2019-01-17
Fix loading of addons from symlinked directories under pharos-addons/

kke:
Previously if you symlinked something under `$PWD/pharos-addon/`:

```
$ mkdir pharos-addons
$ ln -s /home/addons/todoapp-addon pharos-addons/
$ ls -al pharos-addons
total 0
drwx------    3 user  staff     96 Jan 17 14:58 .
drwxr-x---  421 user  staff  13472 Jan 17 15:00 ..
lrwxr-xr-x    1 user  staff     72 Jan 17 14:58 todoapp-addon -> /home/addons/todoapp-addon
```

the addon would not have been loaded.

This PR changes the glob pattern to allow recursing into symlinked directories also.

### DIFF
--- a/lib/pharos/addon_manager.rb
+++ b/lib/pharos/addon_manager.rb
@@ -26,7 +26,7 @@ module Pharos
     # @return [Array<Class<Pharos::Addon>>]
     def self.load_addons(*dirs)
       dirs.each do |dir|
-        Dir.glob(File.join(dir, '**', 'addon.rb')).each { |f| require(f) }
+        Dir.glob(File.join(dir, '*/**', 'addon.rb')).each { |f| require(f) }
       end
 
       addons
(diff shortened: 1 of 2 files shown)


lib/pharos/addon_manager.rb
```diff
@@ -26,7 +26,7 @@ def self.addons
     # @return [Array<Class<Pharos::Addon>>]
     def self.load_addons(*dirs)
       dirs.each do |dir|
-        Dir.glob(File.join(dir, '**', 'addon.rb')).each { |f| require(f) }
+        Dir.glob(File.join(dir, '*/**', 'addon.rb')).each { |f| require(f) }
```
kke: Maybe a simple `*` would suffice, I assume it would just recurse one deep?

kke: (seems to pass the spec with that)

PR #992, opened 2019-01-21
Protect (firewall) cluster using firewalld

### LINKED ISSUE #221: Firewalld (iptables) integration
We should add optional firewalld protection for infrastructures that does not have easily configurable firewall/network rules.

- https://www.firewalld.org/
- https://github.com/kubernetes/contrib/blob/master/ansible/roles/node/tasks/firewalld.yml

jakolehm:
```yaml
network:
  firewalld:
    enabled: true
    open_ports: # these are the defaults
    - port: '22'
      protocol: tcp
      roles:
      - "*"
    - port: '80'
      protocol: tcp
      roles:
      - worker
    - port: '443'
      protocol: tcp
      roles:
      - worker
    - port: '6443'
      protocol: tcp
      roles:
      - master
    - port: 30000-32767
      protocol: tcp
      roles:
      - "*"
    - port: 30000-32767
      protocol: udp
      roles:
      - "*"
    trusted_subnets: [] # array of strings
```

- [x] add cluster addresses to trusted zone
- [x] add 22/80/443/6443 to default zone
- [x] flag to enable/disable firewall (disabled by default)
- [x] node ports
- [x] custom ports/sources

Fixes #221

### DIFF
--- a/lib/pharos/scripts/configure-firewalld.sh
+++ b/lib/pharos/scripts/configure-firewalld.sh
@@ -26,6 +26,7 @@ fi
 add_port "22/tcp" "public"
 add_port "80/tcp" "public"
 add_port "443/tcp" "public"
+add_port "30000-32767/tcp" "public"
 
 if [ "$ROLE" = "master" ]; then
     add_port "6334/tcp" "public"
--- a/lib/pharos/scripts/configure-firewalld.sh
+++ b/lib/pharos/scripts/configure-firewalld.sh
@@ -13,4 +13,5 @@ if [ "$ROLE" = "master" ]; then
 fi
 firewall-cmd --permanent --add-service pharos-worker
 firewall-cmd --permanent --add-source ipset:pharos --zone trusted
+firewall-cmd --add-masquerade --permanent --zone trusted
 firewall-cmd --reload
--- a/lib/pharos/scripts/configure-firewalld.sh
+++ b/lib/pharos/scripts/configure-firewalld.sh
@@ -2,38 +2,15 @@
 
 set -e
 
-add_port() {
-    local port=$1
-    local zone=$2
-    if ! firewall-cmd --query-port "$port" --zone "$zone" ; then
-        firewall-cmd --permanent --add-port "$port" --zone "$zone"
-    fi
-}
-
-add_source() {
-    local source=$1
-    local zone=$2
-    if ! firewall-cmd --query-source "$source" --zone "$zone" ; then
-        firewall-cmd --permanent --add-source "$source" --zone "$zone"
-    fi
-}
-
 if ! systemctl is-active --quiet firewalld; then
     systemctl enable firewalld
     systemctl start firewalld
 fi
 
-add_port "22/tcp" "public"
-add_port "80/tcp" "public"
-add_port "443/tcp" "public"
-add_port "30000-32767/tcp" "public"
-
+firewall-cmd --reload
 if [ "$ROLE" = "master" ]; then
-    add_port "6334/tcp" "public"
+    firewall-cmd --permanent --add-service pharos-master
 fi
-
-for peer in $PEER_ADDRESSES ; do
-    add_source "$peer" "trusted"
-done
-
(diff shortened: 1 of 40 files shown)


lib/pharos/scripts/configure-firewalld.sh
```diff
@@ -0,0 +1,39 @@
...
+add_port "80/tcp" "public"
+add_port "443/tcp" "public"
+add_port "30000-32767/tcp" "public"
+
+if [ "$ROLE" = "master" ]; then
+    add_port "6334/tcp" "public"
+fi
+
+for peer in $PEER_ADDRESSES ; do
+    add_source $peer "trusted"
```
kke:
```suggestion
    add_source "$peer" "trusted"
```

lib/pharos/scripts/configure-firewalld.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+set -e
+
+add_port() {
+    local port=$1
+    local zone=$2
+    if ! firewall-cmd --query-port $port --zone $zone ; then
```
kke:
```suggestion
    if ! firewall-cmd --query-port "$port" --zone "$zone" ; then
```

jakolehm:
@kke any ideas how to remove a host from whitelist (if host is removed from cluster.yml)?

lib/pharos/scripts/configure-firewalld.sh
```diff
@@ -0,0 +1,39 @@
...
+        firewall-cmd --permanent --add-port $port --zone $zone
+    fi
+}
+
+add_source() {
+    local source=$1
+    local zone=$2
+    if ! firewall-cmd --query-source $source --zone $zone ; then
+        firewall-cmd --permanent --add-source $source --zone $zone
```
kke:
```suggestion
        firewall-cmd --permanent --add-source "$source" --zone "$zone"
```

kke:
Maybe `iptables -F` but it sounds a bit risky

lib/pharos/scripts/configure-firewalld.sh
```diff
@@ -0,0 +1,39 @@
...
+    local zone=$2
+    if ! firewall-cmd --query-port "$port" --zone "$zone" ; then
+        firewall-cmd --permanent --add-port $port --zone $zone
+    fi
+}
+
+add_source() {
+    local source=$1
+    local zone=$2
+    if ! firewall-cmd --query-source $source --zone $zone ; then
```
kke:
```suggestion
    if ! firewall-cmd --query-source "$source" --zone "$zone" ; then
```

lib/pharos/scripts/configure-firewalld.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+set -e
+
+add_port() {
+    local port=$1
+    local zone=$2
+    if ! firewall-cmd --query-port "$port" --zone "$zone" ; then
+        firewall-cmd --permanent --add-port $port --zone $zone
```
kke:
```suggestion
        firewall-cmd --permanent --add-port "$port" --zone "$zone"
```

jakolehm:
> Maybe there is some kind of "flush all" option in firewalld that could be run in the beginning of the script

Did not find that kind of option... but maybe there is a command to remove all sources from a zone 🤔Ofc we could somehow loop all existing sources and remove them at the beginning?

kke:
We could generate a list of removed nodes from previous configmap?

jakolehm:
> We could generate a list of removed nodes from previous config?

@jnummelin protipped that generating firewalld conf file might be easier way to achieve this.

jakolehm:
Now it generates service + ipset configs.

jakolehm:
It seems that pod communication is busted if firewalld is first enabled and then disabled (it probably wipes too many iptables rules). How should we handle this? Error out if user tries to disable firewall?

jakolehm:
@kke @jnummelin PTAL (again).

PR #994, opened 2019-01-21
Update Kontena Lens to version 1.4.0-beta.1

nevalla:
Kontena Lens 1.4.0 requires dedicated tiller so add-on will configure tiller in `kontena-lens-tiller` namespace. `lens-helm-user` role will be also pre-configured to allow to give access to cluster users to Application library.

For `helm-api` persistent volume (1GB) is added.

`charts` is new top level configuration option. It contains `repositories` array where user can configure helm repositories in the format of:
```
charts:
  repositories:
  - name: <repository_name>
    url: <repository_url>
```
If nothing is given stable helm repository will be used.

Depends on: 
- https://github.com/kontena/pharos-kube-zipper/pull/76
- https://github.com/kontena/pharos-kube-zipper/pull/77


jakolehm:
@nevalla this probably requires changes to documentation?

PR #1000, opened 2019-01-23
Pull control plane images parallel

jakolehm:
Makes multi-master installation ~40% faster.

### DIFF
--- a/lib/pharos/phases/pull_master_images.rb
+++ b/lib/pharos/phases/pull_master_images.rb
@@ -6,7 +6,7 @@ module Pharos
       title "Pull control plane images"
 
       def call
-        logger.info { "Pulling control plane images ..."}
+        logger.info { "Pulling control plane images ..." }
         cfg = kubeadm.generate_yaml_config
         ssh.tempfile(content: cfg, prefix: "kubeadm.cfg") do |tmp_file|
           @host.ssh.exec!("sudo kubeadm config images pull --config #{tmp_file}")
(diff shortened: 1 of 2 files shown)


lib/pharos/phases/pull_master_images.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      def call
+        logger.info { "Pulling control plane images ..." }
+        cfg = kubeadm.generate_yaml_config
+        ssh.tempfile(content: cfg, prefix: "kubeadm.cfg") do |tmp_file|
+          @host.ssh.exec!("sudo kubeadm config images pull --config #{tmp_file}")
+        end
+      end
+
+      def kubeadm
```
kke: Should maybe be called something else than "kubeadm" as it returns a config generator

lib/pharos/phases/pull_master_images.rb
```diff
@@ -0,0 +1,21 @@
...
+module Pharos
+  module Phases
+    class PullMasterImages < Pharos::Phase
+      title "Pull control plane images"
+
+      def call
+        logger.info { "Pulling control plane images ..." }
+        cfg = kubeadm.generate_yaml_config
+        ssh.tempfile(content: cfg, prefix: "kubeadm.cfg") do |tmp_file|
+          @host.ssh.exec!("sudo kubeadm config images pull --config #{tmp_file}")
```
kke:
```suggestion
          ssh.exec!("sudo kubeadm config images pull --config #{tmp_file}")
```

lib/pharos/phases/pull_master_images.rb
jakolehm: Maybe... but this is c&p from another phase.. does it make this better/worse? 🙈

PR #1003, opened 2019-01-24
Add "pharos reboot" command for rebooting hosts or the entire cluster

### LINKED ISSUE #996: Reboot command

`pharos reboot [--role] [--label] [--address]` (basically similar flags that reset has)

**Why is this needed**:

Sometimes it would be handy to test full cluster reboot ... but currently it's not exactly possible via `pharos ssh -- systemctl reboot` because hosts may hang forever (because of volume mounts). Reboot subcommand could be more clever and drain a node before reboot and uncordon after boot.

kke:
Fixes #996 

Adds `pharos reboot` for rebooting nodes.

If any of the hosts to reboot are masters, they will be rebooted first. The nodes will be drained before the reboot and uncordoned once they have finished rebooting.


lib/pharos/autoload.rb
```diff
@@ -47,6 +47,9 @@ module CoreExt
 
   module SSH
     autoload :Client, 'pharos/ssh/client'
+    autoload :Error, 'pharos/ssh/client'
```
jakolehm: Why do we need these?

lib/pharos/config.rb
```diff
@@ -37,6 +37,11 @@ def self.load(raw_data)
       config.hosts.each do |host|
         host.api_endpoint = config.api&.endpoint
         host.config = config
+        # map bastion.host to an existing host
```
jakolehm: How this is related?

lib/pharos/autoload.rb
kke: `Pharos::SSH::Client` is not yet loaded when the phases are loading.

lib/pharos/config.rb
kke: Otherwise the `bastion.host` points to a duplicate created via `Configuration::Host.new` instead to the actual host in config.

This means, that for example `host.ssh.gateway.shutdown!` when doing a host reboot/reconnect will shut down a gateway that is not the duplicated bastion host's gateway.

lib/pharos/config.rb
```diff
@@ -122,7 +127,7 @@ def kube_client(kubeconfig)
         api_port = 6443
       else
         api_address = 'localhost'
-        api_port = master_host.bastion.host.ssh.gateway(master_host.api_address, 6443)
+        api_port = master_host.bastion.host.ssh.gateway.open(master_host.api_address, 6443)
```
kke: `ssh.gateway` now returns a `Net::SSH::Gateway` instead of directly doing `gateway.open`.

lib/pharos/ssh/client.rb
```diff
@@ -48,23 +59,24 @@ def bastion
...
-      # @param host [String]
-      # @param port [Integer]
-      # @return [Integer] local port number
-      def gateway(host, port)
-        Net::SSH::Gateway.new(@host, @user, @opts).open(host, port)
+      # @return [Net::SSH::Gateway]
+      def gateway
+        @gateway ||= Net::SSH::Gateway.new(@host, @user, @opts).tap do |gw|
+          gw.instance_exec do
+            @thread.report_on_exception = false
```
kke: This was added because when rebooting a node that acts as a bastion, the gateway thread for the host with bastion will die from IOError and report the exception to the terminal.

It seems to manage to reconnect with all the connection/reconnection additions in this PR and the exception appears to be harmless.

lib/pharos/ssh/client.rb
```diff
@@ -134,17 +146,34 @@ def interactive_session
...
+          gateway_shutdown
+          @session = nil
+        end
       end
 
       private
 
       def require_session!
-        raise Error, "Connection not established" if @session.nil? || @session.closed?
+        connect(timeout: 3) unless connected?
```
kke: Changed the "SSH::Client.exec` etc that use `require_session!` attempt to make a connection instead of failing directly.

lib/pharos/ssh/client.rb
```diff
@@ -48,23 +59,24 @@ def bastion
       # @param options [Hash] see Net::SSH#start
       def connect(**options)
         synchronize do
-          logger.debug { "connect: #{@user}@#{@host} (#{@opts})" }
-          if bastion
-            gw_opts = {}
-            gw_opts[:keys] = [bastion.ssh_key_path] if bastion.ssh_key_path
-            gateway = Net::SSH::Gateway.new(bastion.address, bastion.user, gw_opts)
```
kke: Becuse `ssh.gateway` returned `ssh.gateway.open`, the `Net::SSH::Gateway.new` was duplicated here and was left in a "dangling" local variable that could not be accessed for `shutdown!`.

lib/pharos/reboot_command.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      puts pastel.green("==> Sharpening tools ...")
+      cluster_manager.gather_facts
+
+      unless master_hosts.empty?
+        puts pastel.green("==> Rebooting #{master_hosts.size} master node#{'s' if master_hosts.size > 1} ...")
+        cluster_manager.apply_reboot_hosts(master_hosts, parallel: false)
+      end
+
+      puts pastel.green("==> Resharpening tools ...")
```
kke: After rebooting masters, it seemed logical to regather facts, but now I realize it may never reach this point if the master reboot failed to bring up kubelet again. So, this could be skipped I think.

lib/pharos/phases/configure_client.rb
```diff
@@ -18,8 +18,6 @@ def call
         return if @optional && !kubeconfig?
 
         cluster_context['kubeconfig'] = kubeconfig
-        cluster_context['master-ssh'] = ssh
```
kke: Unnecessary, the phase.rb can whip up an ssh client to a host with `master_sort_score.zero?`.

lib/pharos/scripts/reboot-asap.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
jakolehm: Would it be easier to generate this in ruby code?

lib/pharos/phases/reboot_host.rb
```diff
@@ -0,0 +1,42 @@
...
+    class RebootHost < Pharos::Phase
+      title "Reboot hosts"
+
+      EXPECTED_ERRORS = Pharos::SSH::Client::CONNECTION_RETRY_ERRORS + [
+        Pharos::SSH::NotConnected,
+        Pharos::SSH::RemoteCommand::ExecError
+      ].freeze
+
+      def call
+        reboot && reconnect && uncordon
```
jakolehm: Why `&&` ? Methods don't actually return boolean?

lib/pharos/reboot_command.rb
```diff
@@ -0,0 +1,48 @@
...
+
+      master_hosts = filtered_hosts.select(&:master?)
+      worker_hosts = filtered_hosts.select(&:worker?)
+
+      puts pastel.green("==> Sharpening tools ...")
+      cluster_manager.gather_facts
+
+      unless master_hosts.empty?
+        puts pastel.green("==> Rebooting #{master_hosts.size} master node#{'s' if master_hosts.size > 1} ...")
+        cluster_manager.apply_reboot_hosts(master_hosts, parallel: false)
```
kke: Masters are booted sequentially instead of parallelly here.

lib/pharos/phase.rb
```diff
@@ -80,8 +80,23 @@ def host_configurer
...
     def master_ssh
-      return cluster_context['master-ssh'] if cluster_context['master-ssh']
-      fail "Phase #{self.class.name} does not have master ssh"
+      master_hosts = @config.master_hosts
+      master_host = master_hosts.first
+
+      # Select non-self master host when available
+      if master_hosts.size > 1 && master_host == host
+        alt_masters = master_hosts[1..-1].select(&:master_valid?)
+        master_host = alt_masters.first unless alt_masters.empty?
```
jakolehm: What happens when we don't find "valid" master here?

lib/pharos/configuration/bastion.rb
```diff
@@ -3,12 +3,18 @@
...
       attribute :user, Pharos::Types::Strict::String
       attribute :ssh_key_path, Pharos::Types::Strict::String
 
+      def ==(other)
+        address == other.address && user == other.user && ssh_key_path == other.ssh_key_path
+      end
+
       def host
-        Host.new(address: address, user: user, ssh_key_path: ssh_key_path)
+        @host ||= Host.new(address: address, user: user, ssh_key_path: ssh_key_path)
```
jakolehm: `||=` with `attr_writer` feels a bit broken?

lib/pharos/scripts/reboot-asap.sh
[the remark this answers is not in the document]
kke:
```
require 'time'
current_time = Time.parse(ssh.exec!('date +%H:%M:%S`))
if current_time.sec > 55
  ssh.exec!('shutdown -r +1')
  sleep 60
else
  new_time = Time.parse((current_time + 60).strftime('%H:%M:00'))
  ssh.exec!("shutdown -r #{new_time.strftime('%H:%M')}")")
  sleep new_time - current_time
end
```

lib/pharos/phase.rb
kke: The original (self) master will be used (assigned above)

kke: This might be completely useless. The assumption is that the master does work before and after reboot and is capable of `kubectl drain`ing and `uncordon`ing itself, as I'm testing this on one master anyway.

jakolehm: Yes I get that... but does it work if "self" is used?

kke: It does, I removed the whole trickery here.

lib/pharos/configuration/bastion.rb
kke: Well yes, if someone would override the `host=` method created by `attr_writer`, it would not be called.

The "alternative" that would use the writer method would be:

```
def host
  @host || self.host = Host.new
end
```

kke:
Badly out of sync. Maybe wait for #1044 (this will need to refuse booting the localhost I suppose).

kke:
Oh, #1044 is in and is why this conflicted so bad :)

lib/pharos/config.rb
```diff
@@ -123,7 +123,7 @@ def kube_client(kubeconfig)
         api_port = 6443
       else
         api_address = 'localhost'
-        api_port = master_host.bastion.host.transport.gateway(master_host.api_address, 6443)
+        api_port = master_host.bastion.gateway.open(master_host.api_address, 6443)
```
kke: This requires #1090

lib/pharos/scripts/reboot-asap.sh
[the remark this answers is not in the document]
jakolehm: Remind me again why we need to delay reboot?

lib/pharos/phases/configure_etcd.rb
```diff
@@ -21,6 +21,7 @@ def call
           PEER_NAME: peer_name(@host),
           ARCH: @host.cpu_arch.name
         )
+        host.checks['etcd_ca_exists'] = true
```
jakolehm: Are we using this?

lib/pharos/phases/configure_master.rb
```diff
@@ -33,6 +33,7 @@ def call
         end
 
         cluster_context['master-certs'] = pull_kube_certs unless cluster_context['master-certs']
+        host.checks['kubelet_configured'] = true
```
jakolehm: Are we using this?

lib/pharos/phases/configure_master.rb
```diff
@@ -90,6 +91,7 @@ def push_kube_certs(certs)
           transport.file(path).write(contents)
           transport.exec!("sudo chmod 0400 #{path}")
         end
+        host.checks['ca_exists'] = true
```
jakolehm: Are we using this?

lib/pharos/phases/configure_master.rb
```diff
@@ -99,6 +101,7 @@ def pull_kube_certs
           path = File.join(KUBE_DIR, 'pki', file)
           certs[file] = transport.file(path).read
         end
+        host.checks['ca_exists'] = certs.key?('ca.key')
```
jakolehm: Are we using this?

lib/pharos/phases/reboot_host.rb
```diff
@@ -0,0 +1,47 @@
...
+  module Phases
+    class RebootHost < Pharos::Phase
+      title "Reboot hosts"
+
+      EXPECTED_ERRORS = Pharos::Transport::SSH::RETRY_CONNECTION_ERRORS + [
+        Pharos::ExecError
+      ].freeze
+
+      def call
+        if host.local?
```
jakolehm: Is this right?

lib/pharos/phases/configure_master.rb
kke:
```
lib/pharos/phases/validate_host.rb
40:        raise Pharos::InvalidHostError, "Cannot change worker host role to master" if @host.master? && !@host.checks['ca_exists']
41:        raise Pharos::InvalidHostError, "Cannot change master host role to worker" if @host.worker? && @host.checks['ca_exists']
```

lib/pharos/phases/configure_master.rb
kke:
```ruby
      def new?
        !checks['kubelet_configured']
      end

      # @return [Integer]
      def master_sort_score
        if checks['api_healthy']
          0
        elsif checks['kubelet_configured']
          1
        else
          2
        end
      end
```

lib/pharos/phases/reboot_host.rb
kke: No, it's reversed.

lib/pharos/phases/configure_master.rb
jakolehm: I meant are we using it in this PR? How is this relevant?

lib/pharos/reboot_command.rb
```diff
@@ -0,0 +1,60 @@
...
+      master_hosts = filtered_hosts.select(&:master?).reject(&:local?)
+      worker_hosts = filtered_hosts.select(&:worker?).reject(&:local?)
+      local_hosts  = filtered_hosts.select(&:local?)
+
+      puts pastel.green("==> Sharpening tools ...")
+      cluster_manager.gather_facts
+
+      unless local_hosts.empty?
+        puts "  " + pastel.red("!" * 76)
+        puts pastel.red("    The host will remain cordoned (workloads will not be scheduled on it) after the reboot")
```
kke: All this pastel stuff need to be updated too

lib/pharos/scripts/reboot-asap.sh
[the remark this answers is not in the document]
jakolehm: How about:
```ruby
ssh.exec!("shutdown -r now &")
sleep 1 until !ssh.connected?
```
```

File: lib/pharos/reboot_command.rb [hunk 21]
```diff
@@ -0,0 +1,60 @@
...
+        puts pastel.red("    To uncordon, you must use: ") + pastel.cyan("pharos exec -c #{config_yaml.filename} -r master -f -- kubectl uncordon #{local_hosts.first}")
```
kke: Hmm, it could be possible to schedule uncordoning by using something like:

```console
echo "kubectl uncordon #{localhost.hostname}" | at now + 10 minutes
```

before issuing the reboot command on the localhost

lib/pharos/scripts/reboot-asap.sh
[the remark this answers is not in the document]
kke: To disconnect ssh client gracefully.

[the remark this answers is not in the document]
kke:
```
[36] pry(main)> ssh.exec('sudo shutdown -r now &')
IOError: closed stream
```

It still exits instantly and possibly crashes some bastion thread. Maybe the bastion hosts need to be rebooted separately.

The transport PR changed a lot stuff and #1090 (which I would like to have merged before finalizing this one) will add more, this PR was made before all that and is probably not up to date at all.

lib/pharos/reboot_command.rb
```diff
@@ -0,0 +1,60 @@
...
+      end
+
+      unless master_hosts.empty?
+        puts pastel.green("==> Rebooting #{master_hosts.size} master node#{'s' if master_hosts.size > 1} ...")
+        cluster_manager.apply_reboot_hosts(master_hosts, parallel: false)
+      end
+
+      unless worker_hosts.empty?
+        puts pastel.green("==> Rebooting #{worker_hosts.size} worker node#{'s' if worker_hosts.size > 1} ...")
+        cluster_manager.apply_reboot_hosts(worker_hosts, parallel: true)
```
jakolehm: It's not safe to reboot all workers in parallel. At minimum there needs to be an option for this.

lib/pharos/scripts/reboot-asap.sh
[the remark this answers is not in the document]
jakolehm: Fair enough, but I feel that adding timers is not the right way to do this (they always introduce races). Maybe we need to wait until #1090 is merged.

kke:
Cleaned up after the transport changes, I feel there might still be a bit too much logic.

kke:
Hmm I wonder if this is still 2.4.0 -worthy. Seems quite simple nowadays. Maybe add a e2e after "worker up".

PR #1004, opened 2019-01-24
[wip] Validate configuration changes

### LINKED ISSUE #997: Validate configuration changes phase

A phase that validates configuration changes. 

**Why is this needed**:

Some changes are not possible and we should catch those early. For example it's not possible to change `network.provider`.

kke:
Fixes #997 

Here's the mechanism

For addons added a hook:
```ruby
  validate_configuration_changes { |old_config, new_config|
    raise "can't uninstall!" if old_config.enabled && !new_config.enabled
  end
```

For phases / general config changes, made the cluster_manager load previous config earlier and run `ValidateConfigurationChanges` phase if a working master was found during `GatherFacts`.

It appears to work:

image

### DIFF
--- a/lib/pharos/addon.rb
+++ b/lib/pharos/addon.rb
@@ -135,6 +135,10 @@ module Pharos
         hooks[:modify_cluster_config] = block
       end
 
+      def validate_configuration_changes(&block)
+        hooks[:validate_configuration_changes] = block
+      end
+
       def validation
         Dry::Validation.Params(Schema) { yield }
       end
--- a/lib/pharos/phases/validate_configuration_changes.rb
+++ b/lib/pharos/phases/validate_configuration_changes.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+module Pharos
+  module Phases
+    class ValidateConfigurationChanges < Pharos::Phase
+      title "Validate configuration changes"
+
+      DEFAULT_PROC = proc { |key, old_val, new_val| raise Pharos::ConfigError, { key => "can't change #{key} from #{old_val} to #{new_val}"} }
+
+      def call
+        changed?('network.provider', &DEFAULT_PROC)
+      end
+
+      def changed?(config_key_path)
+        old_value = string_dig(config_key_path, previous_config)
+        new_value = string_dig(config_key_path, @config)
+        return false if old_value == new_value
+        return true unless block_given?
+        yield config_key_path, old_value, new_value
+      end
+
+      def string_dig(string, source)
+        string.to_s.split('.').inject(source) { |memo, item| memo.send(item.to_sym) if memo.respond_to?(item.to_sym) }
+      end
+
+      def previous_config
+        cluster_context['previous-config']
+      end
+    end
+  end
+end
+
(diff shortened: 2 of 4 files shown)


lib/pharos/addon.rb
```diff
@@ -135,6 +136,30 @@ def modify_cluster_config(&block)
         hooks[:modify_cluster_config] = block
       end
 
+      def validate_configuration(&block)
+        hooks[:validate_configuration] = block
+      end
+
+      def apply_validate_configuration(old_config, new_config)
```
jakolehm: Yardoc missing.

lib/pharos/config.rb
```diff
@@ -140,5 +140,18 @@ def set(key, value)
     def to_yaml
       YAML.dump(to_h.deep_stringify_keys)
     end
+
+    # @example deep get network provider
+    #   config.deep_get("network.provider")
+    # @param key [String] a dot separated key, such as network.provider
```
jakolehm: `@return` missing.

lib/pharos/phases/validate_configuration_changes.rb
```diff
@@ -0,0 +1,30 @@
...
+      def changed?(config_key)
+        old_value = previous_config&.deep_get(config_key)
+        new_value = @config&.deep_get(config_key)
+        return false if old_value == new_value
+        return true unless block_given?
+
+        yield config_key, old_value, new_value
+      end
+
+      def previous_config
```
jakolehm: Should this be in `Phase`?

kke: It's only used in addon_manager in addition to this phase.

lib/pharos/config.rb
```diff
@@ -140,5 +140,19 @@ def set(key, value)
...
+    # @return [Object,NilClass] nil when any part of the chain is unreachable
+    def deep_get(key)
```
jakolehm: Probably should use `Pharos::CoreExt::DeepKeys`?

kke: This one doesn't need to hashify in the middle. Interestingly, something like `config.deep_get("hosts.first.address")` works with this.

jakolehm: Is  one `.to_h` that bad? It would remove duplicate code.

kke: I threw away the DeepKeys refinement thing and replaced it with HashDiff and renamed `config.deep_get` to `config.dig` which acts like `hash.dig`: `config.dig(:network, :provider)`.

lib/pharos/autoload.rb
```diff
@@ -47,6 +47,7 @@ module CommandOptions
   module CoreExt
     autoload :IPAddrLoopback, 'pharos/core-ext/ip_addr_loopback'
     autoload :DeepTransformKeys, 'pharos/core-ext/deep_transform_keys'
+    autoload :DeepKeys, 'pharos/core-ext/deep_keys'
```
jakolehm: Not needed anymore?

jakolehm:
```suggestion
```

PR #1005, opened 2019-01-25
Fix nested addon configuration validation errors

### LINKED ISSUE #998: Validation error handling cannot handle nested errors
**What happened**:

Pharos throws mystic `NoMethodError` if for example addon is missing nested keys.

```
ERROR: NoMethodError : undefined method `join' for {:region=>["is missing"], :bucket=>["is missing"]}:Hash
```

Validation error handling should not explode.

**cluster.yml:**

```yaml
addons:
  kontena-backup:
   enabled: true
   cloud_credentials: /path/minio_credentials
   aws:
     s3_force_path_style: true
     s3_url: foo.com/minio/test
```

kke:
Fixes #998 

From:
```
ERROR: NoMethodError : undefined method `join' for #<Hash:0x00007f90e1206560>
```

To:

image

### DIFF
--- a/lib/pharos/addon_manager.rb
+++ b/lib/pharos/addon_manager.rb
@@ -7,6 +7,7 @@ require_relative 'kube'
 module Pharos
   class AddonManager
     include Pharos::Logging
+    using Pharos::CoreExt::DeepTransformKeys
 
     class InvalidConfig < Pharos::Error; end
     class UnknownAddon < Pharos::Error; end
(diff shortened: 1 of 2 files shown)


lib/pharos/addon_manager.rb
```diff
@@ -74,8 +75,7 @@ def validate
       with_enabled_addons do |addon_class, config|
         outcome = addon_class.validate(config)
         unless outcome.success?
-          error_msg = "#{addon_class.addon_name} => " + outcome.errors.map { |key, value| "#{key} #{value.join(',')}" }.flatten.join(', ')
-          raise InvalidConfig, error_msg
+          raise InvalidConfig, YAML.dump(addon_class.addon_name => outcome.errors.deep_stringify_keys).gsub(/^---$/, '')
```
jakolehm: This is a really clever fix 👍 Would be great to have specs also.

PR #1006, opened 2019-01-25
Improve kontena-lens addon installation flow

nevalla:
This PR improves `kontena-lens` add-on installation when user management is disabled. Now user management Deployment and authenticator DaemonSet won't be deployed if user management is disabled. Also admin user won't be created and, thus, no credentials are displayed on post installation messages.

However, if user management is first enabled and then disabled we don't remove users or groups and this is on purpose.

### DIFF
--- a/non-oss/pharos_pro/addons/kontena-lens/resources/06-authenticator-daemonset.yml.erb
+++ b/non-oss/pharos_pro/addons/kontena-lens/resources/06-authenticator-daemonset.yml.erb
@@ -1,3 +1,4 @@
+<%- if user_management -%>
 apiVersion: extensions/v1beta1
 kind: DaemonSet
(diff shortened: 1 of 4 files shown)


non-oss/pharos_pro/addons/kontena-lens/resources/06-authenticator-daemonset.yml.erb
```diff
@@ -44,4 +45,5 @@ spec:
             limits:
               memory: "256Mi"
               cpu: "50m"
-      hostNetwork: true
\ No newline at end of file
+      hostNetwork: true
+  <%- end -%>
```
jakolehm:
```suggestion
<%- end -%>
```

PR #1016, opened 2019-01-29
Enable kubelet server tls bootstrap

jakolehm:


jakolehm:
@jnummelin PTAL

jakolehm:
> On a patch upgrade or regular "re-up" with same kube version, the kubelet config is not re-written. 

This is only problem if user upgrades from pre-release, right?

jnummelin:
> This is only problem if user upgrades from pre-release, right?

AFAIK yes. But I do not see why we could not update to config on each run. We might need to fix some kubelet options on patch releases too.

jakolehm:
> AFAIK yes. But I do not see why we could not update to config on each run. We might need to fix some kubelet options on patch releases too.

Yes but even if we store the updated config it's not applied automatically to kubelets (kubelet restart is needed).

PR #1018, opened 2019-01-29
Fix non-oss cluster version validation malformed version string error

kke:
The non-oss `ValidateVersion` phase does not cut the `+oss` from the version string and reports "malformed version string" when trying to go from oss to non-oss.

This PR fixes that.


kke:
Spec added.

PR #1025, opened 2019-01-30
Add pharos-cluster-critical priority class

jakolehm:
Actually working implementation of #1020


jnummelin:
Why are cert-manager, tiller and host-upgrades "cluster-critical"? 

I would sacrifice those before my critical apps

PR #1027, opened 2019-01-30
Allow to set `NO_MASQ_LOCAL` for weave

### LINKED ISSUE #1010: SourceIP preservation with weave-net not possible
**What happened**:
Attempting to enable [SourceIP preservation as described here](https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-type-nodeport) using `externalTrafficPolicy: Local` fails, as the IP-address seen by the pod is always the weave ip of the node.

The client-ip should be preserved and the pod should be able to access it.

* Setup a basic pharos cluster with weave-net (default), no fancy config
* follow the steps over here https://kubernetes.io/docs/tutorials/services/source-ip/#source-ip-for-services-with-type-nodeport

This should be related to weaveworks/weave#2924, as weave masquerades the local traffic by default. This is highly undesired, as applications like ingress-controllers need the source-ip for subnetting, rate-limiting, etc.
According to the issue linked above, this should be fixable by adding the env-var `NO_MASQ_LOCAL=1` over here: 
https://github.com/kontena/pharos-cluster/blob/88162bb9d1086c31fac344ae1162481cd0ce6647/lib/pharos/resources/weave/daemon-set.yml.erb#L18-L29

**Environment**:
- Pharos version (use `pharos --version`): 2.1.2
- Cloud provider or hardware configuration: KVM
- OS (e.g. from /etc/os-release; both for the client and for the used nodes): ubuntu1604

**cluster.yml:**
```yaml
hosts:
  # masters
  - address: "10.0.0.10"
    role: master

  # workers
  - address: "10.0.0.20"
    role: worker

network:
  provider: weave
```

jnummelin:
Setting this up in the config:
```
network:
  weave:
    no_masq_local: true
```

```
$ for node in $NODES; do curl -s $node:$NODEPORT | grep -i client_address; done
client_address=10.38.0.0
client_address=10.32.0.1
client_address=10.44.0.0

$ kubectl patch svc nodeport -p '{"spec":{"externalTrafficPolicy":"Local"}}'
service/nodeport patched

$ for node in $NODES; do curl --connect-timeout 1 -s $node:$NODEPORT | grep -i client_address; done
client_address=91.153.x.y
```
^ the reported address is my actual address :)

Docs PR: https://github.com/kontena/pharos-docs/pull/147

fixes #1010

### DIFF
--- a/lib/pharos/config_schema.rb
+++ b/lib/pharos/config_schema.rb
@@ -113,6 +113,7 @@ module Pharos
           end
           optional(:weave).schema do
             optional(:trusted_subnets).each(type?: String)
+            optional(:no_masq_local).filled(:bool?)
           end
           optional(:calico).schema do
             optional(:ipip_mode).filled(included_in?: %(Always, CrossSubnet, Never))
--- a/lib/pharos/configuration/network.rb
+++ b/lib/pharos/configuration/network.rb
@@ -7,6 +7,7 @@ module Pharos
     class Network < Pharos::Configuration::Struct
       class Weave < Pharos::Configuration::Struct
         attribute :trusted_subnets, Pharos::Types::Array.of(Pharos::Types::String)
+        attribute :no_masq_local, Pharos::Types::Strict::Bool.default(false)
 
         # @param routes [Array<Pharos::Configuration::Host::Routes>]
         # @return [Array<Pharos::Configuration::Host::Routes>]
--- a/lib/pharos/phases/configure_weave.rb
+++ b/lib/pharos/phases/configure_weave.rb
@@ -54,7 +54,8 @@ module Pharos
           version: WEAVE_VERSION,
           firewalld_enabled: !!@config.network&.firewalld&.enabled,
           flying_shuttle_enabled: @config.regions.size > 1,
-          flying_shuttle_version: WEAVE_FLYING_SHUTTLE_VERSION
+          flying_shuttle_version: WEAVE_FLYING_SHUTTLE_VERSION,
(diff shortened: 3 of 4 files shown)


jnummelin:
The option defaults to false to preserve the existing behaviour in existing clusters. Putting it on by default might break network policies etc. for existing clusters.

lib/pharos/config_schema.rb
```diff
@@ -113,6 +113,7 @@ def self.messages
           end
           optional(:weave).schema do
             optional(:trusted_subnets).each(type?: String)
+            optional(:no_masq_local).filled(:bool?)
```
jakolehm: Would `masq_local` (default `true`) make more sense?

jnummelin: As mentioned in the comment, it might actually break an existing cluster.

jakolehm: I think you missed my point, `no_masq_local` vs `masq_local` as config variable name.

jnummelin: ah, sorry. Now I see what you meant...

I think it makes to sense to keep the option 1-to-1 with the source, i.e. Weave env

jakolehm: Ok, fair enough.. I just always feel bad when a boolean variable name starts  with `no` 😄

jnummelin: yeah, it's not perfect. but IMO easier to map to downstream configs & docs when needed

PR #1033, opened 2019-01-31
Rename bin/pharos-cluster to bin/pharos

kke:
Also updates the included examples/documentation to refer to `pharos` instead of `pharos-cluster`.


jakolehm:
Probably not needed anymore.

PR #1044, opened 2019-02-04
Localhost access without SSH

kke:
A host with local address can now be accessed without ssh in all commands.

- The whole `Pharos::SSH` tree was refactored into `Pharos::Transport`.
- Added `pharos exec` alias for `pharos ssh`

72 files changed..


kke:
I successfully ran `pharos up` on a linux host with one of the hosts being `address: 127.0.0.1` using this.

jakolehm:
@kke travis says no (it runs against localhost).

kke:
image

Hmm yes I suppose it should use sudo for file operations. But I think that's actually easy, it can probably use the same RemoteFile thing but with local-exec transport.

kke:
Travis says yes now.

lib/pharos/configuration/host.rb
```diff
@@ -228,6 +223,8 @@ def configure_bastion(bastion)
 
         attributes[:bastion] = bastion
       end
+
+      def_delegators :transport, :tempfile, :exec, :exec!, :exec?, :exec_script!, :interactive_session, :file, :connected?, :disconnect
```
jakolehm: Maybe move this to top?

addons/openebs/addon.rb
```diff
@@ -43,7 +43,7 @@
 
   reset_host { |host|
     data_dir = config.default_storage_pool[:path].strip
-    host.ssh.exec("sudo rm -rf #{data_dir}/*") unless data_dir.empty?
+    host.exec("sudo rm -rf #{data_dir}/*") unless data_dir.empty?
```
kke: These could actually be `transport.exec` and the host wouldn't need the delegator.

lib/pharos/autoload.rb
```diff
@@ -11,6 +11,8 @@
 autoload :K8s, 'k8s-client'
 autoload :Excon, 'excon'
 autoload :K8s, 'k8s-client'
+autoload :Open3, 'open3'
```
jakolehm: Just a note that this autoload thing is starting to look a bit monster-ish... we probably should look how to make autoloading better.

lib/pharos/configuration/host.rb
```diff
@@ -78,25 +80,17 @@ def short_hostname
...
+        IPAddr.new(address).loopback?
+      end
 
-        opts = {}
-        opts[:keys] = [ssh_key_path] if ssh_key_path
-        opts[:send_env] = [] # override default to not send LC_* envs
-        opts[:proxy] = Net::SSH::Proxy::Command.new(ssh_proxy_command) if ssh_proxy_command
-        opts[:bastion] = bastion if bastion
-        @ssh = Pharos::SSH::Client.new(address, user, opts.merge(options)).tap(&:connect)
+      def transport(**options)
```
jakolehm: Yardocs would be nice.

lib/pharos/phases/authenticate_ssh.rb
```diff
@@ -6,7 +6,9 @@ class AuthenticateSSH < Pharos::Phase
       title "Authenticate SSH connection"
 
       def call
-        host.ssh(non_interactive: false)
+        return if host.local?
```
jakolehm: Is this necessary? Authentication is skipped if local.

lib/pharos/phases/configure_client.rb
```diff
@@ -18,19 +18,19 @@ def call
         return if @optional && !kubeconfig?
 
         cluster_context['kubeconfig'] = kubeconfig
-        cluster_context['master-ssh'] = ssh
+        cluster_context['master-ssh'] = host.transport
```
jakolehm:
```suggestion
        cluster_context['master-transport'] = host.transport
```

kke: I dropped the whole thing and replaced with `master_host.transport` where it was used.

kke:
Still need to tweak kubeconfig / ssh subcommands to perform the ssh connection + authentication properly

kke:
Should be PTAL'able again.

I ditched the two-phase ssh auth and made `ssh.connect` do the retry with `non_interactive: false` on its own.

Also added a `pharos kubeconfig` step to drone e2e.

PR #1045, opened 2019-02-04
Add dev gems to Gemfile, drop Gemfile.lock from VCS

kke:
On Ruby 2.6 `bundle exec irb` fails unless you have irb in the bundle.

The gemfile.lock in the repo seems to cause more problems than benefits.

Added byebug while at it.


kke:
Rubocop gets updated because of this and causes a lot of failures. Fixing.

jakolehm:
Conflicts...

kke:
Confict be gone

jakolehm:
Conflict again..

kke:
Conflict away again

pharos-cluster.gemspec
```diff
@@ -49,4 +49,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "rspec", "~> 3.7"
   spec.add_development_dependency "fakefs", "~> 0.13"
   spec.add_development_dependency "rubocop", "~> 0.57"
+  spec.add_development_dependency "irb", "~> 1.0"
```
jakolehm: Should we switch to `pry`?

kke: We can try

kke:
Back in sync again.

PR #1048, opened 2019-02-04
Convert addon { .. } blocks to do .. end

kke:
Purely for cosmetic reasons

### DIFF
--- a/addons/cert-manager/addon.rb
+++ b/addons/cert-manager/addon.rb
@@ -4,31 +4,31 @@ Pharos.addon 'cert-manager' do
   version '0.5.2'
   license 'Apache License 2.0'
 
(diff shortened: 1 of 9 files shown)


jakolehm:
Needs also update to docs.

kke:
You can still use `{ ... }`

jakolehm:
> You can still use `{ ... }`

I know but if `do .. end` is the best practice we should update the docs.

kke:
Currently there are mixed `{}` and `do end`, even in the same schemas. Need to pick one :)

jakolehm:
> Currently there are mixed `{}` and `do end`, even in the same schemas. Need to pick one :)

I thought this PR already picked one? Or did I misunderstood something?

PR #1049, opened 2019-02-04
Extract Host::ResolvConf and Route into separate files

kke:
`configuration/host.rb` is cluttered.


lib/pharos/autoload.rb
```diff
@@ -66,6 +66,8 @@ module Terraform
 
   module Configuration
     autoload :Host, 'pharos/configuration/host'
+    autoload :Route, 'pharos/configuration/route'
```
jakolehm: These are not used directly outside of `Host`... should `Host` just require them?

kke: Both are used by GatherFacts and neither are struct attributes of the Configuration::Host

PR #1066, opened 2019-02-12
Update Kontena Lens to 1.4.1

nevalla:
## Changelog
- Fixed app crash when container state has `null` value
- parameters field is handled properly on Storage Class details and not causing app crash any more


nevalla:
Depending on: https://github.com/kontena/pharos-kube-zipper/pull/84

nevalla:
Docs PR: https://github.com/kontena/pharos-docs/pull/158

PR #1067, opened 2019-02-12
Allow to install latest release/revision for given cri-o version

jakolehm:
Applies to debian/ubuntu.. centos/rhel already had this,.


jakolehm:
@jnummelin @kke PTAL

PR #1078, opened 2019-02-15
Also run OSS specs during NON-OSS spec run

kke:
Runs the specs from the main oss directory for non-oss.

Previously only the non-oss specs were run when `PHAROS_NON_OSS=true`.

Only two specs had to be made conditional. I'm not sure if this is a good or a bad sign.


non-oss/spec/pharos_pro/phases/validate_version_spec.rb
```diff
@@ -1,7 +1,8 @@
-require 'pharos/phases/validate_version'
-require 'pharos_pro/phases/validate_version'
+describe Pharos::Phases::ValidateVersion, unless: Pharos.oss? do
+  before :all do
+    require 'pharos_pro/phases/validate_version'
```
kke: Hmm actually this and the require for the oss-phase may need to be `load`, otherwise it may become flaky based on spec run order.

PR #1079, opened 2019-02-15
Generic env handling for calico, including enabling metrics

### LINKED ISSUE #1031: Expose metrics from calico node
I would like to scrape metrics from the calico-node pods. Since calico-node runs on host network, probably needs to have config to enable it.

I propose config: `network.calico.enable_metrics` to enable it. Things that need to be changed:

- [Enable in felix](https://docs.projectcalico.org/v3.5/reference/felix/configuration) by setting FELIX_PROMETHEUSMETRICSENABLED to true.
- Optionally set FELIX_PROMETHEUSMETRICSPORT from another config item (or stick with default 9091 for simplicity)
- Add port to container ports.
- Add a headless service that selects those pods (I could do that myself if pharos doesn't).

jnummelin:
Calico has lot of config env vars which can be configured. This PR enables to set env vars for the calico daemonset. For example:
```yaml
network:
  provider: calico
  pod_network_cidr: 172.31.0.0/16
  service_cidr: 172.32.0.0/16
  calico:
    environment:
      FELIX_PROMETHEUSMETRICSENABLED: true
      FELIX_PROMETHEUSMETRICSPORT: 9999
```

Metrics is a bit special case, the calico phase figures out that metrics is enabled and goes also automatically and sets up the needed headless service.

Needs still:

- [x]  docs PR

fixes #1031 

docs PR at https://github.com/kontena/pharos-docs/pull/160

### DIFF
--- a/lib/pharos/config_schema.rb
+++ b/lib/pharos/config_schema.rb
@@ -118,6 +118,7 @@ module Pharos
           optional(:calico).schema do
             optional(:ipip_mode).filled(included_in?: %w(Always CrossSubnet Never))
             optional(:nat_outgoing).filled(:bool?)
+            optional(:environment).filled(:hash?)
           end
           optional(:custom).schema do
             required(:manifest_path).filled(:str?)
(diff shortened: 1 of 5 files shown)


captncraig:
Great general solution to my problem, and probably some future ones I would have brought up.

PR #1087, opened 2019-02-19
pharos tf apply/destroy

jakolehm:
Automates `terraform apply` + `terraform output -json` + `pharos up`. Also adds logic to handle Terraform workspaces.

## Cluster provisioning:
```yaml
$ pharos tf apply --workspace production-eu-west-1
```

## Cluster destroy:
```
$ pharos tf destroy --workspace production-eu-west-1
```


lib/pharos/phase_manager.rb
```diff
@@ -11,7 +11,8 @@ class PhaseManager
       Excon::Error,
       K8s::Error,
       Pharos::SSH::RemoteCommand::ExecError,
-      Errno::ECONNRESET
+      Errno::ECONNRESET,
+      Errno::ECONNREFUSED
```
jakolehm: We need to retry these because hosts might be still booting when we execute `up`.

lib/pharos/terraform/base_command.rb
```diff
@@ -0,0 +1,33 @@
...
+require 'open3'
+
+module Pharos::Terraform
+  class BaseCommand < Pharos::Command
+    options :yes?
+
+    option "--workspace", "NAME", "terraform workspace", default: "default"
+
+    def tf_workspace
+      unless run_cmd("terraform workspace select #{workspace} 2> /dev/null").zero?
```
jakolehm: Workspaces make multi-cluster setups easier to handle (same config for all clusters).

lib/pharos/terraform/apply_command.rb
```diff
@@ -0,0 +1,39 @@
...
+      cmd << "-auto-approve" if yes?
+      cmd = cmd + var_list.map { |var| "-var #{var}" } if var_list
+
+      run_cmd! cmd.join(' ')
+    end
+
+    def pharos_up
+      run_cmd "terraform output -json > .#{workspace}.json"
+      cmd = ['--tf-json', ".#{workspace}.json", '-c', config_yaml.filename]
+      cmd << '-y' if yes?
```
kke: `--force` should probably be "proxied" too

jakolehm:
@kke @jnummelin PTAL

lib/pharos/command_options/tf_json.rb
```diff
@@ -0,0 +1,46 @@
...
+        # @param config [Hash]
+        # @return [Hash]
+        def load_terraform(file, config)
+          puts(pastel.green("==> Importing configuration from Terraform ...")) if $stdout.tty?
+
+          tf_parser = Pharos::Terraform::JsonParser.new(File.read(file))
+          config['hosts'] ||= []
+          config['api'] ||= {}
+          config['addons'] ||= {}
+          config['hosts'] += tf_parser.hosts
```
kke:
```suggestion
          config['hosts'].concat(tf_parser.hosts)
```
would avoid intermediate array

lib/pharos/command_options/tf_json.rb
```diff
@@ -0,0 +1,46 @@
...
+          config['api'].merge!(tf_parser.api) if tf_parser.api
+          config['addons'].each do |name, conf|
```
kke: This won't add any terraform addon configs from scratch, only if the cluster.yml has a stub for it? Also, if the addon-config has any hashes or arrays, the merge will overwrite them.

lib/pharos/terraform/apply_command.rb
```diff
@@ -0,0 +1,43 @@
...
+        cmd << "-var-file #{var_file}" if var_file
+        cmd += var_list.map { |var| "-var #{var}" } if var_list
+
+        run_cmd! cmd.join(' ')
+      end
+
+      def pharos_up
+        run_cmd "terraform output -json > .#{workspace}.json"
+        cmd = ['--tf-json', ".#{workspace}.json", '-c', config_yaml.filename]
+        cmd << '-y' if yes?
```
kke: `yes?` needs the `:yes?` option

lib/pharos/terraform/apply_command.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require_relative 'base_command'
+
+module Pharos
+  module Terraform
+    class ApplyCommand < BaseCommand
+      options :load_config
```
kke:
```suggestion
      options :load_config, :yes?
```

lib/pharos/terraform/base_command.rb
```diff
@@ -0,0 +1,36 @@
...
+      def run_cmd!(cmd)
+        code = run_cmd(cmd)
+        signal_error "#{cmd} failed" unless code.zero?
+      end
+
+      # @param cmd [String]
+      # @return [Integer]
+      def run_cmd(cmd)
+        system(cmd)
+        $CHILD_STATUS.exitstatus
```
kke: if just checking for `zero?`, the `system `return is `true` when exit status is zero.

lib/pharos/terraform/apply_command.rb
jakolehm: No it does not, it's already enabled in `BaseCommand`.

kke: Ah, confused `BaseCommand` with `RootCommand`

lib/pharos/terraform/base_command.rb
jakolehm: Refactored to use only boolean status.

lib/pharos/command_options/tf_json.rb
jakolehm: Not part of this PR, I only refactored existing code to `TfJson` module. Feel free to open an issue if you see problems here.

PR #1071, opened 2019-02-12
Add hosts.ssh_port option

kke:
Fixes #1062 
Fixes #1072

(and will need to be reimplemented in #1044)

Adds `ssh_port` option to host options:

```yaml
hosts:
  - address: 10.0.0.1
    ssh_port: 2022
```

And bastion options:

```yaml
hosts:
  - address: 10.0.0.1
    ssh_port: 22
    bastion:
      address: 10.0.0.2
      ssh_port: 20202
```

Refactored the `Net::SSH::Gateway` from `SSH::Client` to `Configuration::Bastion` because `bastion.host.ssh.gateway` opens an extra SSH connection due to `.tap(&:connect)`.

TODO: docs pr

PR #1090, opened 2019-02-19
Add host.ssh_port option and refactor gateway initialization

### LINKED ISSUE #1062: Allow to configure port for SSH access
Allow to configure SSH port for hosts

**Why is this needed**:
It's possible that nodes are behind the same public ip, but they are listening to different ssh port

**Example cluster.yml**:
```yaml
hosts:
  - address: 10.2.2.1
     port: 2022
  - address: 10.2.2.1
     port: 2023
  ...
```

### LINKED ISSUE #1072: Bastion opens an extra ssh connection
When using a bastion host, unused extra ssh connections will be opened when creating a kube client. This is because:

```ruby
bastion.host.ssh.gateway
#         ^--- duplicate host
#             ^--- calls Pharos::SSH::Client.new.tap(&:connect)
#                  ^--- calls Net::SSH.start
```

The extra connection does nothing, but I assume a passphrase/password for it will be prompted separately (twice?).

### LINKED ISSUE #1088: K8s api connection should always be tunneled via ssh

`pharos up` should always use ssh tunnel for k8s api access.

**Why is this needed**:

It would help cases where api port is blocked or where complex *_PROXY envs are used (for example #1075).

kke:
Fixes #1062
Fixes #1072
Fixes #1088 

Replaces #1071 

Adds `ssh_port` option to host options:

```yaml
hosts:
  - address: 10.0.0.1
    ssh_port: 2022
```

And bastion options:

```yaml
hosts:
  - address: 10.0.0.1
    ssh_port: 22
    bastion:
      address: 10.0.0.2
      ssh_port: 20202
```

Refactored the `Net::SSH::Gateway` from `SSH::Client#gateway` to `Transport::Gateway` for better control (should benefit the disconnecting in #1003) and because `bastion.host.ssh.gateway` opened extra SSH connections.

Also adds the missing `ssh_proxy_command` attribute to bastion.

All k8s-client traffic is now tunneled through ssh.

`ConfigureClient` phase is replaced with `WarmUpClientCache` phase.


lib/pharos/configuration/bastion.rb
```diff
@@ -1,14 +1,31 @@
 # frozen_string_literal: true
 
+require 'net/ssh/gateway'
+require 'net/ssh/proxy/jump'
```
kke: Maybe we don't need this jumping thing.

lib/pharos/configuration/host.rb
```diff
@@ -77,6 +78,15 @@ def short_hostname
         hostname.split('.').first
       end
 
+      def ssh_options
+        {}.tap do |opts|
+          opts[:keys] = [ssh_key_path] if ssh_key_path
+          opts[:send_env] = [] # override default to not send LC_* envs
+          opts[:proxycommand] = ssh_proxy_command if ssh_proxy_command
```
kke: This should do the same as `opts[:proxy] = Net::SSH::ProxyCommand.new(...)`: https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/config.rb#L251

lib/pharos/transport.rb
```diff
@@ -1,20 +1,12 @@
 # frozen_string_literal: true
 
-require 'net/ssh'
-require 'net/ssh/proxy/jump'
```
kke: I don't think the `Net::SSH::Proxy::Jump` was actually used

kke:
PTAL

kke:
This probably still needs some way to track clients that are using the gateway, otherwise it will shutdown a gateway that is still used by some other client.

#1088 will likely need something from this.

kke:
Files changed: 28. Seems like this grew out of proportion.

lib/pharos/config_schema.rb
```diff
@@ -78,15 +78,18 @@ def self.messages
                   required(:effect).filled(included_in?: ['NoSchedule', 'NoExecute'])
                 end
               end
-              optional(:user).filled
-              optional(:ssh_key_path).filled
-              optional(:ssh_proxy_command).filled
+              optional(:user).filled(:str?)
+              optional(:ssh_key_path).filled(:str?)
+              optional(:ssh_proxy_command).filled(:str?)
+              optional(:ssh_port).filled(:int?, gt?: 0, lt?: 65_536)
```
kke: Rubocop wants `65536` as `65_536`

lib/pharos/transport/command/ssh.rb
```diff
@@ -12,13 +12,13 @@ def hostname
 
         # @return [Pharos::Transport::CommandResult]
         def run
-          raise Pharos::ExecError, "Connection not established" unless @client.connected?
+          raise Pharos::ExecError.new(@source || cmd, -127, "Connection not established") unless @client.connected?
```
kke: This and the identical one below have been broken forever, but the only way to reach it is via console or if writing a new command that does not `GatherFacts` before operations.

lib/pharos/transport/ssh.rb
```diff
@@ -14,61 +13,44 @@ class SSH < Base
...
+        super(host)
+        @gateway = gateway
+        @port = port || host.ssh_port
       end
 
-      # @return [Hash,NilClass]
-      def bastion
-        @bastion ||= @opts.delete(:bastion)
+      def to_s
+        "SSH #{host.user}@#{host.address}:#{host.ssh_port}#{' using proxy command' if host.ssh_proxy_command}#{" via #{@gateway}->127.0.0.1:#{@port} " if @gateway}"
```
kke: image

kke:
E2e now uses bastion and proxy command

jakolehm:
Conflict..

kke:
Unconflict..

lib/pharos/addon.rb
```diff
@@ -269,6 +267,10 @@ def kube_stack(**vars)
       )
     end
 
+    def kube_client
```
jakolehm: Yardoc

lib/pharos/configuration/host.rb
```diff
@@ -18,23 +19,85 @@ class Host < Pharos::Configuration::Struct
...
 
+      # @return [String,nil]
       def short_hostname
         return nil unless hostname
 
         hostname.split('.').first
       end
 
+      # @return [Hash]
+      def ssh_options
```
jakolehm: IMO this should be under `Transport` (for example `Transport::SshHelpers` module)

lib/pharos/configuration/host.rb
```diff
@@ -18,23 +19,85 @@ class Host < Pharos::Configuration::Struct
...
+        {}.tap do |opts|
+          opts[:keys] = [ssh_key_path] if ssh_key_path
+          opts[:send_env] = [] # override default to not send LC_* envs
+          opts[:proxy] = Net::SSH::Proxy::Command.new(ssh_proxy_command) if ssh_proxy_command && !bastion
+          opts[:port] = ssh_port
+        end
+      end
+
+      # @return [Pharos::Transport::Gateway]
+      def gateway
```
jakolehm: Not completely happy about the idea that `Host` is becoming a "god" object that knows everything. For example `gateway` requires `host` and `Host` requires `Gateway` because of `kube_client` 🤔 /cc @jnummelin

jakolehm: Probably all ssh specific stuff should be under Transport (somehow).

jnummelin: Yes, I agree. Now `Host`, `Bastion`, `Gateway` all entangle pretty badly and really hard to follow which is used where.

kke: I moved stuff around.

Now there's `Pharos::Kube::Client` that understands `Configuration::Host`, so that host can just do `@kube_client ||= Pharos::Kube::Client.new(self)`.

The host ssh options are generated in `Pharos::Transport::SSH.options_for(host)`.

The `Host` now has:
- `def gateway`
- `def transport`
- `def kube_client`
- `def disconnect`

kke: Alternative to having these memoized in Host I guess would be to go back to 
`Pharos::SshManager.clients[host.address]` / `Pharos::GatewayManager.gateways[host.address]` type of connection collections or maybe have some sort of `Pharos::Transport` delegator/decorator and just have `host.transport` and all those `kube_client / exec / file / gateway`  stuff would be contained there.

jnummelin: IMO the latter sounds pretty good

jakolehm: I still think that `gateway` should be masked behind `transport`. Transport already gets host (like Gateway) as a parameter.

jakolehm: @kke @jnummelin are you fine with `gateway` being here?

kke: I'm now WIPping with https://github.com/kontena/net-ssh-proxy-gateway

kke:
#1191 fixes #1062
#1072 and #1088 can be fixed more simply separately now with #1185

PR #1094, opened 2019-02-21
Update the pharos pro license assign to use the new JWT based token

### LINKED ISSUE #1092: Store license jwt token to kubernetes api
Pharos licenses contains now license information encoded in jwt token. We should store this token to kubernetes api instead of current licence token.

`pharos license assign` command should send `uid` of `cluster_info` configmap in `kube-public` namepace as **`cluster_id`** and `cluster_name` stored in `cluster.yml` as **`description`** to licence management api. 

It should parse jwt token from response and store it in `license.jwt` in `pharos-license` secret in `kube-system` namespace.

**Why is this needed**:
We can validate the license better.

kke:
Fixes #1092

Also adds a `pharos license inspect` command for inspecting license tokens.


non-oss/pharos_pro/license_key.rb
```diff
@@ -0,0 +1,93 @@
+# frozen_string_literal: true
+
+require 'time'
+require 'jwt'
+require 'ostruct'
+
+module Pharos
+  class LicenseKey
+    def self.jwt_public_key
```
jnummelin: I don't think we need this on cli side at all. CLI does not need to verify the license, it can parse the details just vie base64 decoding.

PR #1099, opened 2019-02-21
Simplify the config option copying for kubeconfig hint

kke:
Replaces #1096

### DIFF
--- a/lib/pharos/terraform/apply_command.rb
+++ b/lib/pharos/terraform/apply_command.rb
@@ -32,8 +32,10 @@ module Pharos
 
       def pharos_up
         run_cmd "terraform output -json > .#{workspace}.json"
-        @config_options ||= []
-        cmd = @config_options.concat(['--tf-json', ".#{workspace}.json"])
+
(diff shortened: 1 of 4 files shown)


lib/pharos/terraform/apply_command.rb
```diff
@@ -32,7 +32,10 @@ def tf_apply
 
       def pharos_up
         run_cmd "terraform output -json > .#{workspace}.json"
-        cmd = ['--tf-json', ".#{workspace}.json", '-c', config_yaml.filename]
+
+        cmd = @config_options || []
```
jakolehm: I don't get this... why not an empty array always?

kke: `config_yaml.filename` is absolute and is there even when the default `cluster.yml` was used. The `@config_options` has the original param only if one was supplied.

kke: Of course it could be `cmd = []; cmd.concat(@config_options) if @config_options`

PR #1100, opened 2019-02-21
Add ipalloc_default_subnet config option to Weave

jakolehm:
Makes it possible to have unique ipam cidr per cluster on multi-cluster setups. For example Istio requires this.

### DIFF
--- a/lib/pharos/config_schema.rb
+++ b/lib/pharos/config_schema.rb
@@ -116,6 +116,7 @@ module Pharos
             optional(:trusted_subnets).each(type?: String)
             optional(:known_peers).each(type?: String)
             optional(:password).filled(:str?)
+            optional(:ipalloc_cidr).filled(:str?)
             optional(:no_masq_local).filled(:bool?)
           end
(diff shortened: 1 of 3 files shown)


lib/pharos/config_schema.rb
```diff
@@ -116,6 +116,7 @@ def self.messages
             optional(:trusted_subnets).each(type?: String)
             optional(:known_peers).each(type?: String)
             optional(:password).filled(:str?)
+            optional(:ipalloc_cidr).filled(:str?)
```
jakolehm: Just wondering if this is a good name... should it be `ipam_cidr` or something like that?

jnummelin: This is for true power-users anyway, so I'd maybe use something that close to the actual Weave flag getting set so it's easy to refer to their docs too.

jakolehm: `ipalloc_default_subnet:`?

jnummelin: 👍

PR #1102, opened 2019-02-22
Modify addon DSL loader to fix constant namespacing issues

### LINKED ISSUE #562: Add-on constants are not namespaced

kke:
Fixes #562 
Replaces #563 (which would be a better choice)
Replaces #564 
Is #uglyhack 

Uses a dirty regex to replace `Pharos.addon` with `class Pharos::Addons::#{addon_name.camelcase}`

This fixes constant namespacing problems in addons:

```
Pharos.addon "foofoo" do
  FILENAME = "foo"
end

Pharos.addon "barbar" do
  FILENAME = "bar"
end
Warning: Already defined constant FILENAME

Object::FILENAME # => "foo"
Pharos::Addons::Barbar::FILENAME # NameError: uninitialized constant Pharos::Addons::Barbar::FILENAME
```

Does not remove the `Pharos.addon` method, it can be used as previously, just the `addon_manager.load_addons` has been modified.

This does not touch addons defined like:

```ruby
Pharos.addon "foofoo" {
 ...
}
```

because bracket-matching is difficult. Only ones with `do ... end`.

Even though it's using `eval`, securitywise it brings no change. The files could have contained `system("rm -rf /")` that would have been evaluated before as well.

### DIFF
--- a/lib/pharos/addon_manager.rb
+++ b/lib/pharos/addon_manager.rb
@@ -8,6 +8,7 @@ module Pharos
   class AddonManager
     include Pharos::Logging
     using Pharos::CoreExt::DeepTransformKeys
+    using Pharos::CoreExt::StringCasing
 
     class InvalidConfig < Pharos::Error; end
     class UnknownAddon < Pharos::Error; end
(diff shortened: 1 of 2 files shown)


kke:
Dropping the DSL would be easier and better and only require to change one line in any old addons (which can be instructed in error / deprecation warning message). #1104 is a new take on #563

jakolehm:
Closing because we have #1104 and #1108.

PR #1108, opened 2019-02-25
Load addons using AddonContext instance_eval

### LINKED ISSUE #562: Add-on constants are not namespaced

jakolehm:
Fixes #562 
Closes #1102
Closes #1104

Only downside is that addons cannot bring their own refinements.. but IMO that's acceptable trade-off.


kke:
I fixed some problems and dropped the unnecessary `Pharos::Addons::Foofoo` magic

lib/pharos/addon.rb
```diff
@@ -11,18 +11,15 @@ module Pharos
   # @param name [String]
   # @return [Pharos::Addon]
   def self.addon(name, &block)
-    klass = Class.new(Pharos::Addon, &block).tap do |addon|
+    Class.new(Pharos::Addon, &block).tap do |addon|
       addon.addon_location = File.dirname(block.source_location.first)
```
jakolehm: `block.source_location` is empty which makes this PR useless (no resources are actually applied because addon cannot find any yamls). E2E tests did show this ...

jakolehm:
@kke PTAL

PR #1110, opened 2019-02-26
Update outdated colorizations brought in by license PR 

kke:
There were some `pastel.xyz` usages in the LicenseInspectCommand that were not using the Colorize refinement.


kke:
Maybe we need to drop the inspect command.

Or maybe it should get the JWT from the cluster. Maybe it shouldn't call the JWT a "license key", because license assign calls the UUID's license keys.

If it accepts the UUID it needs to POST it to cloud and that will assign it. So, license inspect was pretty much doing everything license assign does, except update the secret. I changed it to only validate JWT and it doesn't even do that, it only validates the contents, not the signature.

Multi-fail.

kke:
> Maybe we need to drop the inspect command.

#1131 now defines how `license inspect` should actually work, looking into it.

nevalla:
@kke can we close this?

kke:
New inspect command coming

PR #1112, opened 2019-02-26
Export complete HTTP proxy environment

madddi:
At the moment only the following environment variables are exported during the setup and in systemd:

* `HTTP_PROXY`
* `HTTPS_PROXY`
* `NO_PROXY`
* `http_proxy`

This leads to weird behavior when http libraries prefer `http_proxy` over `HTTP_PROXY` and then ignore `NO_PROXY` because `no_proxy` is not set. We should either set all variants of capitalization or only one.

### DIFF
--- a/lib/pharos/scripts/pharos.sh
+++ b/lib/pharos/scripts/pharos.sh
@@ -62,12 +62,14 @@ configure_container_runtime_proxy() {
   systemd_proxy_config_file="${systemd_runtime_dir}/http-proxy.conf"
   mkdir -p "${systemd_runtime_dir}"
   # shellcheck disable=SC2154
-  if [ -n "$HTTP_PROXY" ] || [ -n "$HTTPS_PROXY" ] || [ -n "$NO_PROXY" ] || [ -n "$http_proxy" ] ||  [ -n "$FTP_PROXY" ]; then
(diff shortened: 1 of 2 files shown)


lib/pharos/scripts/pharos.sh
```diff
@@ -62,12 +62,14 @@ configure_container_runtime_proxy() {
   systemd_proxy_config_file="${systemd_runtime_dir}/http-proxy.conf"
   mkdir -p "${systemd_runtime_dir}"
   # shellcheck disable=SC2154
-  if [ -n "$HTTP_PROXY" ] || [ -n "$HTTPS_PROXY" ] || [ -n "$NO_PROXY" ] || [ -n "$http_proxy" ] ||  [ -n "$FTP_PROXY" ]; then
+  if [ -n "$HTTP_PROXY" ] || [ -n "$HTTPS_PROXY" ] || [ -n "$NO_PROXY" ] || [ -n "$http_proxy" ] || [ -n "$https_proxy" ] || [ -n "$no_proxy" ] || [ -n "$FTP_PROXY" ]; then
```
kke: You can probably replace the whole thing with something like:
```
(env | cut -d"=" -f1|grep -i -- "_proxy$") | while read -r var; do 
  [ -n "${!var}" ] && echo "Environment=\"${var}=${!var}}\"" >> "${systemd_proxy_config_file}"
done
```

kke: But I guess it's fine the way it is. It can be beautified at a later time if desired.

madddi: I think the reason for this is line 66:
```
      echo "[Service]" > "${systemd_proxy_config_file}"
```

We'd have to check if one of the variables is set, but the inner part can be enhanced.

PR #1114, opened 2019-02-27
Add dns-node-cache

jakolehm:
Fixes DNS issues that some clusters are having (see: https://github.com/kubernetes/kubernetes/issues/56903).

Requires #1113 
Requires https://github.com/kontena/pharos-kube-zipper/pull/91

### DIFF
--- a/lib/pharos/configuration/network.rb
+++ b/lib/pharos/configuration/network.rb
@@ -5,6 +5,8 @@ require 'ipaddr'
 module Pharos
   module Configuration
     class Network < Pharos::Configuration::Struct
+      CLUSTER_DNS = '169.254.20.10'
+
       class Weave < Pharos::Configuration::Struct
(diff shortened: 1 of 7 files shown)


jakolehm:
@nevalla PTAL

jakolehm:
> Should we add `addonmanager.kubernetes.io/mode=Reconcile` and `kubernetes.io/cluster-service: "true"` labels to manifests?

We don't run addon manager so `addonmanager.kubernetes.io/mode=Reconcile` IMO is not needed. Not sure about `kubernetes.io/cluster-service: "true"`... this is not exactly a service 🤔

nevalla:
Okay...after googling found out that add-on manager used to use that label, but it's now deprecated and only use of this label is that `kubectl cluster-info` will show addresses of the services using this label. So the conclusion is that we don't need those.

PR #1115, opened 2019-02-27
Fix license-enforcer addon manifests

jakolehm:


jakolehm:
@nevalla PTAL (drone error is not related to this PR)

PR #1125, opened 2019-02-28
Run all phases in parallel mode

kke:
#805 all over again

Drops the `parallel: true/false` phase parameter and makes them all run in parallel mode.

The phases that had `parallel: false` have been modified to utilize the `Pharos::Phase#mutex` for the parts that (I assume) need to be run synchronously.

There's only one non-host phase-parameter left and that is the weird `Phases::ConfigureClient`'s 
 `optional: true/false` (which gets dropped in #1090)


lib/pharos/phases/configure_master.rb
```diff
@@ -19,20 +19,24 @@ def install?
...
-        if install?
-          logger.info { "Kubernetes control plane is not initialized." }
-          install
-          install_kubeconfig
-        elsif !cluster_context['api_upgraded']
-          reconfigure
-        else
-          logger.info { "Kubernetes control plane is up to date." }
-        end
+        Thread.current.abort_on_exception = true
```
jakolehm: I don't like that Phase needs to configure these.

kke: I think it could safely be enabled for all phases.

lib/pharos/phase_manager.rb
```diff
@@ -29,32 +30,20 @@ def initialize(**options)
...
+          Thread.new do
+            Thread.current.report_on_exception = false
+            Thread.current.abort_on_exception = true
+            Retry.perform(yield_object: phase, logger: logger, exceptions: RETRY_ERRORS, &block)
+          end
+        }
+        threads.map(&:value)
+      rescue StandardError
+        threads.map(&:kill)
+        sleep 0.5 until threads.all?(&:stop?)
```
kke: Without this the threads often output some lines after the exception is printed.

kke:
#1142 has some conflicting changes to the thread management that should probably be applied first.

PR #1126, opened 2019-02-28
Validate and apply addons as phases

kke:
Addons are now validated and applied via regular phases `ValidateAddons` and `ConfigureAddons`.


kke:
image

Maybe a bit christmas tree-y

I think could also generate one phase per addon.

kke:
I think a variation of this is going to be implemented in the future

PR #1128, opened 2019-03-01
Handle non-oss cluster-info migration

jakolehm:
Output:
```
==> Starting to craft cluster ...
==> Migrate master @ default-master-0
    [default-master-0] Migrating cluster info ...
    [default-master-0] Completed phase in 0.13s
```

### DIFF
--- a/lib/pharos/phases/mixins/cluster_version.rb
+++ b/lib/pharos/phases/mixins/cluster_version.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+module Pharos
+  module Phases
+    module Mixins
+      module ClusterVersion
+        # @return [Gem::Version]
+        def existing_version
+          if cluster_version = cluster_context['existing-pharos-version']
+            build_version(cluster_version)
+          else
+            build_version('0.0.1')
+          end
+        end
+
+        # @return [Gem::Version]
+        def pharos_version
+          @pharos_version ||= build_version(Pharos::VERSION)
+        end
+
+        # @param version [String]
+        # @return [Gem::Version]
+        def build_version(version)
+          Gem::Version.new(version.gsub(/\+.*/, ''))
+        end
+      end
+    end
+  end
+end
--- a/non-oss/pharos_pro/phases/migrate_master.rb
+++ b/non-oss/pharos_pro/phases/migrate_master.rb
@@ -34,7 +34,7 @@ module Pharos
         )
         cm_client.create_resource(new_cluster_info)
       rescue K8s::Error::NotFound
-        logger.error { "cluster-info configmap has gone, skipping the migration ..." }
+        logger.info "cluster-info configmap has gone, skipping the migration ..."
       end
     end
   end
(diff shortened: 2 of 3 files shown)


non-oss/pharos_pro/phases/migrate_master.rb
```diff
@@ -1,12 +1,40 @@
...
+          apiVersion: cluster_info.apiVersion,
+          metadata: {
+            name: cluster_info.metadata.name,
+            namespace: cluster_info.metadata.namespace
+          },
+          data: cluster_info.data.to_h
+        )
+        cm_client.create_resource(new_cluster_info)
+      rescue K8s::Error::NotFound
+        logger.error { "cluster-info configmap has gone, skipping the migration ..." }
```
kke: Where has it gone? If this is just a check if something needs to be done, does this need to be an error?

jakolehm: Good point, it could be `info` level message.

jakolehm: Though not sure how well whole cluster functions if cluster-info has gone awol 🤔

kke: Also, logger block format messages only make sense for `debug` and even then if the message contains interpolation :)

jakolehm: > Also, logger block format messages only make sense for `debug` and even then if the message contains interpolation :)

old-habit-die-hard

lib/pharos/phases/mixins/cluster_version.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Pharos
+  module Phases
+    module Mixins
+      module ClusterVersion
+        # @return [Gem::Version]
+        def existing_version
+          if cluster_version = cluster_context['existing-pharos-version']
```
nevalla: module assumes `cluster_context` method exists, but I guess it's a fair assumption(?)

non-oss/pharos_pro/phases/migrate_master.rb
```diff
@@ -1,12 +1,40 @@
...
+          apiVersion: cluster_info.apiVersion,
+          metadata: {
+            name: cluster_info.metadata.name,
+            namespace: cluster_info.metadata.namespace
+          },
+          data: cluster_info.data.to_h
+        )
+        cm_client.create_resource(new_cluster_info)
+      rescue K8s::Error::NotFound
+        logger.info "cluster-info configmap has gone, skipping the migration ..."
```
kke: Almost all phase messages are Capitalized maybe add `The` to avoid capitalizing the config map name :)

```suggestion
        logger.info "The cluster-info configmap has gone, skipping the migration ..."
```

lib/pharos/phases/mixins/cluster_version.rb
jakolehm: This mixin is intended to use only with Phase... so I think it's fair assumption.

PR #1134, opened 2019-03-04
Set or generate names for clusters

### LINKED ISSUE #1080: Cluster.yml should have (optional) cluster name
We should probably generate name if it's not given.

kke:
Fixes #1080 

Allows to set cluster name in config:

```yaml
hosts:
  - address: 10.0.0.1
name: pharos-cluster-123
```

Or through parameters:
```
pharos up -c cluster.yml --name pharos-cluster-123
```

If one is not given, a random name will be generated.

If the cluster already has a name and the config defines a different name, an error will be reported unless `--force` is used (or the same name is set by using `--name`)

The name can be displayed in [kontena-lens](https://pharos.sh/docs/2-0/addons/kontena-lens.html) for example.

### DIFF
--- a/e2e/digitalocean/main.tf
+++ b/e2e/digitalocean/main.tf
@@ -92,6 +92,12 @@ resource "digitalocean_volume_attachment" "pharos_storage" {
   volume_id               = "${element(digitalocean_volume.pharos_storage.*.id, count.index)}"
 }
 
+output "pharos_cluster" {
+  value = {
+    name = "${var.cluster_name}"
+  }
+}
+
 output "pharos_hosts" {
   value = {
     masters = {
--- a/e2e/digitalocean/terraform.tfvars
+++ b/e2e/digitalocean/terraform.tfvars
@@ -12,3 +12,5 @@ worker_size = "8gb"
 image = "ubuntu-18-04-x64"
 
 region = "fra1"
+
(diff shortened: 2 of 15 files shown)


nevalla:
kubeadm ClusterConfiguration has `clusterName` option which we could populate too.

kke:
I think terraform should now work with:

```
output "pharos_cluster" {
  value = {
    name = "foofoofoo"
  }
}
```

Added to e2e to see if it works.

kke:
Seems to work. Was wondering for a while where it got `4249-do` from, but it's in `terraform_setup.sh`:

```ruby
until terraform apply -auto-approve -var "cluster_name=${DRONE_BUILD_NUMBER}-do" -var "image=$1"
```

nevalla:
Docs need to be updated

nevalla:
Docs PR is here: https://github.com/kontena/pharos-docs/pull/164

nevalla:
I can potentially see need for renaming the cluster quite often. It's not obvious for user now that it is possible with `--force` option. I think we should at least hint about it.
> If the cluster already has a name and the config defines a different name, an error will be reported unless --force is used (or the same name is set by using --name)

kke:
Now it looks like:

image

PR #1137, opened 2019-03-04
Fix Host#local? raising IPAddr::InvalidAddressError

jakolehm:


captncraig:
Should local? also look at private_address if defined?

kke:
> Should local? also look at private_address if defined?

Maybe it should, at least for selecting if transport should connect via ssh or not.

jakolehm:
> Should local? also look at private_address if defined?

IMO no because private_address cannot be local... private_address is something that is used for peering with other hosts.

PR #1147, opened 2019-03-06
Read cluster name from pharos-config on license assign

### LINKED ISSUE #1143: License assign command should fetch cluster name from pharos-config configmap

nevalla:
This PR will improve license assign command in several ways:

1. it fixes cluster id in assign request payload: `cluster_id` -> `cluster-id`
2. it removes duplicated [LICENSE KEY] from help output
3. it reads cluster name from `pharos-config` configmap
4. it enhances output messages

Fixes #1143


nevalla:
@kke PTAL

non-oss/pharos_pro/license_key_command_option.rb
```diff
@@ -56,17 +56,26 @@ def cluster_info_configmap
...
         def cluster_id
           cluster_info.dig('metadata', 'uid')
         end
 
         def cluster_name
           Pharos::YamlFile.new(
-            StringIO.new(cluster_info.dig('data', 'kubeconfig')),
-            override_filename: "#{master_host.address}:kube-public/cluster-info.data.kubeconfig"
-          ).load.dig('clusters', 0, 'name')
+            StringIO.new(pharos_config_configmap),
```
kke: Hmm. Can't we just use `load_config.name` now as we have #1134 ?

nevalla: it would require that name is present in `cluster.yml` or user must pass it via options. I think it's more straightforward to read it from the cluster.

kke: Actually, the cluster name and the configmap would both be there if we run GatherFacts like we do in `pharos reset`:

```ruby
    def cluster_manager
      @cluster_manager ||= ClusterManager.new(load_config).tap do |cluster_manager|
        puts "==> Sharpening tools ...".green
        cluster_manager.load
        cluster_manager.gather_facts
      end
    end
```

After this, `cluster_manager.context['previous-config-map']` should contain the current `pharos-config` configmap and `load_config.name` will have the name from the cluster if it was not given.

kke: It will do some unnecessary extra stuff that is only required for Up, but it should also set the `config.master_host` to one that should be the most suitable.

Actually, maybe it should do:

```ruby
      @cluster_manager ||= ClusterManager.new(load_config.tap { |c| c.hosts.keep_if?(&:master)}).tap do |cluster_manager|
```

to only operate on masters.

nevalla:
@kke PTAL one more time.

non-oss/pharos_pro/commands/license_assign_command.rb
```diff
@@ -27,8 +29,108 @@ def execute
...
+                       end
+                       Pharos::LicenseKey.new(license_key, cluster_id: c_id)
+                     end
+    end
+
+    def master_host
+      @master_host ||= load_config.master_host
+    end
+
+    def cluster_info
```
kke: Maybe this should be in GatherFacts and store into cluster_context?

Anyway, as we now have gathered facts, we can do:
```
kube_client = load_config.kube_client(cluster_manager.context['kubeconfig'])
kube_client.api('v1').resource('configmaps', namespace: 'kube-public').get('cluster-info')&.metadata&.uid
```

It's now manipulated in the pro `MigrateMaster` phase.

non-oss/pharos_pro/commands/license_assign_command.rb
```diff
@@ -27,8 +29,108 @@ def execute
...
+    rescue Pharos::ExecError => ex
+      signal_error "Failed to get cluster-info configmap: #{ex.message}"
+    end
+
+    def cluster_id
+      cluster_info.dig('metadata', 'uid')
+    end
+
+    def cluster_name
+      load_config.name
```
kke: If no name exists, this is going to generate a new name and use it for now and it will be forgotten after that because it would be saved in the StoreClusterConfig phase when doing `up`.

kke: Though, you won't be doing `license assign` unless you have `up`ped already. Old clusters won't have a name (but they don't support the jwt either).

kke: New clusters will always have a name.

non-oss/pharos_pro/commands/license_assign_command.rb
nevalla: `cluster-info` is read using `kube_client` now

non-oss/pharos_pro/commands/license_assign_command.rb
```diff
@@ -27,8 +29,103 @@ def execute
...
+      @master_host ||= load_config.master_host
+    end
+
+    def cluster_info
+      @cluster_info ||= cluster_info_configmap
+    rescue K8s::Error::NotFound => ex
+      signal_error "Failed to fetch cluster-info: #{ex.class.name} : #{ex.message}"
+    end
+
+    def cluster_info_configmap
```
kke: I don't think this needs two methods

kke: Well, I guess it's easier to stub in test.

PR #1152, opened 2019-03-07
Allow to pass Lens admin password via envs

nevalla:
This PR makes it possible to pass `LENS_ADMIN_PASSWORD` to `pharos up` command and use the given password as admin password.

### DIFF
--- a/non-oss/pharos_pro/addons/kontena-lens/addon.rb
+++ b/non-oss/pharos_pro/addons/kontena-lens/addon.rb
@@ -189,7 +189,7 @@ Pharos.addon 'kontena-lens' do
 
   # @return [String]
   def admin_password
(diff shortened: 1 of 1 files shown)


nevalla:
@jakolehm WDYT?

Todo: docs

non-oss/pharos_pro/addons/kontena-lens/addon.rb
```diff
@@ -189,7 +189,7 @@ def token_authentication_webhook_config
 
   # @return [String]
   def admin_password
-    @admin_password ||= SecureRandom.hex(8)
+    @admin_password ||= ENV['LENS_ADMIN_PASSWORD'] || SecureRandom.hex(8)
```
jakolehm: Not sure if using ENV is good idea (because we haven't really used it) 🤔

PR #1155, opened 2019-03-08
Allow to enable feature gates

### LINKED ISSUE #1068: Allow to enable feature gates

Possibility to enable feature gates via `cluster.yml`.

**Example cluster.yml**:

```yaml
api:
  feature_gates: {}
kubelet:
  feature_gates: {}
```

nevalla:
This PR allows to configure feature gates in `cluster.yml`
```
...
control_plane:
  feature_gates:
    featureA: true
kubelet:
  feature_gates:
    featureA: true
...
```
feature gates for **control_plane** are passed as extraArgs for `apiServer`, `scheduler` and `controllerManager`  kubeadm components.

`kubeadm-reconfigure.sh` uploads now kubeadm config to cluster and new phase `ReconfigureKubelet` that will read kubelet config from cluster and reloads kubelet if needed.

fixes #1068

### DIFF
--- a/lib/pharos/config_schema.rb
+++ b/lib/pharos/config_schema.rb
@@ -94,7 +94,6 @@ module Pharos
         optional(:name).filled(:str?)
         optional(:api).schema do
           optional(:endpoint).filled(:str?)
-          optional(:feature_gates).filled
         end
         optional(:network).schema do
           optional(:provider).filled(included_in?: %w(weave calico custom))
@@ -189,6 +188,7 @@ module Pharos
         end
         optional(:control_plane).schema do
           optional(:use_proxy).filled(:bool?)
+          optional(:feature_gates).filled
         end
         optional(:telemetry).schema do
           optional(:enabled).filled(:bool?)
--- a/lib/pharos/config_schema.rb
+++ b/lib/pharos/config_schema.rb
@@ -94,6 +94,7 @@ module Pharos
         optional(:name).filled(:str?)
         optional(:api).schema do
           optional(:endpoint).filled(:str?)
+          optional(:feature_gates).filled
         end
         optional(:network).schema do
           optional(:provider).filled(included_in?: %w(weave calico custom))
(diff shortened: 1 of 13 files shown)


lib/pharos/config_schema.rb
```diff
@@ -94,6 +94,7 @@ def self.messages
         optional(:name).filled(:str?)
         optional(:api).schema do
           optional(:endpoint).filled(:str?)
+          optional(:feature_gates).filled
```
jnummelin: IMO this should be moved under `control_plane` key as these feature gates will now be set on all control plane components.

nevalla: Maybe the best solution is to define config options for all components.

jnummelin: If we can safely propagate all the control plane feature gates to all control plane components, assuming that a component will just ignore a feature it does not understand, I think having them defined on `control_plane` level makes users lifes so much easier.

nevalla: oh, we already have `control_plane` option

nevalla: moved api feature gates under `control_plane`

jnummelin:
Overall looks pretty good. I didn't do any real testing though yet.

lib/pharos/kubeadm/kubelet_config.rb
```diff
@@ -30,6 +30,9 @@ def generate
         if @config.kubelet&.read_only_port
           config['readOnlyPort'] = 10_255
         end
+        if @config.kubelet&.feature_gates
```
jakolehm:
```suggestion
        if feature_gates = @config.kubelet&.feature_gates
```

lib/pharos/kubeadm/kubelet_config.rb
```diff
@@ -30,6 +30,9 @@ def generate
...
+          config['featureGates'] = @config.kubelet.feature_gates
```
jakolehm:
```suggestion
          config['featureGates'] = feature_gates
```

PR #1165, opened 2019-03-12
Add Addon#to_s to improve spec error messages

kke:
Before:

```
3) #<Class:0x00007fa788b88d50>#install with given config applies stack with given values
```

After:

```
3) Openebs Addon#install with given config applies stack with given values
```

### DIFF
--- a/lib/pharos/addon.rb
+++ b/lib/pharos/addon.rb
@@ -22,12 +22,6 @@ module Pharos
     using Pharos::CoreExt::DeepTransformKeys
     include Pharos::Logging
 
-    class << self
-      def to_s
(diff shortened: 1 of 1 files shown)


lib/pharos/addon.rb
```diff
@@ -55,6 +55,10 @@ class << self
       attr_reader :addon_name
       attr_writer :addon_location
 
+      def to_s
+        "#{addon_name&.capitalize || 'Base'} Addon"
```
jakolehm:
```suggestion
        "#{addon_name&.capitalize || self.class.name} Addon"
```

jakolehm: @kke ping

PR #1166, opened 2019-03-12
PoC Addon supermagic

kke:
The loader takes `block.source_location` which contains the filename and starting line of the `Pharos.addon "xyz" do` block,

then `File.read`s the lines from that location into a string which has `class Xyz < Pharos::Addon` in the beginning,

then `module_eval`s that thing and sets source location.

This results in:

- The classes have names from the start
- Constants are namespaced
- The addon can be `require`d and ruby can worry about not loading duplicates
- Any `requires` or such that were before `Pharos.addon` are handled normally.
- **Backtraces have correct line numbers**:

```
#openebs/addon.rb:
....
36: install {
37:   raise 'foo'
38: }
```

```
  3) Pharos::Addons::Openebs#install with given partial config applies stack with default values
     Failure/Error: instance_exec(&hooks[:install])

     RuntimeError:
       foo
     # ./addons/openebs/addon.rb:37:in `block in <class:Openebs>'
```

But obviously this is quite ugly.

### DIFF
--- a/lib/pharos/addon.rb
+++ b/lib/pharos/addon.rb
@@ -8,13 +8,19 @@ require_relative 'addons/struct'
 require_relative 'logging'
 
 module Pharos
+  using Pharos::CoreExt::StringCasing
+
   # @param name [String]
   # @return [Pharos::Addon]
   def self.addon(name, &block)
-    Class.new(Pharos::Addon, &block).tap do |addon|
-      addon.addon_location = File.dirname(block.source_location.first)
-      addon.addon_name = name
-    end
+    definition = <<~ADDON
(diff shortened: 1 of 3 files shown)


lib/pharos/addon.rb
```diff
@@ -150,18 +156,18 @@ def apply_validate_configuration(old_config, new_config)
 
         case hook.arity
         when 1
-          hook.call(new_config)
+          instance_exec(new_config, &hook)
```
kke: These are unrelated, didn't bother wasting time cleaning it up.

kke:
Poc pocced. Adding reference to #562

PR #1171, opened 2019-03-13
Modify cluster manager apply_phase hosts param to accept a single host

kke:
Just a simple silly "refactoring" if you can even call it that.

Uses `*hosts` and `hosts.flatten!` in `apply_phase` to allow passing in a single host, a number of hosts or an array of hosts: `apply_phase(PhaseKlass, host)`, `apply_phase(PhaseKlass, host1, host2)` or `apply_phase(PhaseKlass, [host1, host2])`, all work exactly the same.

Now you don't need to force an array, like: `[config.master_host]`, `[etcd_hosts.first]` or `'%w(localhost)`, you can simply pass in the single host.

```ruby
def flat(*hosts)
  hosts.flatten!
  puts hosts.inspect
end

> flat(:a, :b, :c)
[:a, :b, :c]
> flat([:a, :b, :c]])
[:a, :b, :c]
> flat(:a)
[:a]
> flat
[]
```

### DIFF
--- a/lib/pharos/cluster_manager.rb
+++ b/lib/pharos/cluster_manager.rb
@@ -58,23 +58,23 @@ module Pharos
     def gather_facts
       apply_phase(Phases::ConnectSSH, config.hosts.reject(&:local?), parallel: false)
       apply_phase(Phases::GatherFacts, config.hosts, parallel: true)
-      apply_phase(Phases::ConfigureClient, [config.master_host], parallel: false, optional: true)
-      apply_phase(Phases::LoadClusterConfiguration, [config.master_host]) if config.master_host.master_sort_score.zero?
-      apply_phase(Phases::ConfigureClusterName, %w(localhost))
+      apply_phase(Phases::ConfigureClient, config.master_host, parallel: false, optional: true)
+      apply_phase(Phases::LoadClusterConfiguration, config.master_host) if config.master_host.master_sort_score.zero?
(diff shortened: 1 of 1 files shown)


kke:
Today I learned:

Given a method like:

```ruby
def foo(*args, **options)
  puts "args: %p options: %p" % [args, options]
end
```

And called with:

```ruby
b = Bar.new
foo(b)
```

makes ruby do:

```ruby
b.respond_to?(:to_hash) ? b.to_hash : b
```

Which results in `b` going into `**options` as a hash instead of into `*args`  as itself if `b` responds to `.to_hash`.

`Dry::Struct`s have `to_hash` and do not work very well with `YAML.dump` et al without it.

Closing as being unnecessary and broken.

kke:
Btw, this means that we COULD do something like:

```ruby
def ssh(**options)
  Net::SSH.start(options[:address], options[:user], port: options[:ssh_port])
end
```

and call it like: `ssh(config.hosts.first)`

Could be handy, but also very confusing.

kke:
FYI: This does not even require `**kwargs`, regular keyword arguments are enough.

```
def display(*hosts, color: true)
  puts "hosts: %p color: %p" % [hosts, color]
end
```

Calling this with a host object causes: `ArgumentError : unknown keywords: private_address, private_interface, role, user, ssh_key_path, container_runtime, bastion`

PR #1172, opened 2019-03-13
Calico 3.6

jnummelin:
I did a bit of refactoring on Calico manifests to make it easier to keep in sync with their official manifests:
- All CRDs combined into one single yaml file
- RBACs divided into two files, one for calico-node and one for calico-controller
  - Each file has all RBAC related things, roles, SA and binding
- CNI install moved to be an `initContainer` as it is in "official" manifests
- Adds new deployment for calico-controller

TODOs:

- [x] Use correct CDN for images


lib/pharos/resources/calico/20-configmap.yml.erb
```diff
@@ -7,6 +7,15 @@ metadata:
...
   ipv4_pool_cidr: "<%= ipv4_pool_cidr %>"
 
+
+
+  # Configure the Calico backend to use.
+  calico_backend: "bird"
+
+  # TODO Do we need mtu setting?
+  # Configure the MTU to use
+  veth_mtu: "1440"
```
jakolehm: How is this mtu different vs cni_network_config mtu?

jnummelin: By accident, good catch :) Fixed now.

jakolehm: Ok, but why do we need both?

jakolehm: > When using Kubernetes self-hosted manifests, the CNI plugin derives the MTU value from the veth_mtu field of the calico-config ConfigMap, and it is set to 1440 by default. On restart of the calico/node pods, any references to `__CNI_MTU__` are replaced by the veth_mtu value and inserted into the CNI configuration file (aka conflist) at the directory specified by Kubernetes (currently defaults to /etc/cni/net.d/).

jnummelin: Their reference manifests have it in both places so didn't really dare to remove.

lib/pharos/resources/calico/20-configmap.yml.erb
```diff
@@ -7,14 +7,11 @@ metadata:
...
 
-
-
   # Configure the Calico backend to use.
   calico_backend: "bird"
 
-  # TODO Do we need mtu setting?
   # Configure the MTU to use
-  veth_mtu: "1440"
+  veth_mtu: "1500"
```
jakolehm: Not sure if this is the correct number for all modes, we use ipip by default which most likely should use `1440` ..

lib/pharos/config_schema.rb
```diff
@@ -124,6 +124,7 @@ def self.messages
             optional(:ipip_mode).filled(included_in?: %w(Always CrossSubnet Never))
             optional(:nat_outgoing).filled(:bool?)
             optional(:environment).filled(:hash?)
+            optional(:mtu).filled(:int?, gt?: 0)
```
jakolehm: Requires docs pr.

jnummelin: Docs PR at https://github.com/kontena/pharos-docs/pull/172

PR #1185, opened 2019-03-15
Drop Net::SSH::Gateway dependency

kke:
The Net::SSH::Gateway class is [rather simple](https://github.com/net-ssh/net-ssh-gateway/blob/master/lib/net/ssh/gateway.rb), but still a bit inflexible.

For example, it does not allow reusing an existing connection and a new one will be always opened for each gateway.

Port forwarding on Net::SSH is quite simple:

> ### "Um, all I want to do is X, just show me how!"
> #### X == "forward connections on a local port to a remote host"
> ```
> Net::SSH.start("host", "user", :password => "password") do |ssh|
>   ssh.forward.local(1234, "www.google.com", 80)
>   ssh.loop { true }
> end
> ```

The IOError rescue retries are there to make it automatically reconnect both self + tunnel in case the tunnel has died. (The current connection does not know if the underlying tunnel has disconnected until it tries to send / receive)

This on its own doesn't fix anything (because it's missing all the host/bastion dereferencing et all), but having this could simplify #1090 and #1003

### DIFF
--- a/e2e/digitalocean/main.tf
+++ b/e2e/digitalocean/main.tf
@@ -119,6 +119,11 @@ output "pharos_hosts" {
       role              = "worker"
       user              = "root"
       ssh_key_path      = "./ssh_key.pem"
+      bastion = {
+        address           = "${digitalocean_droplet.pharos_master.*.ipv4_address[0]}"
+        ssh_key_path      = "./ssh_key.pem"
+        user              = "root"
+      }
 
       label = {
         "beta.kubernetes.io/instance-type"         = "${var.worker_size}"
--- a/lib/pharos/config.rb
+++ b/lib/pharos/config.rb
@@ -124,7 +124,7 @@ module Pharos
         api_port = 6443
       else
         api_address = 'localhost'
-        api_port = master_host.bastion.host.transport.gateway(master_host.api_address, 6443)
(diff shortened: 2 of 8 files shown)


kke:
I think the exception is different on Linux vs mac. 

I get these with `DEBUG_TRANSPORT=true`
```
D, [2019-03-15T15:19:32.414816 #43061] DEBUG -- Pharos::Transport::SSH[192.168.100.101]: Received Errno::EBADF (expected when tunnel has been closed)
D, [2019-03-15T15:19:32.414916 #43061] DEBUG -- Pharos::Transport::SSH[192.168.100.101]: Closed SSH event loop
D, [2019-03-15T15:19:32.414874 #43061] DEBUG -- Pharos::Transport::SSH[192.168.100.101]: Closed port forward on 65534
D, [2019-03-15T15:19:32.415257 #43061] DEBUG -- Pharos::Transport::SSH[192.168.100.102]: Received Errno::EBADF (expected when tunnel has been closed)
D, [2019-03-15T15:19:32.415334 #43061] DEBUG -- Pharos::Transport::SSH[192.168.100.102]: Closed SSH event loop
D, [2019-03-15T15:19:32.415294 #43061] DEBUG -- Pharos::Transport::SSH[192.168.100.102]: Closed port forward on 65533
```

Added IOError to the rescue and the exceptions seem to be gone from  e2e output now.

PR #1189, opened 2019-03-18
Allow to disable node local DNS cache

### LINKED ISSUE #1187: Node local dns cache may conflict with existing dns servers
Node local dns cache (introduced in 2.3.0-beta) may conflict with existing dns deployments. Pharos should probably give an option to disable it.

jnummelin:
Node local DNS cache can be disabled with:
```
network:
  node_local_dns_cache: false
```

`node_local_dns_cache` defaults to `true`.

fixes #1187 

Needs docs PR too.


jnummelin:
Docs PR at https://github.com/kontena/pharos-docs/pull/174

captncraig:
Does the kubelet config also need to change to point back at the kube-dns service ip if local cache isdisabled?

PR #1198, opened 2019-03-19
Validate required binaries / packages are installed on host

### LINKED ISSUE #1176: Misleading error if ping is not installed
**What happened**:

Tried to install pharos to Scaleway arm64 machine. Installation errored with:

```
ERROR: Pharos::InvalidHostError : Hostname 'localhost' does not seem to resolve to an address on the local host
```

but actually the root cause was missing `ping`:

```
# ping

Command 'ping' not found, but can be installed with:

apt install iputils-ping  
apt install inetutils-ping
```

Better error message or automatic installation of deps.

kke:
Fixes #1176

Adds simple validation to check for command availability on hosts. Ultimately I think it should be in host configurer and it should install the missing packages.

Also adds an alternative way to validate the localhost address resolves to an address on the local host.

### DIFF
--- a/lib/pharos/phases/validate_host.rb
+++ b/lib/pharos/phases/validate_host.rb
@@ -10,6 +10,8 @@ module Pharos
         check_role
         logger.info { "Validating distro and version ..." }
         check_distro_version
+        logger.info { "Validate required packages ..." }
+        check_installed
         logger.info { "Validating host configuration ..." }
         check_cpu_arch
         logger.info { "Validating hostname uniqueness ..." }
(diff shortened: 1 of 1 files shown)


lib/pharos/phases/validate_host.rb
```diff
@@ -50,10 +52,18 @@ def validate_unique_hostnames
         raise Pharos::InvalidHostError, "Duplicate hostname #{@host.hostname} for hosts #{duplicates.map(&:address).join(',')}"
       end
 
+      def check_installed
+        %w(ip grep curl uname hostname head cut sed).each do |cmd|
+          logger.debug { "Checking if command '#{cmd}' exists" }
+          raise Pharos::InvalidHostError, "Host does not seem to have the required binary '#{cmd}' installed." unless transport.exec?("command -v \"#{cmd}\" &> /dev/null")
```
jakolehm: I think we should just install those packages. Configurer could have `#install_prereqs` or similar method that we could call before we start to validating more deeply.

kke: Installing packages during validation phase sounds a bit wrong.

jakolehm: @jnummelin WDYT

kke: Perhaps with something like `--install-dependencies`?

jnummelin: Yeah, we should just install all the needed "tools". But maybe keep these tools installations bit separate from others as these are more for validation purposes and as @kke said, not optimal to install stuff on validation phase.

kke: Also many of these commands are used already before this validation happens.

kke:
Not good

PR #1199, opened 2019-03-19
Pro: pharos license assign for another cluster

### LINKED ISSUE #1194: Command "pharos license assign" should have "--cluster-id" option

kke:
Fixes #1194 

Makes it possible to assign a license to another cluster by giving `--cluster-id` and `--cluster-name`. The `--config` is not needed in that case and there will be no attempt to connect to the cluster.

```
$ pharos license assign --cluster-id abcd --cluster-name foo abcd-defg-123
FDSKJVNDSKJNVkjdsnvcksdjnckdsjnckjsd.CJDNkcjdsnkcjdnskcjn.dskvdsvdsv
$ pharos license assign --cluster-id abcd --cluster-name foo abcd-defg-123
ERROR: License is already in use
````

The `pharos license inspect` can output the cluster id and name with `--cluster-info`:

```
$ pharos license inspect -c cluster.yml --cluster-info
=> Sharpening tools ...
...
...
Cluster ID: 37cxxveb-4961-11e9-9xxx51-0yyy1f34da68
Cluster name: lively-sound-7972
```

### DIFF
--- a/non-oss/pharos_pro/commands/license_assign_command.rb
+++ b/non-oss/pharos_pro/commands/license_assign_command.rb
@@ -8,9 +8,8 @@ module Pharos
 
     parameter "[LICENSE_KEY]", "kontena pharos license key (default: <stdin>)"
     option '--description', 'DESCRIPTION', "license description [DEPRECATED]", hidden: true
-    option '--print-subscription-token', :flag, 'only output subscription token'
 
-    option '--cluster-id', '[ID]', 'request for a specified cluster id (requires --cluster-name)' do |cluster_id|
+    option '--cluster-id', '[ID]', 'request for a specified cluster id. outputs the subscription token. (requires --cluster-name)' do |cluster_id|
       @cluster_id_given = true
       cluster_id
     end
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by jakolehm** [comment]
@nevalla PTAL

**Comment by nevalla** [comment]
`--print-subscription-token` is a bit long flag to type. How about `--output|-o=token`?

**Comment by kke** [comment]
So, an option with only one possible value or some shorter flag?

**Comment by nevalla** [comment]
atm just a shorter flag. Name of the option came from kubectl
```
 -o, --output='': Output format. One of:
json|yaml|wide|name|custom-columns=...|custom-columns-file=...|go-template=...|go-template-file=...|jsonpath=...|jsonpath-file=...
```

**Comment by kke** [comment]
Should there then also be `--output default`?

Feels a bit weird to write something like:

```ruby
    option %w(-o --output), "FIELD", 'possible values: token', attribute_name: :output_field do |field|
      signal_usage_error "valid values for --output are 'token'" unless field == 'token'
      field
    end
```

kke:
Maybe drop the option and always just print the jwt when `--cluster-id` given

nevalla:
> Maybe drop the option and always just print the jwt when `--cluster-id` given

I'm with that. @jakolehm any opinions?

jakolehm:
> > Maybe drop the option and always just print the jwt when `--cluster-id` given
> 
> I'm with that. @jakolehm any opinions?

SGTM

PR #1200, opened 2019-03-19
Enforcer as static pods

jnummelin:


non-oss/pharos_pro/addons/pharos-license-enforcer/addon.rb
```diff
@@ -1,15 +1,71 @@
...
+    <<-EOF
+apiVersion: v1
+kind: Pod
+metadata:
+  name: pharos-license-enforcer
+  namespace: kube-system
+spec:
+  terminationGracePeriodSeconds: 30
+  hostNetwork: true
+  priorityClassName: system-cluster-critical
```
jakolehm: Not sure if this is needed but does not hurt either.

PR #1209, opened 2019-03-20
Pro: check license icebox state during validation

### LINKED ISSUE #1173: Icebox check on pharos up
`pharos up`  should probably check during validation if the cluster has been iceboxed and warn / abort / act accordingly.

`reset` may need to know it too?

kke:
Fixes #1173 

wip/poc/rfc

Didn't find any other validate phase that could be easily extended.

There should be some way to inject new phases into cluster manager from pro.

Also the check seems a bit flimsy and sounds like it could easily give false-positives.


kke:
Looks like it doesn't work either. Seemed to work on my machine.

PR #1211, opened 2019-03-20
Share existing SSH connections for port forwarding and detect multi-hop bastion

### LINKED ISSUE #1072: Bastion opens an extra ssh connection

### LINKED ISSUE #1206: Race condition when host runs commands while portforwarding
The event loop introduced by #1185 has a race condition with the `transport.exec`.

It works fine while the connection is doing nothing but forwarding, but if you try to run commands on a connection that has forwards, the event loop will crash / deadlock.

```ruby
> local_port = host.transport.forward(master.api_address, 6443)
> Excon.get("https://127.0.0.1:6443") # => ok
> host.transport.exec('ls -al') # => ok
> Excon.get("https://127.0.0.1:6443") # => deadlock or IOError
```

This seems to be because the `exec` is also running the event loop:

```
        # Runs the SSH event loop until the channel is no longer active. This is
        # handy for blocking while you wait for some channel to finish.
        #
        #   channel.exec("grep ...") { ... }
        #   channel.wait
        def wait
          connection.loop { active? }
        end
```

I think this can be fixed by changing the event loop to use mutex (since `transport.exec` is also run inside `synchronize { .. }`) and `session.process` (instead of `session.loop`):

```ruby
          Thread.new do
            Thread.current.report_on_exception = logger.level == Logger::DEBUG
            while @session
              synchronize do
                @session.process(0.05)
              end
              Thread.pass
            end
```

Looks like it works.

Currently this does not cause any problems because of the host duplication as mentioned in #1072 but needs to be fixed for/in #1088

kke:
Fixes #1072 

- Makes identical bastion hosts refer to the same object.
  * **Before:** a new ssh session to the bastion host was opened for every host using it.
  * **After:** only one ssh session to the bastion host is opened and each host session is just another port forward on the same connection.
- Makes `bastion.host` refer to actual hosts in the config if all the bastion's attributes match one of the config `hosts:` host's attributes (address, ssh_port, user, ssh key, ..)
  * **Before:** two sessions would have been opened
  * **After:** the existing session for the host is used
- Detects multi-hop bastion (this could be made to work but maybe it isn't worth it):
  * **Before:** actually the port forward is from the first hop to target.
  * **After:** the problem is detected and an error is reported
- Detects bastion address / credentials being identical with the parent host. 
  * **Before:** two sessions are opened
  * **After:** the problem is detected and an error is reported
  * **Note:** This is perhaps unnecessary and counter productive  as it does exactly what has been asked and it works. When using TF for example, it might be easier to just let the one double-connection slip than to try to get one of the hosts not have the bastion configured (tf json parser does not handle per-host customizations very well).

### DIFF
--- a/lib/pharos/config.rb
+++ b/lib/pharos/config.rb
@@ -33,6 +33,29 @@ module Pharos
       config = new(schema_data)
       config.data = raw_data.freeze
 
+      config.hosts.reject { |h| h.bastion.nil? }.each do |host_a|
+        host_b = config.hosts.find { |host| host_a.bastion.attributes_match?(host.attributes) }
+        next unless host_b
+
+        # Creating a new Pharos::Configuration::Bastion entry because otherwise it's an anonymous class
+        host_a.attributes[:bastion] = Pharos::Configuration::Bastion.new(host_a.bastion.attributes).tap { |bastion| bastion.host = host_b }
+
+        if host_a == host_b
+          raise Pharos::ConfigError, 'hosts' => { config.hosts.index(host_a) => ["host and its bastion point to the same address and port (#{host_a.address}:#{host_a.ssh_port})"] }
+        end
+
+        if host_b.bastion
+          raise Pharos::ConfigError, 'hosts' => { config.hosts.index(host_a) => ["multi-hop bastion not supported (#{host_a.address}:#{host_a.ssh_port} -> #{host_a.bastion.address}:#{host_a.bastion.ssh_port} -> #{host_b.bastion.address}:#{host_b.bastion.ssh_port}"] }
+        end
+      end
+
+      # de-duplicate bastions (if two hosts share an identical bastion, make it the same object to avoid multiple gateways)
+      config.hosts.reject { |h| h.bastion.nil? }.permutation(2) do |host_a, host_b|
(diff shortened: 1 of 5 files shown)


lib/pharos/config.rb
```diff
@@ -33,6 +33,29 @@ def self.load(raw_data)
       config = new(schema_data)
       config.data = raw_data.freeze
 
+      config.hosts.reject { |h| h.bastion.nil? }.each do |host_a|
+        host_b = config.hosts.find { |host| host_a.bastion.attributes.all? { |k, v| host.attributes[k] == v } }
```
jakolehm: Quite complex block, maybe could be refactored to a separate method?

kke: I used to have the `def ==(other)` in bastion but i think it looked more confusing:`host_a.bastion == host`

Maybe create some `attributes_equal?` instead of overriding `==`

jakolehm:
@kke rebase?

PR #1215, opened 2019-03-21
Fix autoloading libraries before they are needed

kke:
There were several leaks in the autoload system.

### Before

(with pry disabled from gemspec and `require` monkeypatched to print out requires)

```
$ bundle exec bin/pharos --version | wc -l
     908
$ PHAROS_NON_OSS=true bundle exec bin/pharos --version | wc -l
     916
```

### After

```
$ bundle exec bin/pharos --version | wc -l
      38
$ PHAROS_NON_OSS=true bundle exec bin/pharos --version | wc -l
      46
```


lib/pharos/transport/ssh.rb
```diff
@@ -12,30 +12,35 @@ class SSH < Base
         Net::SSH::Authentication::KeyManagerError
       ].freeze
 
-      # @param host [String]
-      # @param opts [Hash]
-      def initialize(host, **opts)
-        super(host, opts)
-        @user = @opts.delete(:user)
+      def to_s
```
jakolehm: Changes in this file are not relevant to this PR, right?

lib/pharos/core-ext/colorize.rb
```diff
@@ -18,7 +25,7 @@ def self.enabled?
       end
 
       refine String do
-        Pastel::ANSI::ATTRIBUTES.each_key do |meth|
+        ATTRIBUTES.each do |meth|
```
jakolehm: Is this really needed?

lib/pharos/transport/base.rb
```diff
@@ -20,12 +20,10 @@ class Base
 
       attr_reader :host
 
-      # @param host [String]
-      # @param opts [Hash]
-      def initialize(host, **opts)
+      # @param host [Pharos::Configuration::Host]
+      def initialize(host)
```
jakolehm: Not related to autoload.

lib/pharos/core-ext/colorize.rb
kke: Yes but the line below isn't anymore as `underscore` can be removed from the ATTRIBUTES array.

Alternatives would be to define each method delegator manually, add a `method_missing` delegator or going back to regular pastel.

lib/pharos/transport/base.rb
kke: We can bring back transport.rb and remove the requires.

lib/pharos/core-ext/colorize.rb
jakolehm: So what do we gain by copypasting Pastel attributes here?

kke: We get them without loading pastel which loads `tty-color` and `equatable`.

PR #1216, opened 2019-03-21
Fix autoload loading everything always

jakolehm:
Cleaner implementation of #1215

with patched require:
```
$ bundle exec ./bin/pharos -v | wc -l
      48
```


jakolehm:
@jnummelin @kke PTAL

kke:
Still loads 574 files.

jakolehm:
I think we should have follow-up pr that uses https://github.com/jarmo/require_all .

jakolehm:
Skipping Travis e2e because of bintray issues.

PR #1074, opened 2019-02-13
Replace cluster_context hash with a struct

kke:
Adds some robustness to the previously vague non-structured hash. Now there's a central place where all the cluster-context keys are listed.

Fixes #853 by refactoring the `kube_client` initialization

PR #1217, opened 2019-03-21
Move Configuration::Struct to configuration/struct.rb

kke:
The file `lib/pharos/configuration.rb` only defines `Pharos::Configuration::Struct` so it should be in `lib/pharos/configuration/struct.rb` by convention.


kke:
I'm not sure if it was in sync, need to check.

kke:
Oh it was a different PR, thought this was the #1074 with morning eyes.

PR #1218, opened 2019-03-22
Open SSH connections parallelly

kke:
The two-phase authentication was dropped in #1044 

This PR does the same but in one phase.

### DIFF
--- a/lib/pharos/transport/ssh.rb
+++ b/lib/pharos/transport/ssh.rb
@@ -7,6 +7,14 @@ module Pharos
     class SSH < Base
       attr_reader :session
 
+      def self.class_mutex
+        @class_mutex ||= Mutex.new
+      end
+
+      def class_mutex
+        self.class.class_mutex
+      end
+
       RETRY_CONNECTION_ERRORS = [
         Net::SSH::AuthenticationFailed,
         Net::SSH::Authentication::KeyManagerError,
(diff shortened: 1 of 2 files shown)


kke:
Actually this could be taken further by making an implementation of [`Net::SSH::Prompt`](https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/prompt.rb) that uses a mutex. It can be passed to `Net::SSH.start`. This way there needs to be no retry-magic and list of retriable authentication related exceptions.

kke:
There's a bug in Net::SSH that makes it ignore the `password_prompt:` option.

lib/pharos/transport/ssh.rb
```diff
@@ -35,12 +43,15 @@ def connect(**options)
           begin
             @session = session_factory.start(@host, @user, @opts.merge(options).merge(non_interactive: non_interactive))
             logger.debug "Connected"
+            class_mutex.unlock if class_mutex.locked? && class_mutex.owned?
```
jakolehm: There is already `synchronize` (at line 40)...?

kke: It's instance level.

An instance should not try to connect at the same time on multiple threads. Multiple instances can each connect at the same time because they don't share a state.

The class level mutex protects against any of the instances trying to prompt at the same time.

It's not using `class_mutex.synchronize` because you can't  call retry from a block (retry never returns).

jakolehm: Ok, makes sense.

PR #1222, opened 2019-03-26
Fix firewalld reload issues

jakolehm:
- use flock with reload (hopefully fixes random lockups)
- reload only if config changes
- reload triggers network provider deployment rollout (init container will actually reload firewalld rules)

### DIFF
--- a/lib/pharos/scripts/configure-firewalld.sh
+++ b/lib/pharos/scripts/configure-firewalld.sh
@@ -27,6 +27,6 @@ if ! firewall-cmd --query-masquerade > /dev/null 2>&1 ; then
 fi
 
 # reload only if this is first run
-if ! firewall-cmd --info-service pharos-worker 2&>1 /dev/null ; then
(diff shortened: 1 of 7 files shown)


lib/pharos/scripts/configure-firewalld.sh
```diff
@@ -7,11 +7,22 @@ if ! systemctl is-active --quiet firewalld; then
     systemctl start firewalld
 fi
 
-firewall-cmd --reload
+if ! firewall-cmd --info-service pharos-worker 2&>1 /dev/null ; then
```
jnummelin: I don't get what this actually does. If the `pharos-worker` service does not exist in firewalld configs, then reload it? Later on there's the same condition used to actually load the `pharos-worker` service. :/

jakolehm: We need to reload early because otherwise firewalld does not see those written config files and errors out.

jakolehm:
@kke @jnummelin PTAL

PR #1227, opened 2019-03-27
Kubernetes 1.13.5

jakolehm:
Release notes: https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG-1.13.md#changelog-since-v1134 .

Bumps also kubernetes-cni to 0.7.5.


jakolehm:
@jnummelin PTAL

jakolehm:
We had to take back CNI 0.7.5 packages... let's hold this PR until packaging is figured out.

PR #1228, opened 2019-03-27
Make LabelNode phase run on single host

kke:
The phase only uses the k8s api connection and does nothing on the hosts.

Also it was doing several unnecessary requests.


lib/pharos/phases/label_node.rb
```diff
@@ -3,63 +3,31 @@
 module Pharos
   module Phases
     class LabelNode < Pharos::Phase
-      title "Label nodes"
+      title "Configure node labels and taints"
 
       def call
-        if @host.labels.empty? && @host.taints.nil?
```
kke: Hosts always have labels.

PR #1229, opened 2019-03-27
Freeze kubernetes-cni version

jakolehm:
Should fix upcoming problems with CNI 0.7.5, like: https://github.com/kubernetes/kubernetes/issues/75683

### DIFF
--- a/lib/pharos/host/ubuntu/scripts/install-kube-packages.sh
+++ b/lib/pharos/host/ubuntu/scripts/install-kube-packages.sh
@@ -3,9 +3,9 @@
 set -e
 
 export DEBIAN_FRONTEND=noninteractive
-apt-mark unhold kubelet kubectl kubeadm || echo "Nothing to unhold"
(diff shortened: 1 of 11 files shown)


lib/pharos/host/ubuntu/scripts/install-kube-packages.sh
```diff
@@ -3,14 +3,14 @@
 set -e
 
 export DEBIAN_FRONTEND=noninteractive
-apt-mark unhold kubelet kubectl kubeadm || echo "Nothing to unhold"
-apt-get install -y "kubelet=${KUBE_VERSION}-00" "kubectl=${KUBE_VERSION}-00" "kubeadm=${KUBEADM_VERSION}-00"
-apt-mark hold kubelet kubectl kubeadm
+apt-mark unhold kubectl kubeadm kubernetes-cni || echo "Nothing to unhold"
```
jnummelin: we don't need to hold and un-hold cni at this step, right?

jakolehm: Good catch, fixed.

PR #1233, opened 2019-03-28
Always tunnel Kube API client through master ssh

### LINKED ISSUE #1088: K8s api connection should always be tunneled via ssh

kke:
Fixes #1088 

K8s Client is now always tunneled through the SSH connection to primary master host.

Done via making ConfigureClient to actually configure the client, not just load the configuration. 

If the master host is localhost, don't try to forward.

### DIFF
--- a/lib/pharos/addon.rb
+++ b/lib/pharos/addon.rb
@@ -186,19 +186,23 @@ module Pharos
       end
     end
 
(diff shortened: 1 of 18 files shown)


kke:
Worker up does not use k8s-client.

PR #1235, opened 2019-03-28
Pharos worker up command

### LINKED ISSUE #1086: Pharos worker up

`pharos worker up [--config] [--token] [--maybe-others]` sub-command that executes all the needed worker phases locally and joins a worker to the cluster.

**Why is this needed**:

Would make autoscaling etc use cases easier .

kke:
Fixes #1086 

Runs the minimal subset of phases to get a worker up.

```console
$ pharos worker up --help
Usage:
    pharos worker up [OPTIONS] JOIN_COMMAND ADDRESS ...

  Initialize a local or remote host as a Pharos cluster worker.

  Note:
    - Use pharos exec -r master -f sudo kubeadm token create --print-join-command
      to generate a value for the join-command parameter

Parameters:
    JOIN_COMMAND                  cluster join command (see note)
    ADDRESS ...                   master peer address (can be given multiple times)

Options:
    -y, --yes                     answer automatically yes to prompts
    -h, --host [ADDRESS]          ssh host address (default: "localhost")
    --ssh-key-path, -i [PATH]     ssh key path
    --ssh-port, -p [PORT]         ssh port (default: 22)
    --user, -l [USER]             ssh login username
    --insecure-registry [REGISTRY] insecure registry (can be given multipled times)
    --image-repository [IMAGE_REPOSITORY] image repository (default: "registry.pharos.sh/kontenapharos")
    --cloud-provider [NAME]       cloud provider
    --container-runtime [CONTAINER_RUNTIME] container runtime (default: "docker")
    --label [key=value]           node label (can be given multiple times)
    -e, --env KEY=VALUE           environment variable key=value (can be given multiple times)
    --control-plane-proxy         enable proxy for control plane
    --help                        print help
    --[no-]color                  colorize output (default: true)
    -v, --version                 print pharos version
    -d, --debug                   enable debug output (default: $DEBUG)
```


kke:
Rubocop fails, command works(?).

kke:
What about firewalld? There's too much to configure via options.

jakolehm:
> What about firewalld? There's too much to configure via options.

Can we load `cluster.yml` from kube api?

kke:
Should be no problem once the node is joined.

jakolehm:
> Should be no problem once the node is joined.

Actually it's not going to work because not all nodes are in the cluster.yml. I think we need to just document that firewalld is not enabled via `worker up`.

jakolehm:
Related issue #1086

jakolehm:
Regarding firewalld, I think `worker up` should not care about it... we might need to refactor functionality to DaemonSet at some point. @jnummelin WDYT?

lib/pharos/worker_up_command.rb
```diff
@@ -0,0 +1,110 @@
...
+    option %w(--user -l), '[USER]', 'ssh login username'
+    option '--insecure-registry', '[REGISTRY]', 'insecure registry (can be use multipled times)', multivalued: true
+    option '--container-runtime', '[CONTAINER_RUNTIME]', 'container runtime', default: 'cri-o'
+    option '--image-repository', '[IMAGE_REPOSITORY]', 'image repository', default: 'registry.pharos.sh/kontenapharos'
+    option '--label', '[key=value]', 'node label (can be used multiple times)', multivalued: true do |label|
+      signal_usage_error 'invalid --label format' unless label.include?('=')
+      label
+    end
+
+    option '--token', 'JOIN_TOKEN', 'run "sudo kubeadm token create --print-join-command" on master node to get a join token', required: true
```
kke: Maybe `pharos exec -r master sudo kubeadm token create --print-join-command`

kke: Actually replace these token switches with `--join-command` option

jnummelin:
I'm fine for `worker up` skipping FW at this point

kke:
It should pretty much work now, just needs some way to e2e it.

kke:
The `sudo kubeadm token create --print-join-command ` always seems to generate something like: `kubeadm join localhost:6443 --token xyz`, which of course isn't enough to figure out which address to connect to, but still it appears to work. I wonder if this is some sort of configuration error in the master, should it print something else than localhost?

I guess the regular `JoinNode` phase in `pharos up` magically uses the kubelet config which is generated from `@config` before running the join command, instead of / in addition to the passed in `localhost:6443`?

Anyway, the `worker up` has to get a master address from somewhere, so I added an option for it.

kke:
Should `cloud: provider:` be an option?

jakolehm:
> Should `cloud: provider:` be an option?

Yes. That said I think amount of options is quite big already.. maybe we need to come up with a config file in the future.

e2e/drone.sh
```diff
@@ -43,3 +43,58 @@ echo "Checking that kontena-lens is running:"
...
+if [ "${pharos_worker_host}" = "" ]; then
+  echo "Couldn't get pharos worker host from worker.json"
+  exit 1
+fi
+
+ssh_key="e2e/digitalocean/ssh_key.pem"
+ssh_userhost="root@${pharos_worker_host}"
+
+scp -o StrictHostKeyChecking=no -i "${ssh_key}" pharos-cluster*.gem "${ssh_userhost}:"
+ssh -o StrictHostKeyChecking=no -i "${ssh_key}" "${ssh_userhost}" -- "sudo http_proxy=http://10.133.37.156:8888 apt-get update"
```
jakolehm: Multiple `ssh` lines with same envs look really messy. Could we combine these to single `ssh` call?

e2e/drone_setup.sh
```diff
@@ -14,6 +14,15 @@ do
   echo "Apply failed... trying again in 5s"
   sleep 5
 done
-terraform output -json > tf.json
+terraform output -json > tf.json.tmp
+
+jq 'del(.pharos_hosts.value.worker_up)' tf.json.tmp > tf.json
```
jakolehm: Should this be within `if [ "$1" = "ubuntu-18-04-x64" ]; then` ?

e2e/drone.sh
```diff
@@ -43,3 +43,58 @@ echo "Checking that kontena-lens is running:"
...
+  --image-repository registry-tuusula.pharos.sh/kontenapharos \
+  -e "HTTPS_PROXY=http://10.133.37.156:8888" \
+  -e "HTTP_PROXY=http://10.133.37.156:8888" \
+  -e "NO_PROXY=localhost,0,1,2,3,4,5,6,7,8,9" \
+  -e "http_proxy=http://10.133.37.156:8888" \
+  -e "https_proxy=http://10.133.37.156:8888" \
+  -e "no_proxy=localhost,0,1,2,3,4,5,6,7,8,9" \
+  --join-command \""${join_command}"\" \
+  --master-ip "$(jq ".pharos_hosts.value.masters[0].address[0]" e2e/digitalocean/tf.json | sed 's/"//g')" \
+  --control-plane-proxy \
```
jakolehm: Why control-plane-proxy?

lib/pharos/worker_up_command.rb
```diff
@@ -0,0 +1,166 @@
...
+
+    option '--control-plane-proxy', :flag, 'enable proxy for control plane'
+
+    option '--join-command', 'CMD', 'cluster join command (see note)', required: true do |join_command|
+      signal_usage_error 'invalid --join-command' unless join_command.match?(/kubeadm join.*--token \S+/)
+
+      join_command
+    end
+
+    option %w(-m --master-ip), 'ADDRESS', 'master peer address (can be given multiple times)', required: true, multivalued: true
```
jakolehm: `--master-ip` is misleading because it can be also DNS address.

lib/pharos/worker_up_command.rb
```diff
@@ -0,0 +1,166 @@
...
+
+    option %w(-e --environment-variable), 'KEY=VALUE', 'environment variable key=value (can be given multiple times)' do |kv_pair|
+      @env ||= {}
+      @env.merge!(Hash[*kv_pair.split('=', 2)])
+    end
+
+    option '--control-plane-proxy', :flag, 'enable proxy for control plane'
+
+    option '--join-command', 'CMD', 'cluster join command (see note)', required: true do |join_command|
+      signal_usage_error 'invalid --join-command' unless join_command.match?(/kubeadm join.*--token \S+/)
```
jakolehm: Why require `kubeadm join`?

lib/pharos/worker_up_command.rb
```diff
@@ -0,0 +1,166 @@
...
+
+    def apply_stage2_phases
+      cluster_manager.apply_phase(Phases::ReconfigureKubelet, config.worker_hosts)
+    end
+
+    def update_masters
+      master_nodes = YAML.safe_load(host.transport.exec!('sudo kubectl get nodes --kubeconfig /etc/kubernetes/kubelet.conf -l node-role.kubernetes.io/master -o yaml'))
+      new_master_hosts = []
+      master_nodes['items']&.each do |master|
+        internal_ip = master.dig('status', 'addresses')&.select { |m| m['type'] == 'InternalIP' }&.map { |m| m['address'] }&.first
```
jakolehm: We need to check also `ExternalIP`.

e2e/drone_setup.sh
kke: The `worker_up` section needs to be removed always or the generated config isn't valid.

jakolehm: Maybe it should not be under `pharos_hosts` then?

kke:
PTALA

kke:
One option could be to use the `KUBE_SERVER/_TOKEN/_CA` stuff to configure a k8s-client and use that to figure out everything.

PR #1237, opened 2019-03-28
Fix firewalld reload

### LINKED ISSUE #1236: Firewalld fails to load rules on first run
**What happened**:

`pharos up` errors out on first run if `network.firewalld` is enabled.

Everything should just work.

Enable `network.firewalld` and run `pharos up` to a fresh cluster because control-plane port is not opened.

**Environment**:
- Pharos version (use `pharos --version`): 2.3.1

jakolehm:
Fixes #1236 

Improves full cluster e2e suite.


jakolehm:
@jnummelin @kke PTAL

PR #1238, opened 2019-03-28
Remove -00 version suffix madness from debs

jakolehm:
Also removes one deprecated file that I spotted file searching -00 occurrences.


kke:
This does make installing  `x.y.1` impossible if there's `x.y.10` available. Is that a problem?

jakolehm:
> This does make installing `x.y.1` impossible if there's `x.y.10` available. Is that a problem?

Ideas how to fix? Maybe `-*`?

kke:
`-*` or `-??` seem to work but I don't know if that fixes the problem this pr is addressing.

jnummelin:
AFAIK `-*` would allow to select the latest rev of the package

PR #1245, opened 2019-03-31
E2e test proxied environments (airgap-ish)

### LINKED ISSUE #1242: Environment variables not set correctly from Terraform
**What happened**:

Trying to set host environment variables from Terraform fails with error.

Environment variables should just work.

### LINKED ISSUE #1243: Environment variables not set on first run
**What happened**:

Trying to set proxy environment variables... they are applied only after first run.

Environment variables should be effective also on first run.

### LINKED ISSUE #1244: Control-plane use_proxy default does not work
**What happened**:

When setting proxy envs it seems that those envs are applied by default to control-plane.

`use_proxy` default is `false` which indicates that it should not use proxy envs for control-plane by default.

jakolehm:
- fixes #1243 
- fixes #1242 
- fixes #1244


lib/pharos/phases/configure_host.rb
```diff
@@ -9,6 +9,9 @@ def call
         unless @host.environment.nil? || @host.environment.empty?
           logger.info { "Updating environment file ..." }
           host_configurer.update_env_file
+          host.transport.disconnect
```
jakolehm: It seems that this does not work properly. @kke any ideas?

kke: well, one way would be to do a phase split. ConfigureHostEnvironment --> DisconnectSSH -> ConnectSSH

lib/pharos/host/debian/scripts/repos/pharos_stretch.sh
```diff
@@ -5,6 +5,6 @@ set -eu
 if [ ! -e /etc/apt/sources.list.d/pharos-kubernetes.list ]; then
     curl -fsSL https://bintray-pk.pharos.sh/?username=bintray | apt-key add -
     cat <<EOF >/etc/apt/sources.list.d/pharos-kubernetes.list
-deb https://dl.bintray.com/kontena/pharos-debian stretch main
+deb http://dl.bintray.com/kontena/pharos-debian stretch main
```
jnummelin: Why plain http?

jakolehm: It allows packages to be cached at proxy (makes tests faster and less error prone to network gitches). I'll revert these and open separate issue/pr later.

jakolehm:
@jnummelin @kke PTAL

kke:
Localhost "connections" can't disconnect and reconnect.

The only way to perform env reload there is to log in again.  Or set the env for each command like in #1250

jakolehm:
> Localhost "connections" can't disconnect and reconnect.
> 
> The only way to perform env reload there is to log in again. Or set the env for each command like in #1250

I agree but I also think that there are no real use cases for localhost "connections" and airgap environments (yet). I also think that this should be an implementation detail of local transport (maybe it should re-read envs to a hash on `connect` and then pass those to each exec).

lib/pharos/phases/configure_host.rb
```diff
@@ -7,27 +7,30 @@ class ConfigureHost < Pharos::Phase
 
       def call
         unless @host.environment.nil? || @host.environment.empty?
-          logger.info { "Updating environment file ..." }
+          logger.info "Updating environment file ..."
           host_configurer.update_env_file
+          logger.info "Reconnecting ..."
+          @host.disconnect
```
kke: needs to be `transport.disconnect` or `host.transport.disconnect`

PR #1246, opened 2019-04-01
Set host environment before command execution

kke:
This sets the host environment from config to shell sessions.


kke:
Hmm this may be broken

@client.host is actually astring, not a host, i don't know how the `hostname` method above works (probably doesn't)

PR #1249, opened 2019-04-01
Fix cni dependency error on upgrade

jakolehm:


lib/pharos/host/ubuntu/scripts/install-kube-packages.sh
```diff
@@ -4,7 +4,7 @@ set -e
 
 export DEBIAN_FRONTEND=noninteractive
 apt-mark unhold kubectl kubeadm || echo "Nothing to unhold"
-apt-get install -y "kubectl=${KUBE_VERSION}-00" "kubeadm=${KUBEADM_VERSION}-00"
+apt-get install -y "kubectl=${KUBE_VERSION}-00" "kubeadm=${KUBEADM_VERSION}-00" "kubeadm=${KUBE_VERSION}-00"
```
jnummelin: this is not right, should be the same as for Debian above

jakolehm: yeah, I noticed after I pushed.. I'll push a fix once I have tested upgrade path.

jakolehm: Fixed.

jakolehm:
Tested upgrade on Ubuntu 18.04 from 2.3.1 and it seems to work.

jakolehm:
@jnummelin @kke PTAL

PR #1250, opened 2019-04-01
Set environment separately for every command execution

kke:
(base is #1245)

While ugly, this may be the most fool proof way to set the environment reliably both on ssh and local executions and without need to reconnect.


kke:
This appears to pass now.

PR #1253, opened 2019-04-02
Set environment for local command execs

kke:
Base is #1245 

Sets the environment by sourcing /etc/environment for every command execution on localhost connections.


jakolehm:
@kke rebase needed.

kke:
Done

PR #1254, opened 2019-04-02
Drop non-functional shell variable exports when updating env file

kke:
Base is #1245

Drop the unnecessary exports from Configurer (and related specs)


jakolehm:
@kke needs rebase with master.

kke:
Done

PR #1257, opened 2019-04-02
Remove kubernetes-cni on el7 reset

jakolehm:
Fixes cni dep error after reset.


jnummelin:
LGTM, other than shellcheck wants some mods :D

PR #1261, opened 2019-04-03
Fix drone build status badge

jakolehm:


kke:
Should we label the build badges because they look the same (travis and drone)?

jakolehm:
> Should we label the build badges because they look the same (travis and drone)?

How?

PR #1266, opened 2019-04-04
Fix license assign NoMethodError

### LINKED ISSUE #1264: ERROR: NoMethodError : undefined method `print_subscription_token?' for #<Pharos::LicenseAssignCommand:0x00007fd23802c620>
When trying to assign license I got the following error:
```
ERROR: NoMethodError : undefined method `print_subscription_token?' for #<Pharos::LicenseAssignCommand:0x00007fd23802c620>
```
and license is not assigned properly

kke:
Fixes #1264

The `--print-subscription-token` flag was removed and the logic was changed to output the token when `--cluster-id` has been given.

### DIFF
--- a/non-oss/pharos_pro/commands/license_assign_command.rb
+++ b/non-oss/pharos_pro/commands/license_assign_command.rb
@@ -77,7 +77,7 @@ module Pharos
     end
 
     def subscription_token_request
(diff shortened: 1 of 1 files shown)


kke:
This was not catched in specs because `$stdout.tty?` is false during specs:

```ruby
    logger.info "Exchanging license key for a subscription token" if $stdout.tty? && !print_subscription_token?
```

PR #1280, opened 2019-04-06
Fix local transport mode detect

### LINKED ISSUE #1278: localmode detection considred harmful
I have a vm running in 127.0.0.1:2222

pharos switches to local mode and starts to use my host machine.

jakolehm:
Fixes #1278

### DIFF
--- a/lib/pharos/configuration/host.rb
+++ b/lib/pharos/configuration/host.rb
@@ -38,6 +38,8 @@ module Pharos
       end
 
       def local?
+        return false if ssh_key_path || ssh_port || user
+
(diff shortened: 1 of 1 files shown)


lib/pharos/configuration/host.rb
```diff
@@ -38,6 +38,8 @@ def short_hostname
       end
 
       def local?
+        return false if ssh_key_path || ssh_port || user
```
kke: `ssh_port` has default value of 22

jakolehm: Good catch, fixed.

PR #1291, opened 2019-04-15
Read kubernetes api endpoint from cluster config

### LINKED ISSUE #1290: Kontena Lens: does not use api.endoint if given
**What happened**:

`cluster.yml` has `api.endpoint` set but kontena-lens addon still shows master ip in kubeconfig.

Kubeconfig should use `api.endpoint` if set.

nevalla:
This PR reads kubernetes api endpoint from cluster.yml if provided. Kubernetes API endpoint is used in Lens generated `kubeconfig`.

Fixes #1290

### DIFF
--- a/non-oss/pharos_pro/addons/kontena-lens/addon.rb
+++ b/non-oss/pharos_pro/addons/kontena-lens/addon.rb
@@ -70,10 +70,11 @@ Pharos.addon 'kontena-lens' do
     protocol = tls_enabled? ? 'https' : 'http'
     message = "Kontena Lens is configured to respond at: " + "#{protocol}://#{host}".cyan
     message << "\nStarting up Kontena Lens the first time might take couple of minutes, until that you'll see 503 with the address given above."
+    kubernetes_api_endpoint = cluster_config.api&.endpoint || "https://#{master_host_ip}:6443"
     if config_exists?
-      update_lens_name(name) if configmap.data.clusterName != name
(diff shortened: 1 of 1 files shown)


non-oss/pharos_pro/addons/kontena-lens/addon.rb
```diff
@@ -70,10 +70,11 @@
     protocol = tls_enabled? ? 'https' : 'http'
     message = "Kontena Lens is configured to respond at: " + "#{protocol}://#{host}".cyan
     message << "\nStarting up Kontena Lens the first time might take couple of minutes, until that you'll see 503 with the address given above."
+    kubernetes_api_endpoint = cluster_config.api&.endpoint || "https://#{master_host_ip}:6443"
```
jakolehm: Currently `api.endpoint` does not contain protocol or port.

nevalla: can we assume that protocol is `https`?

jakolehm: We have to make same assumptions as with `master_host_ip`.

nevalla: but port is standard `443`?

jakolehm: > but port is standard 443?

No, as I said we need to make same assumptions .. that includes the port.

PR #1303, opened 2019-04-25
Reload firewalld when removing masquerade

### LINKED ISSUE #1302: Weave: NetPol issue with namespaceSelector across different hosts
**What happened**:

On fresh cluster with weave networking enabled (default pod/service cidr) created 2 namespaces (A,B) and a simple network policy, which enables all pods from namespace B (namespaceSelector) to connect to a pod in namespace A.

At the end, the connection requests from namespace B are getting dropped from weave npc. 

The connection requests from namespace B should be successful

- create new cluster (no addons required, 1 master, 2 nodes) with default pod/service cidr
- create 2 namespaces (A,B) and label them
- deploy an app in each namespace and label them
- create nodeaffinity rule to separate apps over the 2 cluster nodes
- create service for app in namespace A (clusterIP)
- create networkpolicy

```
apiVersion: networking.k8s.io/v1
kind: NetworkPolicy
metadata:
  name: monitoring
  namespace: namespace-A
spec:
  podSelector:
    matchLabels:
      app.kubernetes.io/name: application
  policyTypes:
  - Ingress
  ingress:
  - from:
    - namespaceSelector:
        matchLabels:
          app.kubernetes.io/name: monitoring
    ports:
    - protocol: TCP
      port: 9273
```
- try to connect from app in namespace B (e.g. nc) to app in namespace A using service name

According to kubernetes docs there should be no snat for clusterIP services. However weave npc logs show blocked requests from weave network interface, instead of the source pod interface: `WARN: 2019/04/22 10:10:57.664451 TCP connection from 10.46.0.0:32812 to 10.40.0.3:9273 blocked by Weave NPC `

As far as i can see, there are two possible issues:
- snat for clusterIP which should not be there
- networkpolicy with namespaceSelector does not work (perhaps because of snat)

**Environment**:
- pharos 2.3.6+oss
- Hetzner Cloud VMs
- Ubuntu 18.04.2

**cluster.yml:**

```
hosts:
  - address: 1.2.3.4
    role: master
  - address: 1.2.3.5
    role: worker
    labels:
      node-pool.kubernetes.io: app1
  - address: 1.2.3.6
    role: worker
    labels:
      node-pool.kubernetes.io: app2
name: pharos-cluster
network:
  provider: weave
  dns_replicas: 1
  node_local_dns_cache: true
  service_cidr: 10.96.0.0/12
  pod_network_cidr: 10.32.0.0/12
kubelet:
  feature_gates:
    CSINodeInfo: true
    CSIDriverRegistry: true
control_plane:
  feature_gates:
    CSINodeInfo: true
    CSIDriverRegistry: true
```

jnummelin:
#1297 introduced logic to remove masquerade but the script is actually not called during an upgrade. It's only called during fresh installs or when there's changes to FW configs.

fixes #1302

### DIFF
--- a/lib/pharos/phases/configure_firewalld.rb
+++ b/lib/pharos/phases/configure_firewalld.rb
@@ -23,7 +23,7 @@ module Pharos
         write_config('services/pharos-worker.xml', pharos_worker_service)
         write_config('ipsets/pharos.xml', pharos_ipset)
 
-        # We need to call the reload script to disable masquerade
+        # Masquerade was enabled in the past, if it's still enabled we need to reload firewalld rules
         @firewalld_reload = true if masquerade_active?
 
         return unless firewalld_reload?
--- a/lib/pharos/phases/configure_firewalld.rb
+++ b/lib/pharos/phases/configure_firewalld.rb
@@ -23,6 +23,9 @@ module Pharos
         write_config('services/pharos-worker.xml', pharos_worker_service)
         write_config('ipsets/pharos.xml', pharos_ipset)
 
+        # We need to call the reload script to disable masquerade
+        @firewalld_reload = true if masquerade_active?
+
         return unless firewalld_reload?
 
         cluster_context['reload-iptables'] = true unless @host.new?
@@ -98,6 +101,10 @@ module Pharos
           name: 'pharos',
           entries: trusted_addresses
         )
+      end
+
+      def masquerade_active?
+        transport.exec("firewall-cmd --query-masquerade > /dev/null 2>&1").success?
       end
     end
   end


lib/pharos/phases/configure_firewalld.rb
```diff
@@ -23,6 +23,9 @@ def configure_firewalld
         write_config('services/pharos-worker.xml', pharos_worker_service)
         write_config('ipsets/pharos.xml', pharos_ipset)
 
+        # We need to call the reload script to disable masquerade
```
jakolehm:
```suggestion
        # Masquerade was enabled in the past, if it's still enabled we need to reload firewalld rules
```

PR #1306, opened 2019-05-06
Reload firewalld always when config change is detected

jakolehm:
Fixes an issue where new nodes cannot connect to weave mesh.


jakolehm:
@jnummelin PTAL

PR #1317, opened 2019-05-10
cert-manager extra_args

olanystrom:
Allowing cert-manager to use extra_args for dns01-validation
https://docs.cert-manager.io/en/latest/tasks/acme/configuring-dns01/index.html
--dns01-self-check-nameservers "8.8.8.8:53,1.1.1.1:53"

I hope this is enough change for extra_args to work, I am not sure of the syntax in addon.rb

### DIFF
--- a/addons/cert-manager/addon.rb
+++ b/addons/cert-manager/addon.rb
@@ -26,7 +26,7 @@ Pharos.addon 'cert-manager' do
       required(:email).filled(:str?)
       optional(:server).filled(:str?)
     }
-    
+
     optional(:extra_args).each(:str?)
 
     optional(:ca_issuer).schema {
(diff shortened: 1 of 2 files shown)


addons/cert-manager/addon.rb
```diff
@@ -17,13 +17,15 @@
   config {
     attribute :issuer, issuer
     attribute :ca_issuer, ca_issuer.default(proc { ca_issuer.new(enabled: true) })
+    attribute :extra_args, Pharos::Types::Array.default([])
```
jakolehm:
```suggestion
    attribute :extra_args, Pharos::Types::Array.default(proc { [] })
```

addons/cert-manager/addon.rb
```diff
@@ -17,13 +17,15 @@
...
   }
 
   config_schema {
     required(:issuer).schema {
       required(:name).filled(:str?)
       required(:email).filled(:str?)
       optional(:server).filled(:str?)
+      optional(:extra_args).each(:str?)
```
jakolehm: This needs to be directly under `config_schema` (not under `:issuer`).

jakolehm:
@olanystrom only thing missing is the addon docs, see here: https://github.com/kontena/pharos-docs/blob/master/addons/cert-manager.md

PR #1318, opened 2019-05-13
Fix kubelet proxy configuration race condition

kke:
Trying to fix the race condition in `Configure kubelet proxy`

First round: alter the way initial node labels are set


kke:
Specs and e2e pass, can't confirm if this fixes the 

```
ERROR: Pharos::PhaseManager::Error : Phase failed on 2 hosts:
- 667-do-charming-tapir:
RuntimeError: can't add a new key into hash during iteration
- 667-do-first-wildcat:
RuntimeError: can't add a new key into hash during iteration
```

lib/pharos/configuration/host.rb
```diff
@@ -15,7 +15,7 @@ class Host < Pharos::Configuration::Struct
       attribute :private_address, Pharos::Types::Strict::String.optional.default(nil)
       attribute :private_interface, Pharos::Types::Strict::String.optional.default(nil)
       attribute :role, Pharos::Types::Strict::String
-      attribute :labels, Pharos::Types::Strict::Hash
+      attribute :labels, Pharos::Types::Strict::Hash.default(proc { {} })
```
jakolehm: Is this the minimal fix? Other changes are just extra?

kke: My understanding is that setting node labels in `def labels` is the problem. The changes to `initialize` make the host set the node role labels and taints already upon initialization.

kke:
Looks like merge broke it.

PR #1321, opened 2019-05-13
Improved container-runtime upgrade logic

### LINKED ISSUE #1322: 1.14.0 CRI-O update fails
**What happened**:

Update to 1.14.1 based Kube fails on CRI-O.  No new pods can be created as CRI-O fails to use the images:
```
  Normal   Pulling    29s (x4 over 60s)  kubelet, pharos-hetzner-capital-grubworm  Pulling image "registry.pharos.sh/kontenapharos/weave-npc:2.5.1"
  Normal   Pulled     29s (x4 over 59s)  kubelet, pharos-hetzner-capital-grubworm  Successfully pulled image "registry.pharos.sh/kontenapharos/weave-npc:2.5.1"
  Warning  Failed     29s (x4 over 59s)  kubelet, pharos-hetzner-capital-grubworm  Error: image not known
  Normal   Pulling    14s (x5 over 60s)  kubelet, pharos-hetzner-capital-grubworm  Pulling image "registry.pharos.sh/kontenapharos/weave-kube:2.5.1"
```

Most probably upstream issue as they've been shuffling a lot the image&storage libs

CRIO to update smoothly

Install with any version prior to #1300 getting merged and then use #1300 based branch to update.

**Environment**:
- Pharos version (use `pharos --version`): 2.4.0.alpha
- Cloud provider or hardware configuration:
- OS (e.g. from /etc/os-release; both for the client and for the used nodes): 18.04.2 LTS (Bionic Beaver)
- Others:

**cluster.yml:**

jakolehm:
- updates Ubuntu docker.io package to 18.09
- improves how node drain is executed on container-runtime upgrade
- rolls container-runtime upgrade with 10% concurrency
- merges rollout logic from `pro` to `oss`

Fixes #1322


jakolehm:
@jnummelin @kke PTAL

lib/pharos/phases/validate_version.rb
```diff
@@ -25,7 +25,7 @@ def validate_version(cluster_version)
         cluster_version = build_version(cluster_version)
         raise "Downgrade not supported" if cluster_version > pharos_version
 
-        if requirement.satisfied_by?(cluster_version)
+        if requirement.satisfied_by?(cluster_version) || cluster_version.bump == pharos_version
```
kke: Is the intent here to allow upgrades from `1.0.1` to `1.1.0`, but not `1.0.1` to `1.1.1`? That is what it seems to do.

this seems to do it in one conditional:

```ruby
def requirement_satisfied?
  Gem::Requirement.new('<= ' + pharos_version.to_s, ' <= ' + pharos_version.bump.to_s).satisfied_by?(existing_version)
end
```

jakolehm: These should be allowed without warning (btw I added specs for these):
- `1.0.1` -> `1.0.2`
- `1.0.1` -> `1.0.3`
- `1.0.1` -> `1.1.0`

These should throw a warning:
- `1.0.1` -> `1.2.0`

kke: And not `1.0.99` -> `1.1.1`

jakolehm: Good catch 👏 Should be fixed now.

lib/pharos/phase.rb
```diff
@@ -129,12 +151,35 @@ def delete_stack(name)
...
+    def worker_pool(name, size)
+      self.class.worker_pools[name] ||= Concurrent::FixedThreadPool.new(size)
+    end
+
+    # Blocks until work is done via pool of workers
+    def concurrent_work(name, size)
+      size = 1 if size < 1
+      result = nil
+      error = nil
+      latch = Concurrent::CountDownLatch.new(1)
```
kke: This latch stuff could be avoided with:

```ruby
def concurrent_work(name, size, &block)
  task = Concurrent::Future(executor: worker_pool(name, size), &block)
  raise(task.reason) if task.value.nil? && task.reason.is_a?(Exception)

  task.value
end
```

lib/pharos/phase.rb
```diff
@@ -129,12 +151,35 @@ def delete_stack(name)
...
 
-    def self.mutex
-      @mutex ||= Mutex.new
+    # @return [Concurrent::FixedThreadPool]
+    def worker_pool(name, size)
+      self.class.worker_pools[name] ||= Concurrent::FixedThreadPool.new(size)
+    end
+
+    # Blocks until work is done via pool of workers
+    def concurrent_work(name, size)
```
kke: maybe it should be `throttled_work`

lib/pharos/phase.rb
jakolehm: I like it 👍 (refactored)

lib/pharos/phase.rb
jakolehm: Yeah, better. Fixed.

jakolehm:
@jnummelin PTAL

lib/pharos/phases/configure_host.rb
```diff
@@ -30,8 +32,58 @@ def call
...
+
+        logger.info "Reconfiguring container runtime (#{@host.container_runtime}) packages ..."
+        host_configurer.configure_container_runtime!
+
+        return unless master_healthy?
+
+        logger.info "Uncordoning node ..."
+        sleep 1 until master_host.transport.exec("kubectl uncordon #{@host.hostname}").success?
+        logger.info "Waiting for node to be ready ..."
+        sleep 10 until master_host.transport.exec("kubectl get nodes -o jsonpath=\"{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.type}={@.status};{end}{end}\" | grep 'Ready=True'").success?
```
kke: I don't see how this picks up the correct node.

```
$ kubectl get nodes -o jsonpath="{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.type}={@.status};{end}{end}"
host-00:NetworkUnavailable=False;MemoryPressure=False;DiskPressure=False;PIDPressure=False;Ready=True;InvalidLicense=False;host-01:NetworkUnavailable=False;MemoryPressure=False;DiskPressure=False;PIDPressure=False;Ready=True;host-02:NetworkUnavailable=False;MemoryPressure=False;DiskPressure=False;PIDPressure=False;Ready=True;
```

To me, it appears this will be true when there is at least one node with `Ready=True`.

PR #1325, opened 2019-05-15
Allow parallel command exections on hosts

kke:
Makes it possible to run multiple simultaneous commands on the same host.


jakolehm:
I think this is not safe ... executing commands serially to a host is fast enough for now.

PR #1327, opened 2019-05-15
Allow user to configure CFS quota setting for kubelet

jnummelin:
By default CFS quota is on with period of 100ms. That can cause serious latencies in many cases. See:
- https://github.com/kubernetes/kubernetes/issues/51135
- https://github.com/kubernetes/kubernetes/issues/67577
- https://github.com/zalando-incubator/kubernetes-on-aws/issues/1026
- https://github.com/zalando/public-presentations/blob/master/files/2019-05-15_Optimizing_Kubernetes_Resource_Requests_Limits_for_Cost-Efficiency_and_Latency_-_JAX_DevOps_London.pdf

This PR makes both the CFS quota itself as well as the period configurable for users.

### DIFF
--- a/lib/pharos/config_schema.rb
+++ b/lib/pharos/config_schema.rb
@@ -206,6 +206,8 @@ module Pharos
           optional(:read_only_port).filled(:bool?)
           optional(:feature_gates).filled
           optional(:extra_args).each(type?: String)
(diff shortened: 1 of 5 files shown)


jnummelin:
Setting these on live cluster does not effect running pods, only newly created pods are configured with these.

PR #1331, opened 2019-05-16
Disable docker0 bridge

### LINKED ISSUE #1330: Docker bridge might conflict with pod/service cidr
**What happened**:

We don't explicitly configure subnet for docker0. Because of that there is possibility that docker chooses a subnet that conflicts with the pod/service cidr.

jakolehm:
Fixes #1330


jnummelin:
Maybe run E2E round for this?

jakolehm:
@jnummelin E2E is green.

PR #1335, opened 2019-05-17
More careful reconnect on env-file reconfiguration

kke:
Attempting to get rid of the SSH disconnect error on ConfigureHost reconnect when using bastion.


kke:
Looks like it might help.

kke:
Unnecessary as the connection sharing was reverted in #1338

PR #1341, opened 2019-05-27
Add Lens license manager

nevalla:


jakolehm:
We should probably update Lens before merging.

non-oss/pharos_pro/addons/kontena-lens/resources/19-license-enforcer-deployment.yml.erb
```diff
@@ -0,0 +1,39 @@
...
+spec:
+  selector:
+    matchLabels:
+      app: license-enforcer
+  template:
+    metadata:
+      labels:
+        app: license-enforcer
+    spec:
+      serviceAccountName: license-enforcer
```
jakolehm: I think this needs priority class.. eviction might go south.

nevalla: Yes agree. Priority class added.

PR #1354, opened 2019-05-29
Add max time for command executions

kke:
The automatic termination can be canceled when in an interactive terminal by pressing a key during a 30 second safe-period.


kke:
image

Doesn't bite when e2e is stuck outside of pharos..

kke:
Without something like this, some commands may hang forever. For example, I've had this for an hour now:

```
    [host-00] Waiting for etcd to respond ...
```

PR #1355, opened 2019-05-29
Add node_selector and tolerations config options to Kontena Lens

nevalla:


nevalla:
Need to update docs

non-oss/pharos_pro/addons/kontena-lens/resources/07-dashboard-deployment.yml.erb
```diff
@@ -14,6 +14,21 @@ spec:
       labels:
         app: dashboard
     spec:
+      <%- unless node_selector.empty? -%>
```
jakolehm: Why not just `if config.node_selector` and then `config.node_selector.to_h.each`?

nevalla: Yes, good question. Just copied this part from nginx-ingress-controller addon. I will change this and fix this also for nginx controller in separate PR

PR #1359, opened 2019-05-31
Fix kontena-storage mgr upgrade wait

jakolehm:
Upgrade from 2.3.x -> 2.4.0-beta.1 hangs because of this (if kontena-storage is enabled).


jakolehm:
@nevalla @jnummelin PTAL

jnummelin:
LGTM, had to look up where/how the ceph image used is configured that it now really matches. :)

PR #1360, opened 2019-06-02
Fix weave net ipam consensus race cond

jakolehm:
Weave Net [decides](https://github.com/weaveworks/weave/blob/master/prog/weave-kube/launch.sh#L88) initial IPAM consensus based on kubernetes node addresses. This is mostly fine but might cause troubles if:
- kubernetes api has only subset of peers when list is requested
- kubernetes api does not have addresses for all peers

We should either pre-calculate `KUBE_PEERS` always or set proper `IPALLOC_INIT`. This PR implements the `KUBE_PEERS` way.


jnummelin:
specs still failing

jakolehm:
@jnummelin PTAL

PR #1271, opened 2019-04-05
Make package repos configurable

jnummelin:
fixes #1251

PR #1364, opened 2019-06-03
More SSH keepalive in e2e

jnummelin:
Port SSH fixes from #1271 as separate PR

### DIFF
--- a/e2e/drone.sh
+++ b/e2e/drone.sh
@@ -74,7 +74,7 @@ ssh_userhost="root@${pharos_worker_host}"
 
 scp -o StrictHostKeyChecking=no -i "${ssh_key}" pharos-cluster*.gem "${ssh_userhost}:"
 
-ssh -o StrictHostKeyChecking=no -i "${ssh_key}" "${ssh_userhost}" -- "sudo bash -c '\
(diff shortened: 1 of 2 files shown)


lib/pharos/transport.rb
```diff
@@ -17,6 +17,7 @@ def self.for(host, **options)
         opts[:port] = host.ssh_port
         opts[:keepalive] = true
         opts[:keepalive_interval] = 30
+        opts[:keepalive_maxcount] = 5
```
kke: IMO could be just 1. If a packet is lost, the ssh is probably screwed.

e2e/drone.sh
```diff
@@ -87,7 +87,7 @@ join_command=$(pharos exec -c e2e/digitalocean/cluster.yml --tf-json e2e/digital
 
 echo "Running 'pharos worker up' on target host.."
 
-timeout 600 ssh -o StrictHostKeyChecking=no -i "${ssh_key}" "${ssh_userhost}" -- pharos worker up \
+timeout 600 ssh -o StrictHostKeyChecking=no -o ServerAliveInterval=30 -i "${ssh_key}" "${ssh_userhost}" -- pharos worker up \
```
kke: There's a max count option for this too: ServerAliveCountMax

kke: The default seems to be 3.

PR #1370, opened 2019-06-03
Cert-manager: allow to set multiple issuers/clusterissuers

### LINKED ISSUE #1362: Cert manager: long retry loop
**What happened**:

Cert manager addon gets into annoyingly long retry loop:
```
==> Enabling addon cert-manager
    Retrying after 2 seconds (#1) ...
==> Enabling addon cert-manager
    Retrying after 2 seconds (#2) ...
==> Enabling addon cert-manager
    Retrying after 2 seconds (#3) ...
==> Enabling addon cert-manager
    Retrying after 2 seconds (#4) ...
==> Enabling addon cert-manager
    Retried 5 times, increasing verbosity
    Error: POST /apis/certmanager.k8s.io/v1alpha1/namespaces/default/issuers => HTTP 500 Internal Server Error: Internal error occurred: failed calling webhook "issuers.admission.certmanager.k8s.io": the server is currently unable to handle the request
    Retrying after 2 seconds (#5) ...
...
==> Enabling addon cert-manager
    Error: POST /apis/certmanager.k8s.io/v1alpha1/namespaces/default/issuers => HTTP 500 Internal Server Error: Internal error occurred: failed calling webhook "issuers.admission.certmanager.k8s.io": the server is currently unable to handle the request
    Retrying after 2 seconds (#30) ...
==> Enabling addon cert-manager
==> Enabling addon ingress-nginx
...
```

This is caused by the new admission webhook introduced in 0.7.x cert-manager. the webhook seems to be bit slow to boot-up and all the API calls will fail until it gets up.

We should try to get it up first and only then continue with other things.

`pharos up ...` to go with less error looking things :)

Just build up a new cluster with 2.4.x which has newer cert manager addon.

jakolehm:
Allows to set issuers/clusters issues via addon (previously only default namespace issuer):

Example:
```yaml
addons:
  cert-manager:
    enabled: true
    issuers:
      - kind: ClusterIssuer
        metadata:
          name: le-issuer
        spec:
          acme:
            server: https://acme-staging-v02.api.letsencrypt.org/directory
            email: foo@bar.com
            privateKeySecretRef:
              name: le-issuer
            http01: {}
```

Fixes #1362


addons/cert-manager/addon.rb
```diff
@@ -56,45 +58,34 @@ def self.messages
...
+      config.issuers.each do |i|
+        issuers << K8s::Resource.new(
+          i.merge(
+            apiVersion: "certmanager.k8s.io/v1alpha1"
+          )
+        )
+      end
+      issuers_stack = Pharos::Kube::Stack.new('cert-manager-issuers', issuers)
+      Retry.perform(300, exceptions: [K8s::Error::InternalError]) do
+        issuers_stack.apply(kube_client)
```
jnummelin: I'd guess we also need to handle the `cert-manager-issuers` stack separately whendisabling the addon?

jakolehm: Good point. Although I'm not sure what happens to CRD resources if CRD is removed? Do they get garbage collected automatically?

jnummelin: They should get GC'd:
> When you delete a CustomResourceDefinition, the server will uninstall the RESTful API endpoint and delete all custom objects stored in it.

> If you later recreate the same CustomResourceDefinition, it will start out empty.

addons/cert-manager/addon.rb
```diff
@@ -56,45 +58,34 @@ def self.messages
...
-  def migrate_le_acme_cluster_issuers
-    kube_client.api('certmanager.k8s.io/v1alpha1').resource('clusterissuers', namespace: nil).list.each do |issuer|
-      next unless issuer&.spec&.acme&.server == le_acme_v1_endpoint
+    stack.apply(kube_client)
 
-      rc = kube_client.client_for_resource(issuer)
-      rc.merge_patch(issuer.metadata.name, { spec: patch_spec }, strategic_merge: false)
+    unless config.issuers.empty?
+      logger.info "Applying issuers (this might take a moment) ..."
+      issuers = []
```
kke:
```suggestion
      issuers = config.issuers.map do |i|
        K8s::Resource.new(
          i.merge(
            apiVersion: "certmanager.k8s.io/v1alpha1"
          )
        )
      end
```

addons/cert-manager/addon.rb
```diff
@@ -56,45 +58,34 @@ def self.messages
...
+      config.issuers.each do |i|
```
kke:
```suggestion
```

addons/cert-manager/addon.rb
```diff
@@ -56,45 +58,34 @@ def self.messages
...
+        issuers << K8s::Resource.new(
```
kke:
```suggestion
```

addons/cert-manager/addon.rb
```diff
@@ -56,45 +58,34 @@ def self.messages
...
+          i.merge(
```
kke:
```suggestion
```

addons/cert-manager/addon.rb
```diff
@@ -56,45 +58,34 @@ def self.messages
...
+            apiVersion: "certmanager.k8s.io/v1alpha1"
```
kke:
```suggestion
```

addons/cert-manager/addon.rb
```diff
@@ -56,45 +58,34 @@ def self.messages
...
+          )
```
kke:
```suggestion
```

addons/cert-manager/addon.rb
```diff
@@ -56,45 +58,34 @@ def self.messages
...
+        )
```
kke:
```suggestion
```

addons/cert-manager/addon.rb
```diff
@@ -56,45 +58,34 @@ def self.messages
...
+      end
```
kke:
```suggestion
```

kke: (Trying to use this "batch" thing in github suggestion)

jakolehm:
@jnummelin PTAL again

PR #1371, opened 2019-06-04
Terraform 0.12 support

jakolehm:
Terraform 0.12 has different output format which is not compatible with our current parser. This PR includes new output format parser that can be directly merged to `cluster.yml`. This makes integration much more flexible (it should be possible to generate full cluster.yml via terraform output).

Example output config:

```go
output "pharos_cluster" {
    value = {
        hosts = [
            for host in concat(hcloud_server.worker, hcloud_server.master)  : {
                address           = host.ipv4_address
                role              = host.labels["role"]
                user              = "root"
                container_runtime = "${var.container_runtime}"
            }
        ]
        addons = {
            cert-manager = {
                enabled = true
            }
        }
    }
}
```
Old format still works with Terraform < 0.12 so this PR should not break existing setups.


lib/pharos/command_options/tf_json.rb
```diff
@@ -31,18 +32,25 @@ def load_external_config(config_hash)
...
-          config['hosts'].concat(tf_parser.hosts)
-          config['api'].merge!(tf_parser.api) if tf_parser.api
-          config['name'] ||= tf_parser.cluster_name if tf_parser.cluster_name
-          config['addons'].each do |name, conf|
-            if addon_config = tf_parser.addons[name]
-              conf.merge!(addon_config)
+          json = File.read(file)
+          tf_parser = Pharos::Terraform::JsonParser.new(json)
+          if tf_parser.valid?
+            config.deep_merge!(tf_parser.cluster)
```
kke: Here's the defaults:

```
        # @param other [Hash]
        # @param overwrite_arrays [Boolean] when encountering an array, replace the array with the new array
        # @param union_arrays [Boolean] when encountering an array, use Array#union to combine with the existing array
        # @param keep_existing [Boolean] prefer old value over new value
        # @param merge_nil_values [Boolean] overwrite an existing value with a nil value
        # @param merge_non_hash [Boolean] calls .merge on objects that respond to .merge
        def deep_merge(other, overwrite_arrays: true, union_arrays: false, keep_existing: false, merge_nil_values: false, merge_non_hash: false)
```

This will replace any existing arrays with ones from TF, is that intended?

jakolehm: > This will replace any existing arrays with ones from TF, is that intended?

 Not completely.. we should merge top-level `hosts` 🤔

jakolehm:
@jnummelin PTAL

PR #1372, opened 2019-06-04
Retry initial SSH connection opening for 60 seconds

kke:


kke:
image

Is this going to be solved by retrying longer?

jnummelin:
Usually yes. When you spin up infra with TF, TF gives back signal that "everything's done" but the nodes are just booting up the first time. Thus it can take a while to actually get them up and to be able to receive SSH connections.

kke:
Should the retry maybe be in `drone.sh` then?

PR #1387, opened 2019-06-06
Fix kontena-storage toolbox deployment

jakolehm:
Regression from 0.9 update. Service account does not exist anymore.


jnummelin:
https://github.com/kontena/pharos-cluster/blob/master/non-oss/pharos_pro/addons/kontena-storage/addon.rb#L211 <-- needs to be fixed too then I think

jakolehm:
> https://github.com/kontena/pharos-cluster/blob/master/non-oss/pharos_pro/addons/kontena-storage/addon.rb#L211 <-- needs to be fixed too then I think

Fixed.

PR #1393, opened 2019-06-10
Add kube-rbac-proxy to kontena-stats

nevalla:
This PR adds `kube-rbac-proxy` deployment to `kontena-stats` addon. Also, prometheus config map is modified to support `kubernentes_namespace` query param better. Direct request to prometheus service is disabled by network-policy and only access to prometheus is granted to `rbac-proxy` pod.


nevalla:
images are not coming from pharos registry yet.

jakolehm:
Note: requires Kontena Lens update.

nevalla:
@jakolehm PTAL again

PR #1395, opened 2019-06-10
[poc] Normalize command execution

kke:
Normalize command execution environment.

Cons:
- Ugly (sudo is now a function because aliases can't be inherited into subshells)
- Complex conditional logic
- It's not exactly normalized because of ^

Pros:
- Even subshells now have the env they're supposed to
- No need to reconnect
- Should work on `Transport::Local`
- No more `env bash --noprofile --norc -x env -i sudo bash --noprofile --norc -s`

Might be simpler and saner to create some sort of `/usr/local/share/pharos/shell-exec`

### DIFF
--- a/lib/pharos/phases/configure_host.rb
+++ b/lib/pharos/phases/configure_host.rb
@@ -11,9 +11,6 @@ module Pharos
         unless @host.environment.nil? || @host.environment.empty?
           logger.info "Updating environment file ..."
           host_configurer.update_env_file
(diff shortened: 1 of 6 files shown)


kke:
I wonder what's wrong now, why does the `kubectl get nodes` after `up` fail? Is the proxy stuff actually working now and the command doesn't work because of that?

jakolehm:
@kke rebase needed.

PR #1398, opened 2019-06-12
Handle relative ssh key paths in TF JSON and validate file exists

kke:


kke:
I don't know how this is supposed to work anymore.

lib/pharos/exec_command.rb
```diff
@@ -11,13 +11,13 @@ class ExecCommand < Pharos::Command
...
 
     def execute
-      if command_list.empty?
-        signal_usage_error 'interactive mode can not be used with a non-interactive terminal' unless $stdin.tty? && $stdout.tty?
-        run_interactive
-        exit 0
-      end
-
       Dir.chdir(config_yaml.dirname) do
+        if command_list.empty?
```
kke: Config parsing in ssh/exec interactive mode was not done in `config_yaml.dirname` context.

PR #1418, opened 2019-06-18
Allow running without cluster.yml when tf_json present [cluster-e2e]

kke:
With TF 0.12 it's possible to generate the full config from TF, no need to provider cluster.yml then.

### DIFF
--- a/lib/pharos/command_options/load_config.rb
+++ b/lib/pharos/command_options/load_config.rb
@@ -45,8 +45,8 @@ module Pharos
             Pharos::YamlFile.new($stdin, force_erb: true, override_filename: '<stdin>')
           else
             cluster_config = Dir.glob('cluster.{yml,yml.erb}').first
-            signal_usage_error 'File does not exist: cluster.yml' if cluster_config.nil?
-            Pharos::YamlFile.new(cluster_config)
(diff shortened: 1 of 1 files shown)


lib/pharos/command_options/load_config.rb
```diff
@@ -45,8 +45,8 @@ def default_config_yaml
             Pharos::YamlFile.new($stdin, force_erb: true, override_filename: '<stdin>')
           else
             cluster_config = Dir.glob('cluster.{yml,yml.erb}').first
-            signal_usage_error 'File does not exist: cluster.yml' if cluster_config.nil?
-            Pharos::YamlFile.new(cluster_config)
+            signal_usage_error 'File does not exist: cluster.yml' if cluster_config.nil? && !tf_json
```
jakolehm: Is `tf_json` always available?

kke: Now it checks `.nil?`

jakolehm: I meant: will `self.respond_to?(:tf_json)` always?

kke: Not in `terraform apply`, it only has `--config`.

PR #1419, opened 2019-06-18
Ingress-NGINX: allow optionally deploy as service type=loadbalancer

jakolehm:
Previously only `DaemonSet` deployment was supported. This PR adds `LoadBalancer` type deployment which probably makes more sense if cluster has proper loadbalancer implementation (cloud environments, metallb, akrobateo etc).


jakolehm:
Not sure should `replicas` be under `deployment/loadbalancer` or not? Also, should we allow to tweak service `externalTrafficPolicy` so that users could preserve the source ip?

jakolehm:
> I'd try to "merge" the DaemonSet and Deployment stuff into one single template with some if/else structure. Much of those template are the same anyway.

Refactored those to single template.

PR #1429, opened 2019-06-24
Firewalld as DaemonSet

jakolehm:
Moves firewalld from host to Kubernetes DaemonSet.

### DIFF
--- a/lib/pharos/resources/weave/daemon-set.yml.erb
+++ b/lib/pharos/resources/weave/daemon-set.yml.erb
@@ -12,21 +12,7 @@ spec:
     metadata:
       labels:
         name: weave-net
-      annotations:
-        kontena.io/firewalld: "<%= firewalld_enabled %>"
     spec:
-      <% if firewalld_enabled && reload_iptables %>
-      initContainers:
-        # This container performs firewalld reload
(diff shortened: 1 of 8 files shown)


lib/pharos/resources/weave/daemon-set.yml.erb
```diff
@@ -61,12 +47,19 @@ spec:
                   name: weave-passwd
                   key: weave-passwd
           image: '<%= image_repository %>/weave-kube:<%= version %>'
-          livenessProbe:
+          readinessProbe:
```
jakolehm: Upstream has already switched to `readinessProbe`.

lib/pharos/resources/weave/daemon-set.yml.erb
```diff
@@ -61,12 +47,19 @@ spec:
...
             httpGet:
               host: 127.0.0.1
               path: /status
               port: 6784
-            initialDelaySeconds: 30
+          livenessProbe:
```
jakolehm: This ensures that required iptables rules are in place.

jakolehm:
@jnummelin this breaks systems with `POSTROUTING SNAT rule for systemd-resolved stub`.

PR #1432, opened 2019-06-26
Add system:auth-delegator binding to cert-manager-webhook service account

nevalla:
`cert-manager-webhook` service account is missing binding to `system:auth-delegator` role. This is, for example, preventing deleting namespaces properly since service account can't create `subjectaccessreview` resources. This PR adds this missing role binding to service account.


jakolehm:
Is this missing also from upstream manifests?

nevalla:
At least Helm chart has this:
https://github.com/jetstack/cert-manager/blob/master/deploy/charts/cert-manager/webhook/templates/rbac.yaml#L3-L23

PR #1440, opened 2019-08-05
Fix hostname and address uniqueness validation for hosts sharing an IP

### LINKED ISSUE #1439: Validate hostname uniqueness does not take into account ssh port
Same hostname can have multiple nodes in different ssh ports.. now validation blocks these usecases.

kke:
Fixes #1439

Hostname uniqueness check would ignore hosts with the same IP address but a different SSH port.

A host with a non-uniqu address but a different SSH port would have been considered a duplicate, eventhough it's likely a port forward.

### DIFF
--- a/lib/pharos/config_schema.rb
+++ b/lib/pharos/config_schema.rb
@@ -38,6 +38,7 @@ module Pharos
       result[:hosts].each.with_index do |host, idx|
         duplicates = result[:hosts].select { |h| h[:address] == host[:address] && (h[:ssh_port] || 22) == (host[:ssh_port] || 22) }
         next if duplicates.size == 1
+
         messages[:hosts][idx] = { "address:ssh_port" => ["#{host[:address]}:#{host[:ssh_port] || 22} is not unique"] }
       end
 
--- a/lib/pharos/config_schema.rb
+++ b/lib/pharos/config_schema.rb
@@ -34,16 +34,6 @@ module Pharos
       result = schema.call(DEFAULT_DATA.merge(data))
       raise Pharos::ConfigError, result.messages unless result.success?
 
-      messages = { hosts: {} }
-      result[:hosts].each.with_index do |host, idx|
-        duplicates = result[:hosts].select { |h| h[:address] == host[:address] && (h[:ssh_port] || 22) == (host[:ssh_port] || 22) }
-        next if duplicates.size == 1
-
-        messages[:hosts][idx] = { "address:ssh_port" => ["#{host[:address]}:#{host[:ssh_port] || 22} is not unique"] }
-      end
-
-      raise Pharos::ConfigError, messages unless messages[:hosts].empty?
-
       result.to_h
     end
 
@@ -52,6 +42,10 @@ module Pharos
 
       predicate(:hostname_or_ip?) do |value|
         value.match?(/\A\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\z/) || value.match?(/\A[a-z0-9\-\.]+\z/)
+      end
+
+      predicate(:unique_addresses?) do |hosts|
+        hosts.size < 2 || hosts.group_by { |h| "#{h[:address]}:#{h[:ssh_port] || 22}" }.size == hosts.size
       end
     end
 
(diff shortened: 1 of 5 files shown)


lib/pharos/phases/validate_host.rb
```diff
@@ -44,10 +44,10 @@ def check_role
       end
 
       def validate_unique_hostnames
-        duplicates = @config.hosts.reject { |h| h.address == @host.address }.select { |h| h.hostname == @host.hostname }
+        duplicates = @config.hosts.reject { |h| h.address == @host.address }.select { |h| h.hostname == @host.hostname && h.ssh_port == @host.ssh_port }
```
jakolehm: Hmm... why are we calling this unique hostnames and then we validate unique addresses? 🤔

kke: `GatherFacts` does `host.hostname = transport.exec("hostname -f")`.

If it's a port forward to some other host, it should get a different hostname anyway. 

Maybe the unique hostname thing actually needs to just check for duplicate hostnames, no matter what the address / port is, so something like `(@config.hosts - [@host]).find { |h| h.hostname == @host.hostname }` ?

kke: Hmm, in fact, I think instead of reporting hostnames as duplicate, it should have been allowing duplicate hostnames for hosts with the same address?

`@config.hosts.reject { |h| h.address == @host.address }` would have been empty if there were two hosts with the same address but different ssh ports.

jakolehm: We are mixing two different things here:
1. uniqueness of address:port which is used for ssh
2. uniqueness of hostname which is used to identify nodes in kubernetes

kke: config_schema will complain if addresses are equal and does not check ssh_port.

Needs to be fixed also.

kke:
Added host:ssh_port uniqueness validation and removed the unique address validation.

Had to resort to trickery in `ConfigSchema#load` because couldn't think of a way to do it with a dry predicate.

lib/pharos/config_schema.rb
```diff
@@ -34,6 +34,16 @@ def self.load(data)
       result = schema.call(DEFAULT_DATA.merge(data))
       raise Pharos::ConfigError, result.messages unless result.success?
 
+      messages = { hosts: {} }
```
jakolehm: IMO this can be implemented as predicate

```ruby
    def self.build
      # rubocop:disable Lint/NestedMethodDefinition
      Dry::Validation.Params do
        configure do

          # @param hosts [Array<Hash>]
          def unique_addresses?(hosts)
            return false if hosts.group_by { |host| "#{host[:address]}:#{host[:ssh_port]}" }.size < hosts.size

            true
          end
```

```ruby
required(:hosts).filled(:unique_addresses?, min_size?: 1) do
```

kke:
Now uses a predicate.

- It does not tell which host is non-unique
- The message is displayed also when the `hosts:` is missing.
- When it fails, it always makes `size cannot be less than 1` appear also, I suppose it clears the "non valid" `hosts:` and then decides it's too small.

spec/pharos/phases/validate_host_spec.rb
```diff
@@ -150,12 +150,27 @@
...
+        before do
+          config.hosts[0].attributes[:address] = '192.0.2.2'
+          config.hosts[1].attributes[:address] = '192.0.2.2'
+          config.hosts[0].attributes[:ssh_port] = 8022
+          config.hosts[1].attributes[:ssh_port] = 22
+        end
 
-        expect{ subject.validate_unique_hostnames }.to raise_error(Pharos::InvalidHostError, "Duplicate hostname foo for hosts 192.0.2.2,192.0.2.3")
+        it 'raises if duplicates' do
+          expect{ subject.validate_unique_hostnames }.to raise_error(Pharos::InvalidHostError, "Duplicate hostname foo for hosts 192.0.2.2:22,192.0.2.3:22")
```
kke: Hmm, how can this pass? This sort of message is not there anymore

kke: Ah, it's in the phase.

PR #1448, opened 2019-08-07
Warn if local clock and host clock are out of sync

kke:
Shows a warning if localhost and host clocks are more than 60 seconds apart.

image

### DIFF
--- a/lib/pharos/phases/validate_host.rb
+++ b/lib/pharos/phases/validate_host.rb
@@ -6,6 +6,8 @@ module Pharos
       title "Validate hosts"
 
       def call
+        logger.info { "Validating host clock sync ..." }
+        check_clock
         logger.info { "Validating current role matches ..." }
         check_role
         logger.info { "Validating distro and version ..." }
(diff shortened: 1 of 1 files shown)


lib/pharos/phases/validate_host.rb
```diff
@@ -6,6 +6,8 @@ class ValidateHost < Pharos::Phase
       title "Validate hosts"
 
       def call
+        logger.info { "Validating host clock sync ..." }
```
jakolehm: Probably should say that it validates cli <-> host?

kke:
Now it should only display `[default-master-0] SSL_connect returned=1 errno=0 state=error: certificate verify failed (certificate is not yet valid)`

lib/pharos/logging.rb
```diff
@@ -17,7 +17,11 @@ def self.format_exception(exc, severity = "ERROR")
         backtrace = "\n    #{exc.backtrace.join("\n    ")}"
       end
 
-      "Error: #{exc.message.strip}#{backtrace}"
+      if exc.is_a?(Excon::Errors::CertificateError)
+        "Error: #{exc.message.lines.first[/(.+?\))/, 1]}#{backtrace}"
```
jakolehm: Is this really safe? Does excon always have `(something)` in place that we can use for search?

PR #1457, opened 2019-08-13
Kontena Stats addon: fix node-exporter on hybrid clusters

jakolehm:
- use multi-arch image for scraper
- use node-exporter proxy to expose metrics via pod-network (it's accessible across datacenters)


jakolehm:
@jnummelin PTAL

PR #1458, opened 2019-08-13
Create a service account and token during "pharos kubeconfig"

### LINKED ISSUE #1454: Create admin kubeconfig using SA token

Admin kubeconfig should use SA token instead of client cert auth.

**Why is this needed**:

SA token does not expire and it can be easily revoked via kube api if needed. Cert auth does expire and revoke is not possible.

kke:
Fixes #1454 

Create a `pharos-admin` service account and sa token and use those instead of the client certificte in the kubeconfig created during `pharos kubeconfig`.

Instead of downloading the `/etc/kubernetes/admin.conf`, a new config is built from scratch.

### DIFF
--- a/lib/pharos/kube/config.rb
+++ b/lib/pharos/kube/config.rb
@@ -18,12 +18,12 @@ module Pharos
       def config
         @config ||= yaml_content || {
           'apiVersion' => 'v1',
+          'kind' => 'Config',
           'clusters' => [],
           'contexts' => [],
-          'current-context' => nil,
-          'kind' => 'Config',
+          'users' => [],
           'preferences' => {},
-          'users' => []
+          'current-context' => nil
         }
       end
       alias to_h config
@@ -31,7 +31,7 @@ module Pharos
       # Convert to YAML
       # @return [String]
       def dump
-        YAML.dump(config)
+        YAML.dump(JSON.parse(JSON.dump(config))) # dereference to get rid of *1 &1 etc in output
       end
       alias to_s dump
 
--- a/lib/pharos/kubeconfig_command.rb
+++ b/lib/pharos/kubeconfig_command.rb
@@ -4,8 +4,9 @@ module Pharos
   class KubeconfigCommand < Pharos::Command
     options :load_config, :tf_json
 
-    option ['-n', '--name'], 'NAME', 'overwrite cluster name', attribute_name: :new_name
-    option ['-C', '--context'], 'CONTEXT', 'overwrite context name', attribute_name: :new_context
+    option ['-n', '--name'], 'NAME', 'cluster name', attribute_name: :cluster_name, default: 'pharos-cluster'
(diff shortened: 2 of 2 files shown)


kke:
If we want to update the `~/.kube/config` on master, then I guess the service account should be made during `pharos up` instead of `pharos kubeconfig`?

jakolehm:
> If we want to update the `~/.kube/config` on master, then I guess the service account should be made during `pharos up` instead of `pharos kubeconfig`?

Yes.

> Should it be in the `/etc/kubernetes/admin.conf` also?

Maybe yes. @jnummelin wdyt?

kke:
Moved to `ConfigureServiceAccount` phase.

Still needs some specs, though most of it would be `expect(transport).to receive(:exec!).with('xyz')`.

lib/pharos/phases/configure_service_account.rb
```diff
@@ -0,0 +1,102 @@
...
+        end
+
+        config_file.write(config.dump, overwrite: true)
+        config_file.chmod('0600')
+
+        validate
+      end
+
+      def validate
+        transport.exec!('kubectl get -n kube-system serviceaccount/pharos-admin')
```
jakolehm: We are using `kubectl` here because client is not yet configured?

lib/pharos/phases/configure_service_account.rb
```diff
@@ -0,0 +1,102 @@
...
+      end
+
+      def home_kube_dir
+        transport.file(transport.file('~/.kube').readlink(escape: false, canonicalize: true)).tap do |dir|
+          transport.exec!("mkdir '#{dir}' && chmod 0700 '#{dir}") unless dir.exist?
+        end
+      end
+
+      def create_service_account
+        transport.exec!("sudo kubectl get #{KUBECONFIG_PARAM} -n kube-system serviceaccount/#{ADMIN_USER} || sudo kubectl #{KUBECONFIG_PARAM} -n kube-system create serviceaccount #{ADMIN_USER}")
```
jakolehm: IMO we don't need sudo for `kubectl`.

lib/pharos/phases/configure_service_account.rb
kke: I added that just to validate that `kubectl` without sudo on master works without `KUBECONFIG=` or `--kubeconfig=`.

The next phase actually probably should to be changed to use the file from home instead of `/etc/kubernetes`

lib/pharos/phases/configure_service_account.rb
jakolehm: Ok, maybe we need because this points to root readable kubeconfig?

kke: `error: Error loading config file "/etc/kubernetes/admin.conf": open /etc/kubernetes/admin.conf: permission denied`

kke:
Any idea why the validation fails on drone? Works fine on my vagrant.

PR #1489, opened 2019-10-01
Configurer improved version matching (allow RHEL >= 7.4, < 8.0)

kke:
Makes it possible to define more advanced host configurer version matchers. Currently it's only possible to list each version separately:

Before:

```ruby
class Rhel7 < Configurer
  register_config 'rhel', '7.4'
  register_config 'rhel', '7.5'
  register_config 'rhel', '7.6'
  register_config 'rhel', '7.7'
end
```

After:

```ruby
class Rhel7 < Configurer
  register_config 'rhel', /^7\.[4-9]|\d{2,}/
end
```

This has been used in this PR to make the rhel7 configurer match RHEL `>= 7.4 && < 8.0`. (Including something like `7.19-beta2.rc4`)

The matcher can be anything that handles triple equals.

### DIFF
--- a/lib/pharos/host/configurer.rb
+++ b/lib/pharos/host/configurer.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'set'
+
 module Pharos
   module Host
     class Configurer
@@ -13,7 +15,7 @@ module Pharos
 
       # @return [Array]
       def self.configurers
-        @configurers ||= []
+        @configurers ||= Set.new
       end
 
       # @param [Pharos::Configuration::OsRelease]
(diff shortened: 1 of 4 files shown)


lib/pharos/host/configurer.rb
```diff
@@ -13,7 +15,7 @@ def self.load_configurers
 
       # @return [Array]
       def self.configurers
-        @configurers ||= []
+        @configurers ||= Set.new
```
kke: Unrelated, but I don't see why it should iterate through the same class multiple times when searching for a match.

peterwvj:
Any progress on this? Is it possible that we could merge this any time soon? @kke @jakolehm Thanks in advance.

PR #1498, opened 2019-11-04
HCloud CSI driver v1.1.5

jakolehm:


lib/pharos/resources/hcloud/hcloud-csi.yml.erb
```diff
@@ -135,7 +135,7 @@ spec:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
         - name: hcloud-csi-driver
-          image: kontenapharos/hcloud-csi-driver:1.1.4-kontena.1
+          image: hetznercloud/hcloud-csi-driver:1.1.5
```
jnummelin: Should we "prefix" this with `docker.io/...`?

jakolehm: Does not hurt, fixed.

nevalla:
There are some other changes that we need to test out before merge: https://github.com/hetznercloud/csi-driver/pull/43

jakolehm:
Seems to work. I'll open a separate issue about manifest update.

PR #1517, opened 2019-12-11
Packet example: update to use Terraform 0.12

jakolehm:


examples/terraform-packet/main.tf
```diff
@@ -34,63 +34,118 @@ variable "host_os" {
...
 
-
-output "pharos_addons" {
-  value = {
-      packet-ccm = {
-          project_id    = "${var.project_id}"
-          api_key       = "${var.auth_token}"
-      }
-  }
+resource "packet_bgp_session" "pharos_bgb_session" {
```
scottrobertson: There does not seem to be a way to skip this resource if you don't want to use BGP, which results in:

```
Error: BGP is not enabled
  on main.tf line 84, in resource "packet_bgp_session" "pharos_bgb_session":
  84: resource "packet_bgp_session" "pharos_bgb_session" {
```

scottrobertson: There is probably a better solution, but this works:

```
resource "packet_bgp_session" "pharos_bgb_session" {
  count            = "${var.bgp_address_pool != "" ? var.worker_count : 0}"
  device_id        = packet_device.pharos_worker[count.index].id
  address_family   = "ipv4"
}
```

scottrobertson: Actually, i see you use a similar syntax in the file elsewhere. Will PR that.

PR #1545, opened 2020-05-12
Ubuntu 20.04 LTS Support

### LINKED ISSUE #1544: Ubuntu 20.04 LTS Support
What would you like to be added:
I'd like to be able to install Kubernetes cluster on Ubuntu 20.04 LTS hosts.

Why is this needed:
Now that Ubuntu 20.04 LTS is out, it is the most recent LTS version of Ubuntu

techknowlogick:
Fix #1544

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,15 @@ jobs:
         - NETWORK_PROVIDER=calico
         - FOOTLOOSE_IMAGE=quay.io/footloose/ubuntu18.04
     - stage: e2e
+      name: "e2e: focal docker with calico"
+      script: ./e2e/travis.sh
+      rvm: 2.5.4
+      dist: bionic
+      env:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -21,6 +21,15 @@ jobs:
         - CONTAINER_RUNTIME=docker
         - NETWORK_PROVIDER=weave
         - FOOTLOOSE_IMAGE=quay.io/footloose/ubuntu18.04
+    - stage: e2e
+      name: "e2e: focal docker with weave"
+      script: ./e2e/travis.sh
+      rvm: 2.7.1
+      dist: focal
```
jakolehm, line 28: This needs to be bionic (it's the host OS where we execute tests within docker).

techknowlogick: Thanks :) I've just pushed this change

Issue #1557, opened 2020-09-20
Deploy k8s cluster fails with ctr: mkdir /opt/pharos/bin: no such file or directory\n"


raghavanv90:
When I try to deploy the k8s cluster it fails with a below error,
Environment:
VMs - Tried it on both versions - CentOS Linux release 7.4.1708 (Core) & CentOS Linux release 7.8.2003 (Core)
**pharos 3.2.1  chpharos 0.6.2**

`ERROR: Pharos::PhaseManager::Error : Phase failed on 1 host:
- master:
    Pharos::ExecError: "exec failed with code 1: configure-cfssl.sh\n  + sudo env
      -i - http_proxy= https_proxy= no_proxy= HTTP_PROXY= HTTPS_PROXY= NO_PROXY= FTP_PROXY=
      PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin IMAGE=docker.io/jakolehm/cfssl:0.1.1
      bash --norc --noprofile -x -s\n  + set -e\n  + ctr -n pharos image pull docker.io/jakolehm/cfssl:0.1.1\n
      \ docker.io/jakolehm/cfssl:0.1.1: resolving      |\e[32m\e[0m--------------------------------------|
      \n  elapsed: 0.1 s                  total:   0.0 B (0.0 B/s)                                         \n
      \ docker.io/jakolehm/cfssl:0.1.1: resolving      |\e[32m\e[0m--------------------------------------|
      \n  elapsed: 0.2 s                  total:   0.0 B (0.0 B/s)                                         \n
      \ docker.io/jakolehm/cfssl:0.1.1: resolving      |\e[32m\e[0m--------------------------------------|
      \n  elapsed: 0.3 s                  total:   0.0 B (0.0 B/s)                                         \n
      \ docker.io/jakolehm/cfssl:0.1.1: resolving      |\e[32m\e[0m--------------------------------------|
      \n  elapsed: 0.4 s                  total:   0.0 B (0.0 B/s)                                         \n
      \ docker.io/jakolehm/cfssl:0.1.1: resolving      |\e[32m\e[0m--------------------------------------|
      \n  elapsed: 0.5 s                  total:   0.0 B (0.0 B/s)                                         \n
      \ docker.io/jakolehm/cfssl:0.1.1: resolving      |\e[32m\e[0m--------------------------------------|
      \n  elapsed: 0.6 s                  total:   0.0 B (0.0 B/s)                                         \n
[...]

PR #1558, opened 2020-09-22
Added /opt/pharos folder check

Kerwood:
After the `ctr` release of version 1.3.0 the below command will fail if the `/opt/pharos` folder does not exist (instead of creating it).
```
$ ctr -n pharos install --path /opt/pharos --replace docker.io/jakolehm/cfssl:0.1.1
ctr: mkdir /opt/pharos/bin: no such file or directory
```
I added a simple folder check that creates the folder if it does not already exists. 

See issue here: https://github.com/kontena/pharos-cluster/issues/1557

### DIFF
--- a/lib/pharos/host/debian/scripts/configure-cfssl.sh
+++ b/lib/pharos/host/debian/scripts/configure-cfssl.sh
@@ -2,6 +2,10 @@
 
 set -e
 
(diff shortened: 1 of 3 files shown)


Kerwood:
Was fixed in https://github.com/kontena/pharos-cluster/pull/1556

